Issue #11, opened 2020-01-07
code style checker


OlliKiljunen:
Find suitable linter and configure it.
If possible, add GitHub plugin that enforces lint rules (e.g. prevent PR merge if case of lint errors)

PR #20, opened 2020-01-08
Configured checkstyle with gradle and travis  #11

OlliKiljunen:
Using checkstyle with Google Java Style (google_checks). One can check their coding style with `./gradlew check`. The same check is done by Travis and build fails if there are style issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,7 +77,7 @@ fabric.properties
 
 # *.iml
 # modules.xml
-# .idea/misc.xml
+.idea/misc.xml
 # *.ipr
 
 # Sonarlint plugin
@@ -135,3 +135,7 @@ gradle-app.setting
 **/build/
 
 # End of https://www.gitignore.io/api/java,gradle,intellij
+
+.idea/vcs.xml
+.idea/jarRepositories.xml
+.idea/compiler.xml
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ addons:
     token: $SONARCLOUD_TOKEN
 script:
   - sonar-scanner
+  - ./gradlew check
   
 notifications:
-  slack: letech:6iQNboxAjhcmAx2B0CW7LpJq
+  slack: $SLACK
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -134,4 +134,8 @@ gradle-app.setting
 ### Gradle Patch ###
 **/build/
 
-# End of https://www.gitignore.io/api/java,gradle,intellij
\ No newline at end of file
+# End of https://www.gitignore.io/api/java,gradle,intellij
+
+.idea/vcs.xml
+.idea/jarRepositories.xml
+.idea/compiler.xml
```
OlliKiljunen: I suggest we keep IDE config files out of the repo (even though this is an IDE project :D)

.travis.yml
```diff
@@ -9,6 +9,7 @@ addons:
     token: $SONARCLOUD_TOKEN
 script:
   - sonar-scanner
+  - ./gradlew check
   
 notifications:
-  slack: letech:6iQNboxAjhcmAx2B0CW7LpJq
```
OlliKiljunen: Travis showed alert concerning this so I put it in env var

build.gradle
```diff
@@ -24,4 +25,9 @@ patchPluginXml {
     changeNotes """
       Add change notes here.<br>
       <em>most HTML tags may be used</em>"""
+}
+
+checkstyle {
+    configFile = file('/google_checks.xml')
+    maxWarnings = 0
```
OlliKiljunen: This disallows warnings i.e. minor crimes against style. Our code must be perfect(?)

.gitignore
superseacat: oops!

.travis.yml
superseacat: regenerated the token in slack, added to travis

.gitignore
OlliKiljunen: No, I mean, there were no such files in our repo. My IDE just generated some files that would had been part of this commit but I decided to put those in gitignore instead. Apparently it is arguable if files such as `.idea/misc.xml` should be in VCS or not but I think it's better not have them.

superseacat: ahhh... now I understand :) (I first thought I've pushed some "tech" files myself)

PR #21, opened 2020-01-10
Configuration of proper naming

superseacat:


src/main/resources/META-INF/plugin.xml
```diff
@@ -1,19 +1,15 @@
 <idea-plugin>
     <id>fi.aalto.cs.intellij-plugin</id>
-    <name>Intellij IDEA plugin for A+</name>
-    <vendor email="support@yourcompany.com" url="https://research.cs.aalto.fi/LeTech/">Learning + Technology Research Group</vendor>
+    <name>A+ Plugin for IntelliJ</name>
+    <vendor email="" url="https://research.cs.aalto.fi/LeTech/">LeTech Group at Aalto University</vendor>
```
nikke234: intellij-plugin-letech@aalto.fi?

superseacat: why not...

PR #25, opened 2020-01-13
Code coverage

OlliKiljunen:
# Description
The `check` task of Gradle (which is also run by Travis) measures code coverage using JaCoCo and requires 80 % of instructions be tested.

# Testing

- [x] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -139,3 +139,4 @@ gradle-app.setting
 .idea/vcs.xml
 .idea/jarRepositories.xml
 .idea/compiler.xml
+.idea/codeStyles/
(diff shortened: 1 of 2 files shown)


OlliKiljunen:
Screenshot_2020-01-13 intellij-plugin - Codacy - PullRequests
So there is some problem with Codacy. I have no idea how it should work.

.gitignore
```diff
@@ -139,3 +139,4 @@ gradle-app.setting
 .idea/vcs.xml
 .idea/jarRepositories.xml
 .idea/compiler.xml
+.idea/codeStyles/
```
nikke234: Why not just ignore `.idea` entirely?

PR #26, opened 2020-01-13
Initial layout

nikke234:
# Description
Adds a "placeholder" layout for the plugin, which consists of an exercise list window on the right, and a A+-menu top toolbar button.  

The layout can be inspected using `gradlew runIde`. No unit tests yet, as we still need to decide how to test the GUI of the plugin.

The sonar scanner fails due to a configuration issue unrelated to this PR.

# Testing

- [ ] All unit tests pass
- [x] I have tested manually
 

- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/intellij/actions/AboutAction.java
+++ b/src/main/java/fi/aalto/cs/intellij/actions/AboutAction.java
@@ -1,0 +1,13 @@
+package fi.aalto.cs.intellij.actions;
+
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import fi.aalto.cs.intellij.ui.AboutDialog;
+import org.jetbrains.annotations.NotNull;
+
+public class AboutAction extends AnAction {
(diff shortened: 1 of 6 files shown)


OlliKiljunen:
I was worried that the build would fail because of the lack of unit tests, but it seems that since there are no unit tests at all, code coverage verification is skipped, which is... convenient? ... in this case. :)

line 540 of PR build output
```
Task :jacocoTestCoverageVerification SKIPPED
```

Instead, it's SonarCloud which fails the build.

nikke234:
> but could you please check why travis is not building and why codacy is complaining

As mentioned above, the travis build fails due to configuration issue:
```
org.sonar.java.AnalysisException: Please provide compiled classes of your project with sonar.java.binaries property
``` 

Codacy complains about `AboutDialog` not being named `AboutDialogUtil`, since it only has static methods, which I can fix.

nikke234:
Codacy also complains about an unused private field, which isn't actually unused due to the Swing GUI form, I'll see if theres some way around that.

PR #29, opened 2020-01-14
Remove 'plugin' from the plugin name

nikke234:
# Description
Plugin review by JetBrains:
> Hello,
   Please change the name of your plugin. It should not contain such word as 'Plugin'.

> Hello,
    Could you please remove the word 'Plugin' from the name of your plugin.

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -1,6 +1,6 @@
 <idea-plugin>
     <id>fi.aalto.cs.intellij-plugin</id>
-    <name>A+ Plugin for IntelliJ</name>
+    <name>A+ for IntelliJ</name>
     <vendor email="intellij-plugin-letech@aalto.fi" url="https://research.cs.aalto.fi/LeTech/">LeTech Group at Aalto University</vendor>
 
     <description><![CDATA[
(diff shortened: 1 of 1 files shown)


oseppala:
Maybe something like "A+ LMS". Having IntelliJ in the name of an IntelliJ plugin seems repetitive.

nikke234:
> Maybe something like "A+ LMS". Having IntelliJ in the name of an IntelliJ plugin seems repetitive.
Sounds good to me

raphendyr:
Does `A+ LMS` refer too much to that "this plugin is A+ LMS", when we actually want to say "this plugin communicates with A+ LMS"? So, maybe "A+ Course Assistant" or "A+ Integration" (a bot boring).

"A+" refers to A+ Portal and A+ LMS to the project (including other services too). Sadly, "A+" alone isn't very clear (unique). A+ LMS is bit better, but LMS doesn't mean anything to the user (student).

Few more ideas: "A+ Courses", "A+ Course Helper`, "Course helper for A+" .

jsorva:
_A+ LMS_ is OK, I guess, but Jaakko's point (that it sounds a little bit like the plugin **_is_** the A+ LMS) is valid, too.

Jaakko suggested _A+ Courses_. That's simple and to-the-point. I'll vote for that.

(_Course Helper_ and _Course Assistant_ sound a bit clunkier to me and they don't really add any info beyond _A+ Courses_. _A+ Integration_ is vague and doesn't speak the plugin user's language.)

PR #43, opened 2020-01-16
Making hints on missing plugins

superseacat:
# Description

So it's a big bigger class, so I guess it might be a good point to start some discussion on how do we want to have things. I'll add points that I'd like to discuss explicitly as comments.

# Testing

- [x] All unit tests pass
- [x] I have tested manually

- [x] Yes
- [ ] Not relevant


src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
```diff
@@ -0,0 +1,211 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(RequiredPluginsCheckerAction.class);
+
+  private static final Map<String, String> requiredPluginNames = new HashMap<>();
+  private static Map<String, String> missingOrDisabledPluginNames = new HashMap<>();
+  private static List<IdeaPluginDescriptor> missingOrDisabledIdeaPluginDescriptors;
+  private static List<IdeaPluginDescriptor> availableIdeaPluginDescriptors;
+
+  /**
+   * The plugin startup activity.
+   */
```
superseacat: **comments:** do we comment at all and what level of comments you feel reasonable (currently briefly commented key points)

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
```diff
@@ -0,0 +1,211 @@
...
+      /**
+       * Activate all the required plugins.
+       */
+      @Override
+      public void actionPerformed(@NotNull AnActionEvent e, @NotNull Notification notification) {
+        descriptors.forEach(descriptor -> Objects
+            .requireNonNull(PluginManager.getPlugin(descriptor.getPluginId())).setEnabled(true));
+        notification.expire();
+      }
+    });
```
superseacat: **potential testing**: how do you feel about the anonymous actions (as it might come as an issue if we ever decide to test actions)?

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
```diff
@@ -0,0 +1,211 @@
...
+  /**
+   * Get the list of all the available in the main JetBrains plugin repository.
+   * <p>
+   * Note! The descriptors for the not installed plugins do not exist in IJ until now.
+   */
+  private void getAvailablePluginsFromMainRepo() {
+    try {
+      availableIdeaPluginDescriptors = RepositoryHelper.loadPlugins(new BgProgressIndicator());
+    } catch (IOException ex) {
+      logger.error("Could not retrieve plugins data from main repository.", ex);
```
superseacat: **logging:** a suggested approach to logging. also, do we log any/some events additionally at this point (for say troubleshooting purposes, as currently, the first feedback course stuff gets is when the assignment is submitted) @nikke234))?

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
```diff
@@ -0,0 +1,211 @@
...
+
+  /**
+   * The plugin startup activity.
+   */
+  @Override
+  public void runActivity(@NotNull Project project) {
+    populateRequiredPluginNamesMap();
+    filterMissingOrDisabledPluginNames();
+    createListOfMissingOrDisabledPluginDescriptors();
+    checkPluginsStatusAndNotify();
```
superseacat: **methods style:** there are various approached, but in general:
* can be void, take no parameters, operate on class fields
* can be normal functions (smth in, something out)

superseacat: in the IJ itself, many things are using the first approach, especially when it comes to building the UI (Swing "likes" voids): https://github.com/JetBrains/intellij-community/blob/21dbcdb8b9f07b950ad8440cd217f7fc8a920366/platform/core-impl/src/com/intellij/ide/plugins/PluginManagerCore.java
but also in other places quite a lot

on the other hand, testing of void methods is kinda impossible, but if they are private as well, then not explicitly required

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.Objects;
+import java.util.StringJoiner;
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+//TODO THINK HOW TO TEST & REFACTOR THIS
+public class RequiredPluginsCheckerAction implements StartupActivity {
+
+  private static final Logger logger = LoggerFactory.getLogger(RequiredPluginsCheckerAction.class);
```
superseacat: **logging:** default logging (slf4j) implementation used within the IJ framework

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
```diff
@@ -0,0 +1,211 @@
...
+      missingOrDisabledPluginNames.forEach((name, id) -> {
+        PluginId missingPluginId = PluginId.getId(id);
+        availableIdeaPluginDescriptors
+            .stream()
+            .filter(
+                availableDescriptor -> availableDescriptor.getPluginId().equals(missingPluginId)
+            )
+            .findFirst()
+            .ifPresent(missingOrDisabledIdeaPluginDescriptors::add);
+      });
```
superseacat: **stream API:** how comfortable are you with Stream API of Java8

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  /**
+   * Notify with with an option to enable all the required plugins.
+   */
+  private void notifyAndSuggestPluginsEnabling(List<IdeaPluginDescriptor> descriptors) {
+    Notification notification = new Notification(
+        "A+",
+        "A+",
+        "Some plugins must be and enabled for the A+ plugin to work properly "
```
superseacat: **constants:** sometimes constant Strings are externalized into a separate companion constant classes. the evident benefit is shorter variable names, but the actual visibility gets reduced a bit

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,212 @@
...
+//TODO THINK HOW TO TEST & REFACTOR THIS
+public class RequiredPluginsCheckerActivity implements StartupActivity {
+
+  private static final Logger logger = LoggerFactory
+      .getLogger(RequiredPluginsCheckerActivity.class);
+
+  private static final Map<String, String> requiredPluginNames = new HashMap<>();
+  private static Map<String, String> missingOrDisabledPluginNames = new HashMap<>();
+  private static List<IdeaPluginDescriptor> missingOrDisabledIdeaPluginDescriptors;
+  private static List<IdeaPluginDescriptor> availableIdeaPluginDescriptors;
```
OlliKiljunen: I find it confusing that these are static. I guess it doesn't really matter as there is only one instance of this class but I would find it more intuitive if these were members of that instance instead of static.

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,212 @@
...
+            PluginDownloader pluginDownloader = PluginDownloader.createDownloader(descriptor);
+            pluginDownloader.prepareToInstall(new BgProgressIndicator());
+            pluginDownloader.install();
+          } catch (IOException ex) {
+            logger.error("Could not install plugin" + descriptor.getName() + ".", ex);
+          }
+        });
+        notification.expire();
+        PluginManagerConfigurable
+            .shutdownOrRestartApp("Plugins required for A+ course are now installed");
```
OlliKiljunen: This is probably correct but I just don't get what is actually shutdown or restarted here. IDEA?

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
OlliKiljunen: The level you have commented seems very reasonable and easy to understand to me. :+1:

Only this particular comment is perhaps misplaced (ain't it more like the title of the whole class than just this method?). Also, there can be more activities that are run on startup but this comment makes it seem as this is _the_ (one and only) startup activity.

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
OlliKiljunen: If we want to unit test those actions they can be easily refactored into classes. Until there are unit tests, I see no reason to do that. IMO this is fine. :+1:

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
OlliKiljunen: Juha didn't prioritize that kind of logging, and I'd say we shouldn't implement it because no one has asked us and thus it is unsure if anyone would ever read those logs. (I understand that we are now speaking about logs that would be read by the course staff, not the devs/maintainers of this plugin.)

OlliKiljunen: And btw this suggested approach seems very good. :+1:

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
superseacat: it restarts the IDE ultimately. 

`PluginManagerConfigurable.shutdownOrRestartApp(...` is just a wrapper in PluginManagerConfigurable class that actually refers to a deeper service, that does the restart (as some parts of Plugin Management require restarting the IDE). 

for details see:
https://github.com/JetBrains/intellij-community/blob/f91f570d6ebf62c71742e99e7ac08a07ad4d17b3/platform/platform-impl/src/com/intellij/ide/plugins/PluginManagerConfigurable.java#L1315-L1323

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
OlliKiljunen: "normal functions (smth in, something out)" is my personal favourite of these two (as I am always a bit scared of objects whose state changes a lot :D) but your implementation is very clear and nice so there is no strong reason to "fix" it now.
I agree that "normal functions" are usually easier to write unit tests on but I don't see how that would be impossible for void functions either (void function should cause certain side-effects, in this case change the object's state, so in a unit test we check that these changes have happened).
Also, I am not sure if a method being `private` is a good enough reason to not test it.

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
OlliKiljunen: "shorter variable names" you mean variable names are shorter than string literals? Another benefit is easier localization (different language versions) but it is hardly an issue in our case. This is fine now and shouldn't be difficult to refactor later if needed.

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
OlliKiljunen: Oh right. And that text is the title of the popup message window, nice. I didn't notice it when I first tested.
image

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
superseacat: the assumption is, that the used name of the String is somehow shorter than the actual contents

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
superseacat: accessibility of privates is usually difficult, as the tests are residing elsewhere. but there is no reason making util methods public just for the sake of testings

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,212 @@
...
+    PluginId pluginId = PluginId.getId(id);
+    return !PluginManager.isPluginInstalled(pluginId) || PluginManager
+        .isDisabled(pluginId.getIdString());
+  }
+
+  /**
+   * If there any plugins missing, creates a list of the plugin descriptors for them based on the
+   * publicly available ones.
+   */
+  private void createListOfMissingOrDisabledPluginDescriptors() {
```
superseacat: **method naming:** I usually have longer method names like here, any comments/objections/proposals?

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
OlliKiljunen: If we wanted to unit test those methods, we could make them package-private (i.e. remove the `private` modifier).

Another question is, whether methods that are only meant to be used internally in a class should be unit tested at all. But in this case, I'd say it's much easier to test these four methods separately (plus some basic tests for the entirety) than to write a lot of different tests to the public interface trying to cover all the corner cases. But this is something we don't have to resolve now.

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
nikke234: great stuff!

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
nikke234: The suggested approach looks good :+1: Additional logging at this point is not needed in my opinion, as it isn't useful right now and it's easy add logging of more events later

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,214 @@
...
+    });
+
+    Notifications.Bus.notify(notification);
+  }
+
+  /**
+   * Join plugin descriptor names with a comma.
+   */
+  @NotNull
+  private StringJoiner getPluginsNamesString(List<IdeaPluginDescriptor> descriptors) {
```
nikke234: Not terribly important, but this could be `private static`.

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,213 @@
...
+    });
+
+    Notifications.Bus.notify(notification);
+  }
+
+  /**
+   * Join plugin descriptor names with a comma.
+   */
+  @NotNull
+  private static StringJoiner getPluginsNamesString(List<IdeaPluginDescriptor> descriptors) {
```
avihavai: Check Collectors.joining(","), e.g. something along the lines of:
descriptors.stream().map(IdeaPluginDescriptor::getName).collect(Collectors.joining(", "));

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,213 @@
...
+    });
+
+    Notifications.Bus.notify(notification);
+  }
+
+  /**
+   * Notify with an option to install all the required plugins and suggest restart.
+   */
+  private void notifyAndSuggestPluginsInstallation(List<IdeaPluginDescriptor> descriptors) {
+    Notification notification = new Notification(
```
avihavai: I'd consider extracting this to a separate method, e.g. "notifyAboutMissingPlugins.". You could also have a separate "notifyAbout"-method that then would be used from both the "notifyAboutMissingPlugins" and "notifyAboutDisabledPlugins".

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,213 @@
...
+    } else if (disabledPluginDescriptors.size() > 0) {
+      notifyAndSuggestPluginsEnabling(disabledPluginDescriptors);
+    }
+  }
+
+  /**
+   * Notify with an option to enable all the required plugins.
+   */
+  private void notifyAndSuggestPluginsEnabling(List<IdeaPluginDescriptor> descriptors) {
+    Notification notification = new Notification(
```
avihavai: I'd consider extracting this to a separate method, e.g. "notifyAboutDisabledPlugins".

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,213 @@
...
+      availableIdeaPluginDescriptors = RepositoryHelper.loadPlugins(new BgProgressIndicator());
+    } catch (IOException ex) {
+      logger.error("Could not retrieve plugins data from main repository.", ex);
+    }
+  }
+
+  /**
+   * Sorts required plugins into missing and disabled and shows respective notifications.
+   */
+  private void checkPluginsStatusAndNotify() {
```
avihavai: I'd consider splitting this into two pieces, where one would be about notifying about the missing plugins, one about notifying disabled plugins. Either as separate methods or as separate logical pieces of code, as now the functionality is interleaved. 

e.g.
List<IdeaPluginDescriptor> missingPluginDescriptors = missingOrDisabledIdeaPluginDescriptors.stream().filter(bla).collect(Collectors.toList());
if (!missingPluginDescriptors.isEmpty()) {
    notifyAndSuggestPluginsInstallation(missingPluginDescriptors);
}

List<IdeaPluginDescriptor> disabledPluginDescriptors = missingOrDisabledIdeaPluginDescriptors.stream().filter(bla).collect(Collectors.toList());
if (!disabledPluginDescriptors.isEmpty()) {
    notifyAndSuggestPluginsEnabling(disabledPluginDescriptors);
}

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,213 @@
...
+
+    for (IdeaPluginDescriptor descriptor : missingOrDisabledIdeaPluginDescriptors) {
+      if (!PluginManager.isPluginInstalled(descriptor.getPluginId())) {
+        missingPluginDescriptors.add(descriptor);
+      } else if (PluginManager.isDisabled(descriptor.getPluginId().getIdString())) {
+        disabledPluginDescriptors.add(descriptor);
+      }
+    }
+
+    if (missingPluginDescriptors.size() > 0) {
```
avihavai: You could also use if (!list.isEmpty()), but either way is ok. I typically prefer isEmpty (unless there's an existing agreement on how to handle this.

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,213 @@
...
+  private void notifyAndSuggestPluginsEnabling(List<IdeaPluginDescriptor> descriptors) {
+    Notification notification = new Notification(
+        "A+",
+        "A+",
+        "Some plugins must be and enabled for the A+ plugin to work properly "
+            + getPluginsNamesString(descriptors) + ".",
+        NotificationType.WARNING);
+
+    notification.addAction(new NotificationAction(
+        "Enable the required plugin(s) (" + getPluginsNamesString(descriptors) + ").") {
```
avihavai: It seems that indentation is set to 2 spaces, but here it is 4. Is this intentional?

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,213 @@
...
+   */
+  private void notifyAndSuggestPluginsInstallation(List<IdeaPluginDescriptor> descriptors) {
+    Notification notification = new Notification(
+        "A+",
+        "A+",
+        "Additional plugin(s) must be installed and enabled for the A+ plugin to work "
+            + "properly (" + getPluginsNamesString(descriptors) + ").",
+        NotificationType.WARNING);
+
+    notification.addAction(new NotificationAction(
```
avihavai: For readability, I'd consider extracting the NotificationAction from here e.g. to a separate class.

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,213 @@
...
+  /**
+   * If there are any plugins missing, creates a list of the plugin descriptors for them based on the
+   * publicly available ones.
+   */
+  private void createListOfMissingOrDisabledPluginDescriptors() {
+    if (missingOrDisabledPluginNames.size() > 0) {
+      getAvailablePluginsFromMainRepo();
+      missingOrDisabledIdeaPluginDescriptors = new ArrayList<>();
+      missingOrDisabledPluginNames.forEach((name, id) -> {
+        availableIdeaPluginDescriptors
```
avihavai: This won't work if downloading of plugins failed.

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,213 @@
...
+  /**
+   * Get the list of all the available in the main JetBrains plugin repository.
+   *
+   * <p>Note! The descriptors for the not installed plugins do not exist in IJ until now.
+   */
+  private void getAvailablePluginsFromMainRepo() {
+    try {
+      availableIdeaPluginDescriptors = RepositoryHelper.loadPlugins(new BgProgressIndicator());
+    } catch (IOException ex) {
+      logger.error("Could not retrieve plugins data from main repository.", ex);
```
avihavai: We should do something else in addition to adding a log entry, e.g. a notification. There's bound to be people with e.g. internet down. The possibility of failing to download plugin information should be taken into account in runActivity-method.

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * A startup activity for the plugin.
+ */
+public class RequiredPluginsCheckerActivity implements StartupActivity {
+
+  private static final Logger logger = LoggerFactory
+      .getLogger(RequiredPluginsCheckerActivity.class);
+
+  private final Map<String, String> requiredPluginNames = new HashMap<>();
+  private Map<String, String> missingOrDisabledPluginNames = new HashMap<>();
```
avihavai: Need to consider the need of these attributes. How relevant are they to the class? Check comment on runActivity.

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -0,0 +1,213 @@
...
+  private List<IdeaPluginDescriptor> missingOrDisabledIdeaPluginDescriptors;
+  private List<IdeaPluginDescriptor> availableIdeaPluginDescriptors;
+
+  /**
+   * An actual startup work gets done here.
+   */
+  @Override
+  public void runActivity(@NotNull Project project) {
```
avihavai: I'd consider structuring this so that you do not need the class attributes. E.g. the method populateRequiredPluginsMap would be colled getRequiredPluginsMap (and return the current set). Then, you would pass this to the getMissingOrDisabledPlugins method that'd return a list of missing or disabled plugins based on the parameter (merging methods filterMissingOrDisabledPluginNames and createListOfMissingOrDisabledPluginDescriptors). Then, finally, this list would be passed to the checkPluginsStatusAndNotify-method, which would iterate the list and act accordingly.

src/main/java/fi/aalto/cs/intellij/actions/RequiredPluginsCheckerAction.java
superseacat: another potential candidate, meet JB's own [ApiStatus.Internal](https://github.com/JetBrains/java-annotations/blob/450e2e8e109680eee86faaa80520d0819b1821fa/common/src/main/java/org/jetbrains/annotations/ApiStatus.java#L51)

src/test/java/fi/aalto/cs/intellij/actions/EnablePluginsNotificationActionTest.java
```diff
@@ -0,0 +1,58 @@
...
+  @Test
+  public void testActionPerformed() {
+    // given
+    List<IdeaPluginDescriptor> descriptorList = getDummyPluginsListOfTwo();
+    Map<PluginId, IdeaPluginDescriptor> descriptors = new HashMap<>();
+
+    descriptorList.forEach(descriptor -> {
+      descriptor.setEnabled(false);
+      descriptors.put(descriptor.getPluginId(), descriptor);
+    });
```
OlliKiljunen: I guess it's a matter of taste but I'd avoid loops and such complex code constructions in tests and try to write more "idiot proof" like:
```
descriptor1.setEnabled(false);
descriptor2.setEnabled(false);

descriptors.put(descriptor1);
descriptors.put(descriptor2);
```

src/main/java/fi/aalto/cs/intellij/actions/InstallPluginsNotificationAction.java
```diff
@@ -0,0 +1,128 @@
...
+   * of this interface is {@code InstallPluginsNotificationAction::proposeRestart}.
+   */
+  @FunctionalInterface
+  public interface RestartProposer {
+
+    /**
+     * Calls for an IJ framework to trigger a restart of the IDE.
+     */
+    void proposeRestart();
+  }
```
OlliKiljunen: I find this a nice enough solution at least for now. If the framework interface of this class gets wider from this in the future, we can consider constructing a single facade of these.

src/test/java/fi/aalto/cs/intellij/actions/InstallPluginsNotificationActionFTest.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class InstallPluginsNotificationActionFTest extends PluginsTestHelper {
+
+  @Test
+  public void testProposeRestart() {
+    assertThrows(
+        RuntimeException.class,
+        "Restart IntelliJ IDEA to apply changes in plugins?",
+        InstallPluginsNotificationAction::proposeRestart
+    );
```
OlliKiljunen: This is cool. :sunglasses:

src/test/java/fi/aalto/cs/intellij/actions/EnablePluginsNotificationActionTest.java
```diff
@@ -0,0 +1,58 @@
...
+import com.intellij.openapi.extensions.PluginId;
+import fi.aalto.cs.intellij.PluginsTestHelper;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class EnablePluginsNotificationActionTest extends PluginsTestHelper {
```
OlliKiljunen: Should only FTest classes extend `PluginTestHelper`. I see that methods like `getDummyPluginsListOfTwo` are used in this tests but, then again, `getDummyPluginsListOfTwo` does not use IntelliJ's test framework, right? So, perhaps, the class that provides helper methods like `getDummyPluginsListOfTwo` should be accessed through composition instead of inheritance.

OlliKiljunen: But this is a very minor issue. Absolutely no reason to refactor it now!

src/test/java/fi/aalto/cs/intellij/actions/InstallPluginsNotificationActionTest.java
```diff
@@ -0,0 +1,45 @@
...
+    Notification notification = mock(Notification.class);
+
+    //when
+    InstallPluginsNotificationAction installPluginsNotificationAction =
+        new InstallPluginsNotificationAction(
+            descriptorList,
+            (descriptor) -> {
+              numberOfCallsOfInstall.getAndIncrement();
+            },
+            numberOfCallsOfRestartProposal::getAndIncrement);
```
OlliKiljunen: Could Mockito be used here too, (so we could avoid those AtomicIntegers which, yeah I know, my bad, are a bit awkward)? I don't know.

src/test/java/fi/aalto/cs/intellij/actions/EnablePluginsNotificationActionFTest.java
```diff
@@ -0,0 +1,42 @@
...
+    Notification notification = Mockito.mock(Notification.class);
+    enableMissingPluginsAction.actionPerformed(anActionEvent, notification);
+
+    //then
+    assertTrue("The core plugin is enabled.",
+        ideaPluginDescriptor.isEnabled());
+    assertEquals("Action text contains a name of a core plugin.",
+        "Enable the required plugin(s) (IDEA CORE).",
+        enableMissingPluginsAction.getTemplateText());
+    verify(notification, times(1)).expire();
```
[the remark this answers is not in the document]
superseacat: I'm not sure, it's just the way in compiles :laughing:

src/test/java/fi/aalto/cs/intellij/actions/EnablePluginsNotificationActionFTest.java
```diff
@@ -0,0 +1,42 @@
...
+        new EnablePluginsNotificationAction(descriptorList);
+
+    AnActionEvent anActionEvent = Mockito.mock(AnActionEvent.class);
+    Notification notification = Mockito.mock(Notification.class);
+    enableMissingPluginsAction.actionPerformed(anActionEvent, notification);
+
+    //then
+    assertTrue("The core plugin is enabled.",
+        ideaPluginDescriptor.isEnabled());
+    assertEquals("Action text contains a name of a core plugin.",
```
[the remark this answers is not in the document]
superseacat: more or less

PR #50, opened 2020-01-20
Try testing approach

superseacat:
# Description
So, one of the examples, I'll add some thoughts/comments/questions in the sources.

# Testing

- [ ] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,7 @@ repositories {
 
 dependencies {
     testCompile group: 'junit', name: 'junit', version: '4.12'
+    testCompile group: 'org.mockito', name: 'mockito-core', version: '3.2.4'
 }
 
 intellij {
--- a/src/main/java/fi/aalto/cs/intellij/actions/EnablePluginsAction.java
+++ b/src/main/java/fi/aalto/cs/intellij/actions/EnablePluginsAction.java
@@ -1,0 +1,35 @@
+package fi.aalto.cs.intellij.actions;
+
+import com.intellij.ide.plugins.IdeaPluginDescriptor;
+import com.intellij.ide.plugins.PluginManager;
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.extensions.PluginId;
+import java.util.Objects;
+import org.jetbrains.annotations.Nls;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+
+public class EnablePluginsAction extends NotificationAction {
+
+  private List<IdeaPluginDescriptor> descriptors;
+
+  public EnablePluginsAction(
+      @Nls(capitalization = Nls.Capitalization.Title) @Nullable String text,
+      @NotNull List<IdeaPluginDescriptor> descriptors) {
+    super(text);
+    this.descriptors = descriptors;
+  }
+
+  /**
+   * Activate all the required plugins.
+   */
+  @Override
+  public void actionPerformed(@NotNull AnActionEvent e, @NotNull Notification notification) {
+    descriptors.forEach(descriptor -> descriptor.setEnabled(true));
+    notification.expire();
+  }
+}
--- a/src/main/java/fi/aalto/cs/intellij/actions/EnablePluginsAction.java
+++ b/src/main/java/fi/aalto/cs/intellij/actions/EnablePluginsAction.java
@@ -1,17 +1,13 @@
 package fi.aalto.cs.intellij.actions;
 
 import com.intellij.ide.plugins.IdeaPluginDescriptor;
-import com.intellij.ide.plugins.PluginManager;
 import com.intellij.notification.Notification;
 import com.intellij.notification.NotificationAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
-import com.intellij.openapi.extensions.PluginId;
-import java.util.Objects;
+import java.util.List;
 import org.jetbrains.annotations.Nls;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
-
-import java.util.List;
 
 public class EnablePluginsAction extends NotificationAction {
 
--- a/src/main/java/fi/aalto/cs/intellij/actions/EnablePluginsAction.java
+++ b/src/main/java/fi/aalto/cs/intellij/actions/EnablePluginsAction.java
@@ -1,10 +1,12 @@
 package fi.aalto.cs.intellij.actions;
 
 import com.intellij.ide.plugins.IdeaPluginDescriptor;
+import com.intellij.ide.plugins.PluginManager;
 import com.intellij.notification.Notification;
 import com.intellij.notification.NotificationAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import java.util.List;
+import java.util.Objects;
 import org.jetbrains.annotations.Nls;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -25,7 +27,8 @@ public class EnablePluginsAction extends NotificationAction {
    */
   @Override
   public void actionPerformed(@NotNull AnActionEvent e, @NotNull Notification notification) {
-    descriptors.forEach(descriptor -> descriptor.setEnabled(true));
+    descriptors.forEach(descriptor -> Objects
+        .requireNonNull(PluginManager.getPlugin(descriptor.getPluginId())).setEnabled(true));
     notification.expire();
   }
 }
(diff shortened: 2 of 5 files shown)


src/test/java/fi/aalto/cs/intellij/actions/EnablePluginsActionTest.java
```diff
@@ -0,0 +1,47 @@
...
+    IdeaPluginDescriptorImpl ideaPluginDescriptor = getIdeaPluginDescriptor();
+    ideaPluginDescriptor.setEnabled(false);
+    List<IdeaPluginDescriptor> descriptors = new ArrayList<>();
+    descriptors.add(ideaPluginDescriptor);
+
+    assertFalse(ideaPluginDescriptor.isEnabled());
+
+    EnablePluginsAction enableMissingPluginsAction = new EnablePluginsAction("",
+        descriptors);
+    enableMissingPluginsAction.actionPerformed(anActionEvent, notification);
```
OlliKiljunen: I was first unsure if `actionPerfomed(...)` should be called only by the framework but after doing some research it seems, this is the best (and maybe only) solution here. :+1:

src/test/resources/plugins/a+/plugin.xml
```diff
@@ -0,0 +1,19 @@
+<idea-plugin>
+  <version>0.1.101</version>
+  <idea-version since-build="193.5662" until-build="193.*"/>
+  <id>fi.aalto.cs.intellij-plugin</id>
+  <name>A+ Courses</name>
```
OlliKiljunen: Perhaps change the content (and name?) of this file to indicate it is for the testing purposes only. I can see myself in the future confused on where the change notes I just wrote disappeared when I wasn't careful enough about which file I was modifying. :P

build.gradle
```diff
@@ -16,6 +16,7 @@ repositories {
 
 dependencies {
     testCompile group: 'junit', name: 'junit', version: '4.12'
+    testCompile group: 'org.mockito', name: 'mockito-core', version: '3.2.4'
```
OlliKiljunen: Mockito seems a good choice for a mocking framework. 👍

src/main/java/fi/aalto/cs/intellij/actions/EnablePluginsAction.java
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Activate all the required plugins.
+   */
+  @Override
+  public void actionPerformed(@NotNull AnActionEvent e, @NotNull Notification notification) {
+    descriptors.forEach(descriptor -> Objects
+        .requireNonNull(PluginManager.getPlugin(descriptor.getPluginId())).setEnabled(true));
+    notification.expire();
+  }
```
superseacat: **actions:** so one approach I suggest is **not to use** anonymous actions, so to get them tested easier

src/test/java/fi/aalto/cs/intellij/actions/EnablePluginsActionTest.java
```diff
@@ -0,0 +1,46 @@
...
+    ideaPluginDescriptor.setEnabled(false);
+    List<IdeaPluginDescriptor> descriptors = new ArrayList<>();
+    descriptors.add(ideaPluginDescriptor);
+
+    assertFalse(ideaPluginDescriptor.isEnabled());
+
+    EnablePluginsAction enableMissingPluginsAction = new EnablePluginsAction("",
+        descriptors);
+    AnActionEvent anActionEvent = Mockito.mock(AnActionEvent.class);
+    Notification notification = Mockito.mock(Notification.class);
```
superseacat: **actions:** on the other hand, the logix performed within an action can always be wrapped into a function, and so there won't be a need to mock anything, you just test that function

src/test/java/fi/aalto/cs/intellij/actions/EnablePluginsActionTest.java
superseacat: > **actions:** on the other hand, the logix performed within an action can always be wrapped into a function, and so there won't be a need to mock anything, you just test that function

so there might a way...

src/test/java/fi/aalto/cs/intellij/actions/EnablePluginsActionTest.java
```diff
@@ -0,0 +1,46 @@
...
+    enableMissingPluginsAction.actionPerformed(anActionEvent, notification);
+
+    assertTrue(ideaPluginDescriptor.isEnabled());
+  }
+
+  @NotNull
+  private IdeaPluginDescriptorImpl getIdeaPluginDescriptor() throws IOException, JDOMException {
+    File filePath = new File("src/test/resources/plugins/a+/plugin.xml");
```
superseacat: **test data:** we could some dummy plugin.xml s like in here, so there won't be a need for a manual creation of a IdeaPluginDescriptor (an entity that seem to represent the plugin entity ([there is another one](https://github.com/JetBrains/intellij-community/blob/505e701fbcd34783ee8f5f01b1a4a3363154e55f/platform/platform-impl/src/com/intellij/ide/plugins/PluginNode.java), but it seems to used for a different purpose))

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -36,13 +37,13 @@
   private static final Logger logger = LoggerFactory
       .getLogger(RequiredPluginsCheckerActivity.class);
 
-  private final Map<String, String> requiredPluginNames = new HashMap<>();
+  public final Map<String, String> requiredPluginNames = new HashMap<>();
   private Map<String, String> missingOrDisabledPluginNames = new HashMap<>();
   private List<IdeaPluginDescriptor> missingOrDisabledIdeaPluginDescriptors;
   private List<IdeaPluginDescriptor> availableIdeaPluginDescriptors;
```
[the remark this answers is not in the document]
superseacat: and also as there not gonna be class fields, then access modifiers won't matter

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
```diff
@@ -147,25 +148,16 @@ private void notifyAndSuggestPluginsEnabling(List<IdeaPluginDescriptor> descript
...
-    notification.addAction(new NotificationAction(
-        "Enable the required plugin(s) (" + getPluginsNamesString(descriptors) + ").") {
-
-      /**
-       * Activate all the required plugins.
-       */
-      @Override
-      public void actionPerformed(@NotNull AnActionEvent e, @NotNull Notification notification) {
-        descriptors.forEach(descriptor -> Objects
-            .requireNonNull(PluginManager.getPlugin(descriptor.getPluginId())).setEnabled(true));
```
superseacat: **testing approach:** what happens inside here... basically a loop that performs some actions done by the framework, do we need to test this?

src/test/java/fi/aalto/cs/intellij/actions/EnablePluginsActionTest.java
```diff
@@ -0,0 +1,46 @@
...
+  @Test
+  public void testActionPerformed() throws IOException, JDOMException {
+    IdeaPluginDescriptorImpl ideaPluginDescriptor = getIdeaPluginDescriptor();
+    ideaPluginDescriptor.setEnabled(false);
+    List<IdeaPluginDescriptor> descriptors = new ArrayList<>();
+    descriptors.add(ideaPluginDescriptor);
+
+    assertFalse(ideaPluginDescriptor.isEnabled());
+
+    EnablePluginsAction enableMissingPluginsAction = new EnablePluginsAction("",
```
superseacat: and now it actually fails (as supposed to), because of the PluginManger is getting a `null` [here](https://github.com/Aalto-LeTech/intellij-plugin/pull/50/files#diff-317038de384a7f0c8773d8af5b76827cR30-R32). So it either should be mocked too, or, some base class of the IDE used ([like this one](https://github.com/JetBrains/intellij-community/blob/7862cef1afd5f574d3f6e4a4b1d6bf372c1a71f0/platform/testFramework/src/com/intellij/testFramework/fixtures/BasePlatformTestCase.java)) which I'm to try next

src/main/java/fi/aalto/cs/intellij/activities/RequiredPluginsCheckerActivity.java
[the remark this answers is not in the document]
superseacat: and testing all the "util" functions, thus, we'll have to make them public but annotating with [smth like this](https://github.com/JetBrains/java-annotations/blob/450e2e8e109680eee86faaa80520d0819b1821fa/common/src/main/java/org/jetbrains/annotations/ApiStatus.java#L51)

Issue #44, opened 2020-01-16
add new startup notification saying the plugin is in beta/dev


superseacat:
something like the TMC has: https://github.com/superseacat/tmc-intellij/blob/22f8399de81aca7e86d1e9fff88bf0012f3efd42/tmc-plugin-intellij/src/main/java/fi/helsinki/cs/tmc/intellij/actions/StartupEvent.java#L57-L60

PR #51, opened 2020-01-20
Show beta warning [Issue #44]

OlliKiljunen:
# Description

This task took me a ridiculous amount of time considering how small task it is, but anyway, here we go. So on startup, the version number is checked (version number is stored in `build-info.properties` resource file), and if version is less than 1.0.0 then a warning notification is shown to a user.

# Testing

- [x] All unit tests pass
- [x] I have tested manually

- [x] Yes (Some Javadoc comments)
- [ ] Not relevant


build.gradle
```diff
@@ -1,34 +1,50 @@
...
-checkstyle {
-    configFile = file('/google_checks.xml')
-    maxWarnings = 0
-}
+plugins {
+    id 'java'
+    id 'checkstyle'
+    id 'org.jetbrains.intellij' version '0.4.15'
+    id "org.sonarqube" version "2.8"
+    id 'jacoco'
```
OlliKiljunen: JaCoCo is back! Okay, this is here just so that I can create coverage reports locally. Travis CI is not involved. :P

sonar-project.properties
```diff
@@ -4,3 +4,5 @@ sonar.analysis.mode=publish
 
 sonar.java.binaries=build/classes
 sonar.java.libraries=build/libs
+
+sonar.coverage.exclusions=src/test
```
OlliKiljunen: What I tried was to exclude tests themself from the coverage report (in SonarCloud) but it didn't work. :( Any ideas?

build.gradle
OlliKiljunen: And Jacoco results are sent to Sonar.

src/main/java/fi/aalto/cs/intellij/util/StringSplitter.java
```diff
@@ -0,0 +1,71 @@
...
+ * <pre>
+ * {@code
+ * StringSplitter s = new StringSplitter("hello-123-world", "-");
+ * s.readNext();     // returns "hello"
+ * s.readNextInt();  // returns 123
+ * s.readNext();     // returns "world"
+ * }
+ * </pre>
+ */
+public class StringSplitter {
```
OlliKiljunen: Sonar wasn't happy with regexes so I created this small helper class to parse strings. :(

superseacat: can something like this be used https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#split-java.lang.String-
there are regexes under the hood as well. but as it's Java's own API, then sonar should not care:
`  
  @Test
  public void testStringSplit() {
    String string = "John,,Paul,Ringo Starr,George,";
    String[] stringsArray = string.split(",");

    Assert.assertEquals("John", stringsArray[0]);
    Assert.assertEquals("", stringsArray[1]);
    Assert.assertEquals("Paul", stringsArray[2]);
    Assert.assertEquals("Ringo Starr", stringsArray[3]);
    Assert.assertEquals("George", stringsArray[4]);

//    this fails, as the last trailing is being cut off
//    Assert.assertEquals("", stringsArray[5]);
//
  }
`
 Or the behavior of preserving the last empty "trailing" string is critical?

src/main/java/fi/aalto/cs/intellij/notifications/Notifier.java
```diff
@@ -0,0 +1,10 @@
+package fi.aalto.cs.intellij.notifications;
+
+import com.intellij.notification.Notification;
+import com.intellij.openapi.project.Project;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
```
superseacat: would you feel OK to mark functional interfaces as such, as it would improve the readability: https://docs.oracle.com/javase/8/docs/api/java/lang/FunctionalInterface.html

src/main/java/fi/aalto/cs/intellij/util/StringSplitter.java
OlliKiljunen: I considered that but I was afraid that Sonar would still keep nagging because the separator has to be given as an oh-so-dangerous regex. I can check what happens. Would be nice if we could get rid of this class.
Other chance could be introduce something like Apache Commons library to our project, and use utility functions provided by that.

src/main/java/fi/aalto/cs/intellij/notifications/Notifier.java
OlliKiljunen: Wasn't aware of that. Certainly!

src/main/java/fi/aalto/cs/intellij/common/Resources.java
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Returns properties from a properties resource file.
+   * @param resourceName Name of a resource.
+   * @return Contents of the resource in a {@link Properties} object.
+   */
+  @Nullable
+  public Properties getProperties(@NotNull String resourceName) {
+    try (InputStream stream = resourceProvider.getResourceAsStream(resourceName)) {
+      return getProperties(stream);
```
superseacat: this eventually works and `try` seems to be able to take parameters (I'm kinda surprised :open_mouth:), thus, for the sake of simplicity I suggest to use more conventional approach like:

`   
    try {
      InputStream stream = resourceProvider.getResourceAsStream(resourceName);
      return getProperties(stream);
    } catch (Exception ex) {
`

https://docs.oracle.com/javase/tutorial/essential/exceptions/try.html

src/main/java/fi/aalto/cs/intellij/common/Resources.java
```diff
@@ -0,0 +1,71 @@
...
+      LOGGER.error("Could not read a resource: " + resourceName, ex);
+      return null;
+    }
+  }
+
+  @Nullable
+  private Properties getProperties(@Nullable InputStream stream) throws IOException {
+    if (stream == null) {
+      LOGGER.error("Stream is null.");
+      return null;
```
superseacat: I do not have a strong opinion on this, but sometimes returning `null` is considered a "not so good" practice. there are several approached on this: https://www.codebyamir.com/blog/stop-returning-null-in-java , however, I conceptually like `Exceptions` and maybe `Optional`s can be something to consider: https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html

src/main/java/fi/aalto/cs/intellij/common/Resources.java
OlliKiljunen: I didn't know about that kind of `try` clauses either. First I wrote:
```
    InputStream stream = null;
    try {
      stream = resourceProvider.getResourceAsStream(resourceName);
      return getProperties(stream);
    } catch (Exception ex) {
      LOGGER.error("Could not read a resource: " + resourceName, ex);
      return null;
    }
    finally {
      if (stream != null) {
        try {
          stream.close();
        } catch (IOException e) { }
      }
    }
```
but then IntelliJ said I should use that simpler form. `Alt+Shift+Enter` and - boom! - I got the committed code.
image

src/main/java/fi/aalto/cs/intellij/notifications/APlusNotifications.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationType;
+import fi.aalto.cs.intellij.common.BuildInfo;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A container class for notifications used by A+ plugin.
+ */
+public class APlusNotifications {
```
superseacat: I see quite little gain from having a container for "A+" notifications, but let's discuss it!

superseacat: if we aim at creating some generic base for A+ notification, then I'd propose something similar to this:
```java
public class APlusNotification extends Notification {

  public APlusNotification(@Nls(capitalization = Capitalization.Sentence) @NotNull String title,
      @NotNull String content,
      @NotNull NotificationType type) {
    // so to emphasize, that this is an "A+" notification
    // eventually, we might need more constructors to match "Notification"s own
    super("A+", title, content, type);
  }
}
```

but, again, the gain from this is quite modest, as there might be a need for more constructors to match the IJ's API.

src/main/java/fi/aalto/cs/intellij/notifications/APlusNotifications.java
```diff
@@ -0,0 +1,40 @@
...
+              + "Use this plugin with caution and on your own risk!",
+          NotificationType.WARNING);
+      this.version = version;
+    }
+
+    @NotNull
+    public BuildInfo.Version getVersion() {
+      return version;
+    }
+  }
```
superseacat: this can be a separate class, as it (I believe) will be a bit more explicit. the instant benefit, is that it becomes searchable "by className" inside the IDE

src/main/java/fi/aalto/cs/intellij/notifications/APlusNotifications.java
superseacat: a drawback, that the https://github.com/Aalto-LeTech/intellij-plugin/pull/51/files#diff-4285c89d754ceac24452c9ab2f58cde1 will keep growing...

OlliKiljunen: I agree it might be unnecessary. Why I ended up creating a class for this (only once used) notification, was that I wanted to write separate unit tests to notifications content and whether it is dispatched or not. So I tested the `notification.getMessage()`, `notification.getTitle()` etc. in *APlusNotificationsTest.java* and in *VersionCheckerActivityTest.java* I only tested that the notification is an instance of the correct class and that its state (indicated by `getVersion()`) is correct.

src/main/java/fi/aalto/cs/intellij/notifications/APlusNotifications.java
OlliKiljunen: Oh, all right. I just replied to your earlier comment where I discussed why this class exists at all, but now I see your earlier comment was just about that Notifications should be in separate files (non-nested classes) which is a different thing. Yes, we can get rid of nested classes; that's totally fine for me.

src/main/java/fi/aalto/cs/intellij/activities/VersionCheckerActivity.java
```diff
@@ -0,0 +1,41 @@
...
+
+  VersionCheckerActivity(@NotNull BuildInfo.Version version,
+                         @NotNull Notifier notifier) {
+    this.version = version;
+    this.notifier = notifier;
+  }
+
+  public VersionCheckerActivity() {
+    this(BuildInfo.SingletonUtility.INSTANCE.version, Notifications.Bus::notify);
+  }
```
superseacat: guess there is no way things will work wo this :unamused:

OlliKiljunen: What do you mean?

src/test/java/fi/aalto/cs/intellij/notifications/NotificationsUtilityTest.java
```diff
@@ -0,0 +1,20 @@
...
+  public void testBetaVersionWarning() {
+    APlusNotifications.BetaVersionWarning notification =
+        new APlusNotifications.BetaVersionWarning(new BuildInfo.Version(0, 22, 315));
+
+    Assert.assertEquals("0.22.315", notification.getVersion().toString());
+    Assert.assertEquals("A+", notification.getGroupId());
+    Assert.assertEquals("A+ Courses plugin is under development", notification.getTitle());
+    Assert.assertThat(notification.getContent(), CoreMatchers.containsString("0.22.315"));
+  }
+}
```
superseacat: if we try to abstract away to some higher levels, then what is done here is to some extent testing of the IJ framework. I'm more or less sure, that the constructor of `Notification` will work as expected, and changing this crucial implementation on the JB's side is unlikely. However, if that is to happen, we can still stick to the "older" (used initially) version of the SDK, but this is highly hypothetical. Or, if we actually aim at assuring, that the `BetaVersionWarning` is still working, then I'd say one single class testing the actual IJ's `Notification` class would be enough for the whole plugin project.

src/main/java/fi/aalto/cs/intellij/common/Resources.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+    Properties props = new Properties();
+    props.load(stream);
+    return props;
+  }
+
+  interface ResourceProvider {
```
superseacat: will this also qualify as functional interface?

src/main/java/fi/aalto/cs/intellij/common/Resources.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Opens an input stream to the resource indicated by {@code name}.
+     * @param name Name of the resource.
+     * @return An open {@link InputStream} which corresponds to {@code name} or {@code null} if
+     *         the stream could not be opened.
+     */
+    @Nullable
+    InputStream getResourceAsStream(@NotNull String name);
```
superseacat: guess this is a matter of taste, but additionally to the [InputStream API](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html) there is a [Files API](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html) which is a bit reacher and might be easier to work with

src/main/java/fi/aalto/cs/intellij/common/Resources.java
OlliKiljunen: I also prefer well-handled Exceptions over null-returns (and perhaps the most often heard argument against exceptions - that they "are slow" - is certainly NOT an issue here). I actually even started with a solution where Resources and BuildInfo throw exceptions which would then be handled by the activity class which can also print more meaningful logs.

But then there was a weird issue that the IDE (one I was running when testing) flashed a red **error** icon in the lower right corner of the window indicating that something was wrong when these exceptions were thrown (I tested with some invalid data) **even though** the exceptions were properly caught. (It sounds weird and I cannot easily reproduce this behavior so there's a possibility that i was mistaken, but I don't think so.)

Of course the IDE wasn't wrong: an exception was thrown because something was "wrong" (e.g. a resource file wasn't found). But I reasoned it is not the responsibility of this classes to decide if that is a big enough issue that it should be reported to the user. So after a moment of uncertainty about how to solve that I remembered a good ol' `null` that represents "nothing". 

Calls `getProperties("foo.properties")`: _Please, give me what you have on 'foo.properties'._

Returns `null`: _Sure, there is no such resource, so here is nothing._

Not sure if this is a good enough argument to keep null-returns but this what I reasoned when I wrote the code.

Also the `@NotNull` and `@Nullable` annotations helped a lot finding potential sources of `NullPointerException`s.

src/main/java/fi/aalto/cs/intellij/common/Resources.java
```diff
@@ -0,0 +1,71 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A class that provides methods to read the resource files of the plugin application.
+ * Normally, this class should be accessed via {@code Resources.SingletonUtility.INSTANCE}.
+ */
+class Resources {
```
superseacat: I'd propose checking the [Files API](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html) for reading the resources, as it has many useful methods there already (and no need to test them too :wink:

src/main/java/fi/aalto/cs/intellij/common/Resources.java
OlliKiljunen: Yes, i'll add the annotation. 👍

src/main/java/fi/aalto/cs/intellij/notifications/APlusNotifications.java
```diff
@@ -0,0 +1,40 @@
...
+              + "which is a pre-release version of the plugin and still under development. "
+              + "Some features of this plugin are still probably missing, "
+              + "and the plugin is not yet tested thoroughly. "
+              + "Use this plugin with caution and on your own risk!",
+          NotificationType.WARNING);
+      this.version = version;
+    }
+
+    @NotNull
+    public BuildInfo.Version getVersion() {
```
superseacat: this seems to be exposed only for the purpose of testing. as we have a `Version` functionality covered comprehensively with tests, not sure if this (and checks for it in tests) are required, but it's an open quiestion

src/main/java/fi/aalto/cs/intellij/common/Resources.java
OlliKiljunen: I am not sure if Files API helps a lot if we still want to access resources via class loader (instead of reading the files from the file system). `ClassLoader` provides essentially two methods to read a resource: `getResource(String)` which returns a URL and `getResourceAsStream(String)` which returns a stream. The former could be used (I'd assume) to get URL that is then opened as a file via Files API but that would just increase complexity IMO.

Not so surprisingly, it seems that `Properties` class have been designed to work in this context as there is a method `load(InputStream)`. This `Resources` class i wrote just provides a slightly simpler interface to that functionality, as it:
1. Instantiates a `Properties` object so the client code doesn't have to do that.
2. Opens and closes the stream.
3. Take care of different kind of error situations (resource doesn't exist, stream couldn't be opened, the content of the stream didn't match the properties format etc.) and indicates any failure by returning `null`. (In client code, which wants to read some properties from resources, it usually makes no difference what kind of problem there was. If those properties are not available for whatever reason, either live with it, i.e. use some defaults, or panic exit.)

src/test/java/fi/aalto/cs/intellij/notifications/NotificationsUtilityTest.java
OlliKiljunen: I'd see this test more like a description on how we want the `BetaVersionWarning` to work written in a form of a unit test.

It says that we expect this notification has "A+" as group ID, "A+ Courses plugin is under development" as title, and we expect that the version we give to its constructor is mentioned somewhere in the message. Now, if someone wants to modify or refactor or do whatever to that class, they can freely do so. As long as this test passes, it has at least those properties, we (or perhaps just *I*?) cared about. Instead, if this test fails, the modifier should consider whether the expected behavior of the class has changed (and revise the test code) or did s/he mess something up (and fix their code).

src/main/java/fi/aalto/cs/intellij/util/StringSplitter.java
OlliKiljunen: You were right! Sonar doesn't consider `String.split(String regex)` as a "security hotspot" (as it did with a general use of regex). So I removed this class. 👍

src/test/java/fi/aalto/cs/intellij/common/BuildInfoTest.java
```diff
@@ -0,0 +1,114 @@
...
+import org.junit.Test;
+
+public class BuildInfoTest {
+
+  @Test
+  public void testCreateVersion() {
+    String versionString = "3.5.24";
+    BuildInfo.Version version = new BuildInfo.Version(3,5,24);
+
+    Assert.assertEquals(3, version.major);
```
superseacat: I'd suggest to import static for the cases when it does not reduce readability

src/test/java/fi/aalto/cs/intellij/activities/VersionCheckerActivityTest.java
```diff
@@ -0,0 +1,42 @@
...
+          Assert.assertNull(project);
+          Assert.assertThat(notification,
+              CoreMatchers.instanceOf(APlusNotifications.BetaVersionWarning.class));
+          Assert.assertEquals("0.22.315",
+              ((APlusNotifications.BetaVersionWarning) notification).getVersion().toString());
+        });
+
+    activity.runActivity(Mockito.mock(Project.class));
+
+    Assert.assertEquals(1, numberOfCalls[0]);
```
superseacat: I understand why and wo functional interfaces a kind of `Mockito.verify()` could be tried, but I can't figure out if it's pluggable here

src/test/java/fi/aalto/cs/intellij/common/BuildInfoTest.java
OlliKiljunen: A good point! Much better readability now. 👍

src/main/java/fi/aalto/cs/intellij/common/BuildInfo.java
```diff
@@ -0,0 +1,123 @@
...
+      return Version.fromString(version);
+    }
+
+    /**
+     * Returns a {@link Version} object based on the value of the given string.
+     * @param version A version string of format "{major}.{minor}.{build}".
+     * @return A {@link Version} object, or {@code null} if {@code version} is invalid or null.
+     */
+    @Nullable
+    public static Version fromString(@Nullable String version) {
```
superseacat: JB themselves use [a simple String for version](https://github.com/JetBrains/intellij-community/blob/863153aec517a3c9b65bfd1993384f4304e79f78/platform/core-impl/src/com/intellij/ide/plugins/IdeaPluginDescriptorImpl.java#L75), but this is fancy!

src/main/java/fi/aalto/cs/intellij/common/BuildInfo.java
```diff
@@ -0,0 +1,123 @@
...
+    public boolean equals(Object obj) {
+      return obj instanceof BuildInfo.Version && toString().equals(obj.toString());
+    }
+  }
+
+  public static class SingletonUtility {
+    public static final BuildInfo INSTANCE = new BuildInfo(Resources.SingletonUtility.INSTANCE);
+
+    private SingletonUtility() { }
+  }
```
superseacat: I don't understand what does this do?

src/test/java/fi/aalto/cs/intellij/activities/VersionCheckerActivityTest.java
OlliKiljunen: I'd guess we could also use Mockito's dependency injection to mock, say, IntelliJ's `Notifications.Bus.notify(Nofification)`. Personally, I like abstract interfaces more because it makes it explicit which entries of the framework each class use (forcing coder to actually think: *should* this class use this feature of the framework) and also provides greater independence of implementation details of the framework.

With functional interfaces (which I haven't use that much before this project), this approach even looks clean in the code.

src/main/java/fi/aalto/cs/intellij/common/BuildInfo.java
OlliKiljunen: I implemented this kind of version class because we have to compare versions. In future, when we have to check if this course XML is compatible with this version of plugin, we can perhaps use some kind of semantic versioning: e.g. if course requires plugin version "1.6", all versions "1.6", "1.7", "1.8" are fine but "2.0" is not (because a new major version implies breaking changes). But this is something we don't have to care about now.

src/main/java/fi/aalto/cs/intellij/common/BuildInfo.java
OlliKiljunen: Just a quite often-used (anti-?) pattern of lazy initialization. See, if
```
public static final BuildInfo INSTANCE = new BuildInfo(Resources.SingletonUtility.INSTANCE);
```
was in `BuildInfo` class (not in the nested class) it would be executed always when the program execution starts. Now, when it is static nested class it is not executed until some one calls it. In our case that is inside the constructor of `VersionCheckerActivity` (IMO, this is not intuitive at all but just how Java works).

An easier-to-read lazy singleton initialization would of course be:
```
class BuildInfo {
  private static BuildInfo INSTANCE;

  public static BuildInfo getInstance() {
    if (INSTANCE == null) {
      INSTANCE = new BuildInfo(...);
    }
    return INSTANCE;
  }
}
```
but that is not thread safe. Making it synchronized would help but many programmers avoid it because of "performance" (which, honestly, in our case would not be an issue, not even close). In early days of Java there were basically no chance to make non-locking thread-safe lazy initialization except this nested static class weirdness (sometimes referred as "Bill Pugh Singleton") and I guess it kinda remained as a pattern that coders (like me) keep repeating (because we are too lazy to think what would be more suitable solution until careful code reviewers catch us). :P

Perhaps the best solution is: we don't need lazy initialization at all so it can be moved out of the nested class. :D

src/main/java/fi/aalto/cs/intellij/common/BuildInfo.java
```diff
@@ -0,0 +1,38 @@
...
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(BuildInfo.class);
+  private static final String RESOURCE_NAME = "build-info.properties";
+
+  public final Version version;
+
+  BuildInfo(@Nullable Properties props) {
+    Version ver = Version.fromProperties(props);
+    if (ver == null) {
+      ver = new Version(0,0,0);
```
superseacat: I'm somehow unsure about this, why "0, 0, 0"?

superseacat: I'd suggest that possibly better composition would be achieved, if maybe `Version` class itself could tell if the provided data is invalid and reject such request already on the instance initialization stage

src/test/java/fi/aalto/cs/intellij/common/ResourcesTest.java
```diff
@@ -0,0 +1,71 @@
...
+    assertFalse(stream.hasCloseMethodBeenCalled);
+
+    Properties props = res.getProperties("correct-properties");
+
+    assertTrue(stream.hasCloseMethodBeenCalled);
+
+    assertNotNull(props);
+    assertEquals("x", props.getProperty("a"));
+    assertEquals("y", props.getProperty("b"));
+  }
```
superseacat: I'm a bit skeptical about the [MockInputStream approach](https://github.com/Aalto-LeTech/intellij-plugin/pull/51/files#diff-15e23dbb3e8a5b277f36c9663056e8edR58-R69), but I can't find a way of using `Mockito.verify(...)` or `Mockito.spy(...)`, but I'm still looking into this

src/main/java/fi/aalto/cs/intellij/common/BuildInfo.java
OlliKiljunen: (Sorry for a long post... :P)

I think your point is very valid, but please let me explain my thought on this. :)

What I have tried to achieve (with pretty much all the code in this PR), is that I could read the version info of the software like this
```java
int versionMajor = BuildInfo.INSTANCE.version.major;
// or BuildInfo.getInstance().getVersion().getMajor() or whatever; that doesn't matter now
```
without (in the context of this "client code")
* having to handle any exceptions,
* being unsure if `INSTANCE` or `version` could be null,
* being unsure if version information is initialized yet or not.

Simply: just give me the version of this plugin where I need it.

I think it is totally acceptable to just assume that the version info is always available at run time. Why? I don't say it won't happen (it probably will happen at least once before we are done with thorough integration testing). But if it's *not* available, it's because there is **a bug** (either in our code or in our build configuration that failed to create a valid properties file and make it accessible at run time, that is, include it in the JAR).

When there is bug, the program fails, right? (That is the definition of bug, I guess.) Here the failure is the softest one (just write error log and use "0.0.0" as version, allowing the program execution go on even though in a state where it does not work as it should). The other option is to throw a RuntimeException and let the plugin (and IntelliJ?) die, which, IMO, is not a bad option either. What is common to both of these, neither of them should ever happen in production (but bugs still happen sometimes, of course). *

So why handle this issue here and not in `Version` class. One reason is that it might be totally legit to call `Version.fromString(String)` with an invalid argument. Because that argument may, for example, come from a course config file where it tells the compatible version of the plugin to be used with that file. And that string can be anything (perhaps it's mistyped by some teacher). In that case, we don't want an error log to be written (it's not an error of our program). We probably just abort loading the course and inform the user that the file they tried to use had invalid compatibility information (or just show a warning or whatever we'll decide to do). The point is, the situation is totally different and, thus, is handled totally differently.

Nevertheless, I don't think this constructor is the best place for this "if something is wrong, fail program softly" code. Now when I think it, the correct place would be a static function `BuildInfo.initialize()`, which would call this constructor. This constructor should throw (non-runtime) exception if version is null. The `INSTANCE` field should be initialized like `BuildInfo INSTANCE = initialize();`. So I'll change this, if you agree that that would be better.

*) The best option is perhaps the combination of the two: throw some specific runtime exception but make sure it is caught by some common runtime error handler, which writes the error log and then either recovers or terminates. But I don't know how to create such a handler with this framework, and it's so out of the scope of this task that I won't do it now. If something like that is later implemented, these code locations of "soft failures" can be easily found searching "Logger.error".

src/test/java/fi/aalto/cs/intellij/notifications/NotificationsUtilityTest.java
superseacat: fair point, let's do so!

src/main/java/fi/aalto/cs/intellij/common/BuildInfo.java
```diff
@@ -0,0 +1,38 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Holds information that was gathered during the build of the program (such as version).
+ * Normally, this class should be accessed via {@code BuildInfo.SingletonUtility.INSTANCE}.
+ */
+public class BuildInfo {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(BuildInfo.class);
```
nikke234: Borrowing from tmc-intellij, I've used `logger` in my code. Also see [this](https://stackoverflow.com/questions/1417190/should-a-static-final-logger-be-declared-in-upper-case).

I'm fine with `LOGGER` as well, as long as we're consistent with it everywhere.

src/main/java/fi/aalto/cs/intellij/common/BuildInfo.java
nikke234: missing spaces after commas :D

.gitignore
```diff
@@ -140,3 +140,5 @@ gradle-app.setting
 .idea/jarRepositories.xml
 .idea/compiler.xml
 .idea/codeStyles/
+.idea/checkstyleidea-libs/
+.idea/checkstyle-idea.xml
```
nikke234: Just curious, is there a reason not to ignore .idea entirely?

src/main/java/fi/aalto/cs/intellij/activities/VersionCheckerActivity.java
```diff
@@ -0,0 +1,41 @@
...
+import fi.aalto.cs.intellij.notifications.Notifier;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * An activity that checks, on startup, if the version of the plugin is a pre-release, and, if that
+ * is the case, shows a {@link BetaVersionWarning}.
+ */
+public class VersionCheckerActivity implements StartupActivity {
+
+  private final Notifier notifier;
```
nikke234: Why do we need this instead of just using `Notifications.Bus::notify`?

src/main/java/fi/aalto/cs/intellij/common/BuildInfo.java
superseacat: this is so cool I can't resist: https://gist.github.com/rxaviers/7360908

nikke234: :trollface:

src/main/java/fi/aalto/cs/intellij/activities/VersionCheckerActivity.java
```diff
@@ -0,0 +1,41 @@
...
+  private final Version version;
```
nikke234: This also isn't necessarily needed as an instance variable, since it comes from the build info singleton and never changes

.gitignore
OlliKiljunen: The reason was that it is somewhat interesting to get noticed when the IDE creates files that developers of gitignore.io have considered important enough to be included in repo. But now that this list keeps growing, I think it is good idea just ignore the whole dir.

src/main/java/fi/aalto/cs/intellij/common/Resources.java
```diff
@@ -0,0 +1,72 @@
...
+    @Nullable
+    InputStream getResourceAsStream(@NotNull String name);
+  }
+
+  public static class SingletonUtility {
+    public static final Resources INSTANCE = new Resources(Resources.class.getClassLoader());
+
+    private SingletonUtility() { }
+  }
+}
```
nikke234: Could you explain why this is a singleton and not a utility class with some static methods? Not saying it's bad, just don't quite follow the logic here

src/main/java/fi/aalto/cs/intellij/notifications/BetaVersionWarning.java
```diff
@@ -0,0 +1,32 @@
...
+            + "Use this plugin with caution and on your own risk!",
+        NotificationType.WARNING);
+    this.version = version;
+  }
+
+  @NotNull
+  public Version getVersion() {
+    return version;
+  }
+}
```
nikke234: does this have a use case beyond testing?

src/main/java/fi/aalto/cs/intellij/common/Version.java
```diff
@@ -0,0 +1,89 @@
...
+    String[] parts = version.split("\\.");
+    if (parts.length != 3) {
+      return null;
+    }
+
+    try {
+      major = Integer.parseInt(parts[0]);
+      minor = Integer.parseInt(parts[1]);
+      build = Integer.parseInt(parts[2]);
+    } catch (Exception ex) {
```
nikke234: If we get version strings from "untrusted" sources (i.e. course config file), should we check that these aren't negative?

.gitignore
nikke234: Apparently it [could be useful](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems) to share some files, but since we currently ignore it completely anyways, I'm for ignoring the whole dir.

src/main/java/fi/aalto/cs/intellij/activities/VersionCheckerActivity.java
OlliKiljunen: Just trying to decouple the VersionCheckerActivity from those parts of the framework that would make unit testing difficult. I mean, we could probably use Mockito to do some dependency injection to `Notifications.Bus.notify(...)` and the unit test would be like
```
public void testVersionChecker() {
  Mockito
    .doSomeMagicTo(Notifications.Bus.class)
    .soThatWhenSomeOneCalls('notify')
    .doNotDoWhatItWouldNormallyDoButInsteadDo((notification, project) -> {
      // Assert here that arguments are what they should be.
    })
    .andVerifyThatThisIsCalledExactlyThisManyTimes(1);

    new VersionCheckerActivity(new Version(...)).runActivity(...);
}
``` 
This seems actually OK to me. What I don't like a lot in this code is that the writer of the code has to "just know" that `VersionCheckerActivity.runActivity(...)` calls exactly that method (`Notifications.Bus.notify(...)`) of the framework.

I'd say that accessing `Notifications.Bus.notify(...)` through an interface makes it more explicit (for the client code) in which ways `VersionCheckerActivity` communicates with the framework and decouples it more from the implementation details of the framework.

src/test/java/fi/aalto/cs/intellij/notifications/NotificationsUtilityTest.java
```diff
@@ -0,0 +1,21 @@
+package fi.aalto.cs.intellij.notifications;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+
+import fi.aalto.cs.intellij.common.Version;
+import org.junit.Test;
+
+public class NotificationsUtilityTest {
```
superseacat: should the class name be more specific to `BetaVersionWarning`?

src/main/java/fi/aalto/cs/intellij/common/Resources.java
OlliKiljunen: It's perhaps an outdated and unnecessary Java pattern (sometimes called 'Bill Pugh Singleton'). See discussion here:
https://github.com/Aalto-LeTech/intellij-plugin/pull/51#discussion_r369008279
I think we should remove it. :+1:

src/main/java/fi/aalto/cs/intellij/activities/VersionCheckerActivity.java
nikke234: My understanding is that `Notifications.Bus.notify` is the "standard" way for plugins to display notifications, so in that sense I would actually prefer to assume that specifically `Notifications.Bus::notify` is called and get a failed test if it isn't called. Then again I'm not sure how difficult your pseudocode test approach would be to implement. Is there a (apart from testing) use case that uses a different notifier?

nikke234: Now thinking about it, it's quite convenient to be able to provide the notifier, but I still fail to see a "real world" use case

src/main/java/fi/aalto/cs/intellij/common/Resources.java
superseacat: you don't remove things cause teammates don't understand :stuck_out_tongue:

nikke234: Sorry I was unclear, I mean the entire Resources class, not the inner class

OlliKiljunen: You are right, it perhaps should not be. (And it is not even in a strict sense singleton now because constructors are not private so there is possibility to instantiate multiple objects). What I was thinking that the only use case is to read resources from the our projects class loader so the object would always be constructed the same way. So it seemed a good idea to store that object in a static field and always use the same object instead of creating always a new object.

But now when I think it, it certainly should not be singleton. If the client code wants to reuse a single instance of it, it is the responsibility of theirs, but not the one of this class. :+1:

src/main/java/fi/aalto/cs/intellij/notifications/BetaVersionWarning.java
OlliKiljunen: Perhaps not. The point here was that because a `BetaVersionWarning` object (in its initial state) can be fully described by the version it was given in the constructor (that is, two `BetaVersionWarning` objects that has the same `version` field do not differ each other in their initial states) it is nice to provide an access to that field.

src/main/java/fi/aalto/cs/intellij/common/Version.java
OlliKiljunen: A valid point. I'll add the checks. :+1:

src/test/java/fi/aalto/cs/intellij/notifications/NotificationsUtilityTest.java
OlliKiljunen: Yep, a good catch. :+1:

src/main/java/fi/aalto/cs/intellij/common/Resources.java
nikke234: If understood correctly and it's only used with a specific class loader, then one option would be for `resourceProvider` to be a static private, and the methods using it could be static as well? But your approach sounds good to me :+1:

OlliKiljunen:
I did some major refactoring based on the discussions with code reviewers on Wednesday. I am very thankful for your insights which, at least IMO, made this code so much better.

The most significant change is that previously null-returning methods no longer return nulls but instead throw exceptions.

I hope the code is now more readable and comprehensible. As for me, I am quite content with it now.

It should be noted that the implementations of `Resources` class and new `PropertyReader` class are structurally very similar. I even considered making both of them subclasses of one abstract and generic superclass that would handle the logic that are common to both of them. However, as these two classes are not semantically related to each other, that kind of approach would be questionable and certainly make code more difficult to understand.

I am sorry that I have had to do multiple major refactorings to this code during the review, as that had caused you a lot of review work. Optimistically, I am positive that the future PRs will require  substantially smaller amount of effort. :)

src/test/java/fi/aalto/cs/intellij/common/VersionTest.java
```diff
@@ -0,0 +1,134 @@
...
+  public void testCreateVersionWithNegativeMajorVersion() {
+    try {
+      new Version(-1, 4, 5);
+    } catch (IllegalArgumentException ex) {
+      return;
+    }
+    fail();
+  }
+
+  @Test
```
superseacat: could this approach be applicable when working with exceptions? 
```java
@Test(expected = NullPointerException.class)
public void whenExceptionThrown_thenExpectationSatisfied() { 
...
```

src/main/java/fi/aalto/cs/intellij/common/Resources.java
```diff
@@ -0,0 +1,83 @@
...
+   * @throws ResourceException If properties could not be read either because the resource was not
+   *                           found or it was badly formatted.
+   */
+  @NotNull
+  public Properties getProperties(@NotNull String resourceName) throws ResourceException {
+    Properties props = new Properties();
+    try (InputStream stream = getStream(resourceName)) {
+      props.load(stream);
+    } catch (IOException ex) {
+      throw new ResourceException(resourceName, "Resource could not be parsed to properties.", ex);
```
superseacat: :+1:

src/main/java/fi/aalto/cs/intellij/common/BuildInfo.java
superseacat: hm... @OlliKiljunen check [this](https://github.com/JetBrains/intellij-community/blob/e6de8d8c64e5e2fbacf43c9abd7bc2475427239e/platform/platform-impl/src/com/intellij/openapi/updateSettings/impl/PluginDownloader.java#L215-L217) out

OlliKiljunen: A nice discovery. :+1

I am, however, skeptical if we should rely on that. It is seems to be poorly documented and use a lot of "smart" logic to handle many different version string patterns, which actually is more a downside for us because we want to our version strings to match only a certain pattern decided by us. What is more, there seems to be no easy way to check e.g. "is the same major version", which is something that we might need (because major versions usually imply breaking changes so for example a plugin v2.0 does not necessarily support files created for v1.18 so perhaps a warning should be logged/shown in that situation).

Link to the implementation:
https://github.com/JetBrains/intellij-community/blob/master/platform/util-rt/src/com/intellij/util/text/VersionComparatorUtil.java

There they say:
> See TeamCity documentation on requirements comparison for formal description.

Only documentation I was able to find quickly:
https://www.jetbrains.com/help/teamcity/configuring-agent-requirements.html

There is also a "dumber" version comparer:
https://github.com/JetBrains/intellij-community/blob/92e4c7785bc46fcdda7e56ee0d808df287be98c7/platform/util/src/com/intellij/openapi/util/text/StringUtil.java#L2643

Based on their comments on that file,  it seems to be something that can be deprecated in the future.

src/main/java/fi/aalto/cs/intellij/activities/VersionCheckerActivity.java
OlliKiljunen: " how difficult your pseudocode test approach would be to implement"

I've understood that Mockito is specifically designed to so things like that so it shouldn't be difficult. I think @superseacat has done some research on this.

OlliKiljunen: > I can't imagine a "real world" use case

The most important 'use case' for this is certainly unit testing, and I wouldn't underrate its importance.

Based on my experience, this is a very customary way to write code, and I have understood that making code unit-testable this way is considered a good practice (even though there are other approaches, too, I guess). I did some google search about how to write testable code and at least the first article in the results seems to highly recommend using this pattern so I don't think it's outdated by any means. Link to that article: 
https://www.toptal.com/qa/how-to-write-testable-code-and-why-it-matters

src/main/java/fi/aalto/cs/intellij/activities/VersionCheckerActivity.java
OlliKiljunen: >> how difficult your pseudocode test approach would be to implement

> I've understood that Mockito is specifically designed to so things like that so it shouldn't be difficult. I think @superseacat has done some research on this.

And... I was wrong :P

Apparently Mockito can't mock static methods. See e.g. https://stackoverflow.com/questions/21105403/mocking-static-methods-with-mockito

src/test/java/fi/aalto/cs/intellij/common/VersionTest.java
OlliKiljunen: Yes if the properties of the exception (message, cause, etc.) don't have to be checked. So in this case that would be a good way to go. There is also ExpectedException rule in JUnit but to me, it seemed to add little value.

Junit 5 provides even more sophisticated solutions but I think testing exceptions this way is quite ok. The biggest issue is that `fail()` imay be easily forgotten.

src/main/java/fi/aalto/cs/intellij/activities/VersionCheckerActivity.java
nikke234: Good point on the importance of testing; I'm convinced of this approach now :+1:

src/test/java/fi/aalto/cs/intellij/common/BuildInfoTest.java
```diff
@@ -0,0 +1,58 @@
...
+    Properties props = new Properties();
+
+    try {
+      new BuildInfo(props);
+    } catch (PropertyException ex) {
+      assertEquals("The property key of exception should be one that is missing.",
+          BuildInfo.PropertyKeys.VERSION, ex.getPropertyKey());
+      return;
+    }
+    fail("Constructor should throw exception if properties do not contain necessary data.");
```
superseacat: :+1:

PR #52, opened 2020-01-21
Course configuration file parsing

nikke234:
# Description

Adds a `PluginSettings` application service that contains plugin-wide objects, such as the currently loaded course. Also adds `Course` and `Module` classes that contain information about courses and modules, and functionality for parsing a course configuration file.

# Testing

- [x] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant


superseacat:
> SonarCloud Quality Gate failed.
> 
> [Bug](https://sonarcloud.io/project/issues?id=Aalto-LeTech_intellij-plugin&pullRequest=52&resolved=false&types=BUG) [A](https://sonarcloud.io/project/issues?id=Aalto-LeTech_intellij-plugin&pullRequest=52&resolved=false&types=BUG) [0 Bugs](https://sonarcloud.io/project/issues?id=Aalto-LeTech_intellij-plugin&pullRequest=52&resolved=false&types=BUG)
> [Vulnerability](https://sonarcloud.io/project/issues?id=Aalto-LeTech_intellij-plugin&pullRequest=52&resolved=false&types=VULNERABILITY) [A](https://sonarcloud.io/project/issues?id=Aalto-LeTech_intellij-plugin&pullRequest=52&resolved=false&types=VULNERABILITY) [0 Vulnerabilities](https://sonarcloud.io/project/issues?id=Aalto-LeTech_intellij-plugin&pullRequest=52&resolved=false&types=VULNERABILITY) (and [Security Hotspot](https://sonarcloud.io/project/issues?id=Aalto-LeTech_intellij-plugin&pullRequest=52&resolved=false&types=SECURITY_HOTSPOT) [0 Security Hotspots](https://sonarcloud.io/project/issues?id=Aalto-LeTech_intellij-plugin&pullRequest=52&resolved=false&types=SECURITY_HOTSPOT) to review)
> [Code Smell](https://sonarcloud.io/project/issues?id=Aalto-LeTech_intellij-plugin&pullRequest=52&resolved=false&types=CODE_SMELL) [A](https://sonarcloud.io/project/issues?id=Aalto-LeTech_intellij-plugin&pullRequest=52&resolved=false&types=CODE_SMELL) [7 Code Smells](https://sonarcloud.io/project/issues?id=Aalto-LeTech_intellij-plugin&pullRequest=52&resolved=false&types=CODE_SMELL)
> 
> [](https://sonarcloud.io/component_measures?id=Aalto-LeTech_intellij-plugin&pullRequest=52&metric=new_coverage&view=list) [0.0% Coverage](https://sonarcloud.io/component_measures?id=Aalto-LeTech_intellij-plugin&pullRequest=52&metric=new_coverage&view=list)
> [](https://sonarcloud.io/component_measures?id=Aalto-LeTech_intellij-plugin&pullRequest=52&metric=new_duplicated_lines_density&view=list) [0.0% Duplication](https://sonarcloud.io/component_measures?id=Aalto-LeTech_intellij-plugin&pullRequest=52&metric=new_duplicated_lines_density&view=list)

could you please start with reviewing the sonar's complaints (they are, as usual, mostly cosmetic)

build.gradle
```diff
@@ -16,6 +16,7 @@ repositories {
 
 dependencies {
     testCompile group: 'junit', name: 'junit', version: '4.12'
+    implementation 'org.json:json:20190722'
```
superseacat: `implementation group: 'org.json', name: 'json', version: '20190722'` to preserve the style?

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
```diff
@@ -0,0 +1,106 @@
...
+   */
+  public static CourseInformation getInstance() {
+    return ServiceManager.getService(CourseInformation.class);
+  }
+
+  /**
+   * Parses the given course configuration file.
+   * @param pathToCourseConfig The path to the course configuration file.
+   */
+  public void parse(String pathToCourseConfig) {
```
superseacat: consider splitting this into multiple smaller methods

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
```diff
@@ -0,0 +1,106 @@
...
+    courseName = "";
+    moduleUrls = new HashMap<>();
+    FileReader file;
+    try {
+      file = new FileReader(pathToCourseConfig);
+    } catch (FileNotFoundException e) {
+      // TODO: notify the user of the error with a notification
+      logger.error("Failed to find course configuration file '{}'", pathToCourseConfig);
+      return;
+    }
```
superseacat: e.g. this becomes 
```java
File readFile(String/Path pathToCourseConfig){ ... }
```

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
superseacat: after longish discussions, the approach we currently gravitate to: normal functions with proper inputs and outputs and as little as possible class variables (only for the special cases)

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
```diff
@@ -0,0 +1,106 @@
...
+      logger.error("Invalid JSON in course configuration file: {}", e.getMessage());
+    }
+  }
+
+  /**
+   * Returns the name of the currently loaded course.
+   */
+  public String getCourseName() {
+    return courseName;
+  }
```
superseacat: this is not used atm (according to IDEA) and is kinda boilerplate, whereas, on the other hand, one might expect this to be present already, so I'd discuss this further

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * Returns the names of the code modules for the currently loaded course.
+   */
+  public List<String> getModuleNames() {
+    List<String> moduleNames = new ArrayList<String>();
```
superseacat:
```java 
= new ArrayList<>();
``` 
would be perfectly enough

src/main/java/fi/aalto/cs/intellij/ui/ModuleList.java
```diff
@@ -0,0 +1,72 @@
...
+      @Override
+      public void mouseClicked(MouseEvent e) {
+        JBList list = (JBList) e.getSource();
+        if (e.getClickCount() == 2) {
+          int index = list.getSelectedIndex();
+          String moduleName = (String) list.getModel().getElementAt(index);
+          try {
+            Desktop.getDesktop().browse(courseInformation.getModuleUrl(moduleName).toURI());
+          } catch (URISyntaxException | IOException ex) {
+            logger.error("Failed to open module url in browser");
```
superseacat: it might be a bit more informative if the stack trace gets printed as well:
```java
logger.error("Failed to open module url in browser", ex);
```

src/main/java/fi/aalto/cs/intellij/ui/ModuleList.java
```diff
@@ -0,0 +1,72 @@
...
+  private static final Logger logger = LoggerFactory
+      .getLogger(CourseInformation.class);
+
+  private JBList<String> modules;
+  private JPanel basePanel;
+
+  /**
+   * Initializes a panel with a list of modules.
+   */
+  public ModuleList() {
```
superseacat: would it be possible to extract some methods out of the constructor?

src/main/java/fi/aalto/cs/intellij/ui/ModuleList.java
```diff
@@ -0,0 +1,72 @@
...
+    modules = new JBList<String>();
+
+    CourseInformation courseInformation = CourseInformation.getInstance();
+
+    List<String> moduleNames = courseInformation.getModuleNames();
+    DefaultListModel<String> listModel = new DefaultListModel<String>();
+    for (String moduleName : moduleNames) {
+      listModel.addElement(moduleName);
+    }
+    modules.setModel(listModel);
```
superseacat: like this becomes method #1

superseacat:
general advice: run `checkstyleMain` and `checkstyleTest` gradle tasks

src/main/java/fi/aalto/cs/intellij/ui/ModuleList.java
```diff
@@ -0,0 +1,72 @@
...
+
+    modules.addMouseListener(new MouseAdapter() {
+      @Override
+      public void mouseClicked(MouseEvent e) {
+        JBList list = (JBList) e.getSource();
+        if (e.getClickCount() == 2) {
```
superseacat: could this be?
```java 
boolean isDoubleClick(MouseEvent e){ ... } 
```

src/main/java/fi/aalto/cs/intellij/ui/ModuleList.java
```diff
@@ -0,0 +1,72 @@
...
+    basePanel.setLayout(new GridLayoutManager(1, 1, new Insets(0, 0, 0, 0), -1, -1));
+    modules = new JBList<String>();
+
+    CourseInformation courseInformation = CourseInformation.getInstance();
+
+    List<String> moduleNames = courseInformation.getModuleNames();
+    DefaultListModel<String> listModel = new DefaultListModel<String>();
+    for (String moduleName : moduleNames) {
+      listModel.addElement(moduleName);
+    }
```
superseacat:
```java
courseInformation.getModuleNames().forEach(listModel::addElement);
``` 
?

src/main/java/fi/aalto/cs/intellij/ui/ModuleList.java
```diff
@@ -0,0 +1,72 @@
...
+   * Initializes a panel with a list of modules.
+   */
+  public ModuleList() {
+    basePanel = new JPanel();
+    basePanel.setLayout(new GridLayoutManager(1, 1, new Insets(0, 0, 0, 0), -1, -1));
+    modules = new JBList<String>();
+
+    CourseInformation courseInformation = CourseInformation.getInstance();
+
+    List<String> moduleNames = courseInformation.getModuleNames();
```
superseacat: would it make sense to check, if the `moduleNames` are empty?

src/main/java/fi/aalto/cs/intellij/ui/ModuleList.java
```diff
@@ -0,0 +1,72 @@
...
+          try {
+            Desktop.getDesktop().browse(courseInformation.getModuleUrl(moduleName).toURI());
+          } catch (URISyntaxException | IOException ex) {
+            logger.error("Failed to open module url in browser");
+          }
+        }
+      }
+    });
+
+    // GUI designer
```
superseacat: I understand it as "generated by GUI designer"?

src/main/java/fi/aalto/cs/intellij/actions/ModulesToolWindow.java
```diff
@@ -0,0 +1,21 @@
...
+import fi.aalto.cs.intellij.ui.ModuleList;
+import org.jetbrains.annotations.NotNull;
+
+public class ModulesToolWindow implements ToolWindowFactory {
+
+  @Override
+  public void createToolWindowContent(@NotNull Project project, @NotNull ToolWindow toolWindow) {
+    ModuleList moduleList = new ModuleList();
+    ContentFactory contentFactory = ContentFactory.SERVICE.getInstance();
+    Content content = contentFactory.createContent(moduleList.getBasePanel(), "", true);
```
superseacat: just curious what would happen, if to add a name instead of that empty `String` here

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
```diff
@@ -0,0 +1,106 @@
...
+   */
+  public void parse(String pathToCourseConfig) {
+    logger.info("Parsing the course configuration file '{}'", pathToCourseConfig);
+    courseName = "";
+    moduleUrls = new HashMap<>();
+    FileReader file;
+    try {
+      file = new FileReader(pathToCourseConfig);
+    } catch (FileNotFoundException e) {
+      // TODO: notify the user of the error with a notification
```
superseacat: I propose to agree, at what level we start to notify users (students) about failing of the underlying processes. for example, if the read of the downloaded (read provided) file fails, can they really do something?

build.gradle
nikke234: good idea, will do :+1:

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
nikke234: It's not used currently, but will almost certainly be used in the future

src/main/java/fi/aalto/cs/intellij/ui/ModuleList.java
nikke234: `JBList` (the GUI component used to display the modules) displays a text with something like "Nothing to show hree" if the list is empty, which I think is good enough for now

src/main/java/fi/aalto/cs/intellij/ui/ModuleList.java
nikke234: yes, i'll change it to that

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    JSONTokener tokenizer = new JSONTokener(file);
+    JSONObject jsonObj = new JSONObject(tokenizer);
+    try {
+      courseName = jsonObj.getString("name");
+
+      JSONObject modules = jsonObj.getJSONObject("modules");
+      Iterator<String> it = modules.keys();
+      while (it.hasNext()) {
```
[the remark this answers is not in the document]
nikke234: The good reason is that it doesn't compile, since an iterator doesn't implement `Iterable` :D but good point about abstracting the loop away here, i'll fix that

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
```diff
@@ -0,0 +1,106 @@
...
+  public String getCourseName() {
+    return courseName;
+  }
+
+  /**
+   * Returns the names of the code modules for the currently loaded course.
+   */
+  public List<String> getModuleNames() {
+    List<String> moduleNames = new ArrayList<String>();
+    moduleNames.addAll(moduleUrls.keySet());
```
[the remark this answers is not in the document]
nikke234: Good catch! It's certainly desireable to list the modules in the order that they are in the course config file, so I'll make the order deterministic here and add a mention of the order to the javadoc.

src/main/java/fi/aalto/cs/intellij/ui/ModuleList.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.List;
+import javax.swing.DefaultListModel;
+import javax.swing.JPanel;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ModuleList {
+  private static final Logger logger = LoggerFactory
+      .getLogger(CourseInformation.class);
```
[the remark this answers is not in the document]
nikke234: oops, nice catch

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.json.JSONTokener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CourseInformation {
```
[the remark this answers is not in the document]
nikke234: Yes, your understanding is spot on. The main issue I see with this approach is thread safety, which becomes a relevant issue once we allow the user to load another course.

I agree that the current approach doesn't quite "feel right". Your approach sounds good, although it raises the question of where the `currentlyOpenCourse` instance is stored, to which I suspect that the answer is some bigger plugin level singleton (`PluginSettings` or something similar), which sort of just defers the problem. But taking a more OO approach for the `CourseInformation` class and making it a "regular" class sounds good to me. Let's discuss this tomorrow before I refactor it.

Coupling the class with the IntelliJ service manager guarantees that only a single instance of the class exists (which is initialized on-demand in a thread safe manner) for the entire IntelliJ application, which is convenient for this class. However, you're correct in that all of this can be done with "plain old Java".

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
nikke234: Let's discuss this tomorrow :+1:

nikke234:
> general advice: run `checkstyleMain` and `checkstyleTest` gradle tasks

This passes both :wink:

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
nikke234: Good point, probably a better idea to let this method throw exceptions, although the question of where to catch them is somewhat complicated by the current service manager approach.

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
```diff
@@ -0,0 +1,106 @@
...
+    List<String> moduleNames = new ArrayList<String>();
+    moduleNames.addAll(moduleUrls.keySet());
+    return moduleNames;
+  }
+
+  /**
+   * Returns the URL where the code module is located.
+   */
+  public URL getModuleUrl(String moduleName) {
+    return moduleUrls.get(moduleName);
```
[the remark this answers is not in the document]
nikke234: Yes, especially if the other methods throw as well, good point

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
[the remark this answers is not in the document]
OlliKiljunen: Wait... if they are written, in JSON, as an object (or is it called dictionary or map or what but I mean this thing `{...}`) as key-value pairs, then the order does not matter in the JSON either. AFAIK, JSON strings `{"a": 1, "b": 2}` and `{"b": 2, "a": 1}` should be always considered equal, so making the order matter violates this. If the order should matter, either we can store modules in a list/array (`[...]` thing) or use alphabetical/lexicographical order.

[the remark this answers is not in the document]
nikke234: It would probably work in practice, but yes, if we want to use a specific order, it would be better to use an array and have the names of the modules be part of the module objects stored in the modules array

src/main/java/fi/aalto/cs/intellij/services/CourseInformation.java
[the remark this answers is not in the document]
OlliKiljunen: Something to check out: I found that, as Iterable<> is a functional interface, you can just write
```
Iterable<String> keys = modules::keys;
for (String key : keys) {
  ...
}
```
but know IntelliJ wants you to externally annotate `JSONObject.keys()` as `@NotNull` (which ain't probably an issue).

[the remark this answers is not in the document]
nikke234: neat! albeit somewhat uglified by intellij complaining, but switched to this, thx for the pointer

src/main/java/fi/aalto/cs/intellij/ui/ModuleList.java
nikke234: good idea, but not changed for now, since this is kind of a placeholder feature anyways

src/main/java/fi/aalto/cs/intellij/services/PluginSettings.java
```diff
@@ -0,0 +1,55 @@
...
+import fi.aalto.cs.intellij.notifications.CourseConfigurationError;
+import fi.aalto.cs.intellij.notifications.Notifier;
+
+import java.io.FileNotFoundException;
+
+import org.jetbrains.annotations.NotNull;
+
+public class PluginSettings {
+  @NotNull
+  private Course currentlyLoadedCourse;
```
OlliKiljunen: A possibility of a data race here? As a quick fix consider marking this `volatile`. We can think of more systematical synchronization approaches later. Or even more easily, just make it `final` until later.

src/main/java/fi/aalto/cs/intellij/common/Course.java
```diff
@@ -0,0 +1,219 @@
...
+public class Course {
+  @NotNull
+  private final String name;
+
+  @NotNull
+  private final List<Module> modules;
+
+  // Maps ids of required plugins to the names of required plugins.
+  @NotNull
+  private final Map<String, String> requiredPlugins;
```
OlliKiljunen: Plugins could also be represented as objects but IMHO, this is totally fine for now.

src/main/java/fi/aalto/cs/intellij/common/Course.java
```diff
@@ -0,0 +1,219 @@
...
+    return name;
+  }
+
+  /**
+   * Returns the names of all the modules in the course. If the course object is created with
+   * {@link Course#fromConfigurationFile}, then the module names are returned in the order
+   * in which they are listed in the course configuration file.
+   */
+  @NotNull
+  public List<String> getModuleNames() {
```
OlliKiljunen: Is this method more about presentation logic (and thus its place is perhaps not in a model class)? Could this class have a getter for a list of module objects instead? But this is a very minor issue (if an issue at all).

src/main/java/fi/aalto/cs/intellij/common/Course.java
```diff
@@ -0,0 +1,219 @@
...
+        throw new MalformedCourseConfigurationFileException(path,
+            "\"modules\" value should be an array of objects", null);
+      }
+      modules.add(getIndividualModule(moduleObject, path));
+    }
+    return modules;
+  }
+
+  @NotNull
+  private static Module getIndividualModule(@NotNull JSONObject moduleObject, @NotNull String path)
```
OlliKiljunen: Would a more logical place for this (and maybe the method above too) be in `Module` class?

src/main/java/fi/aalto/cs/intellij/services/PluginSettings.java
nikke234: yes there is, i'll make it final for now, and then once we add the possibility of loading a different course, we'll consider synchronization

src/main/java/fi/aalto/cs/intellij/common/Course.java
nikke234: Will keep this in mind if we ever decide to add more specific plugin information (more than just names and ids) to the course configuration file :+1:

src/main/java/fi/aalto/cs/intellij/common/Course.java
nikke234: Good point! I'll make this  `List<Module> getModules` and move the name-mapping logic to the UI code

src/main/java/fi/aalto/cs/intellij/actions/ModulesToolWindow.java
superseacat: I mean try it, there is some extra tab appearing and I was not sure, if it was intended :wink:

src/main/java/fi/aalto/cs/intellij/common/MalformedCourseConfigurationFileException.java
```diff
@@ -0,0 +1,21 @@
+package fi.aalto.cs.intellij.common;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class MalformedCourseConfigurationFileException extends Exception {
```
superseacat: the count of exceptions is growing, do they deserve a separate folder (like /common/exceptions) or something?

superseacat:
try "Optimize Imports" in addition to gradle check :stuck_out_tongue:

src/main/java/fi/aalto/cs/intellij/services/PluginSettings.java
```diff
@@ -0,0 +1,55 @@
...
+  @NonInjectable
+  public PluginSettings(@NotNull Notifier notifier) {
+    this.notifier = notifier;
+    try {
+      // Replace this with the correct path when testing with a local course configuration file.
+      currentlyLoadedCourse = Course.fromConfigurationFile("o1.json");
+    } catch (FileNotFoundException | MalformedCourseConfigurationFileException e) {
+      currentlyLoadedCourse = Course.createEmptyCourse();
+      notifier.notify(new CourseConfigurationError(e.getMessage()), null);
+    }
```
superseacat: this will only notify user about the issue via UI, is **not logging/printing stack trace** intended here?

src/main/java/fi/aalto/cs/intellij/common/MalformedCourseConfigurationFileException.java
```diff
@@ -0,0 +1,21 @@
...
+  @NotNull
+  private final String pathToConfigurationFile;
+
+  public MalformedCourseConfigurationFileException(@NotNull String pathToConfigurationFile,
+                                                   @NotNull String message,
+                                                   @Nullable Throwable cause) {
```
superseacat: this approach with each parameter on a new line reads very nice, I think I'll apply it to own code!

src/main/java/fi/aalto/cs/intellij/common/Course.java
```diff
@@ -0,0 +1,219 @@
...
+      throws MalformedCourseConfigurationFileException {
+    HashMap<String, String> requiredPlugins = new HashMap<>();
+    JSONObject requiredPluginsJson;
+    try {
+      requiredPluginsJson = jsonObject.getJSONObject("requiredPlugins");
+    } catch (JSONException ex) {
+      throw new MalformedCourseConfigurationFileException(path,
+          "Missing or malformed \"requiredPlugins\" key", ex);
+    }
+    Iterable<String> keys = requiredPluginsJson::keys;
```
superseacat: a bit puzzled here, if the first `try-catch` throws `JSONException exception` and `requiredPluginsJson` is still uninitialized. what would then happen, if `requiredPluginsJson::keys;` is called? 
to my understanding, the IDE should be ruby-red about `JSONObject requiredPluginsJson` - a method variable is not initialized... or smth like that

src/main/java/fi/aalto/cs/intellij/common/Course.java
```diff
@@ -0,0 +1,219 @@
...
+      throw new MalformedCourseConfigurationFileException(path,
+          "Module objects should contain \"name\" and \"url\" keys with string values", ex);
+    } catch (MalformedURLException ex) {
+      throw new MalformedCourseConfigurationFileException(path,
+          "Malformed URL in module object", ex);
+    }
+  }
+
+  @NotNull
+  private static Map<String, String> getCourseRequiredPlugins(@NotNull JSONObject jsonObject,
```
superseacat: this method might still be quite challenging to test, I suggest considering splitting it even more

src/main/java/fi/aalto/cs/intellij/common/Course.java
OlliKiljunen: The `catch` block throws a new exception so the execution never goes that far.

src/main/java/fi/aalto/cs/intellij/common/MalformedCourseConfigurationFileException.java
nikke234: This would definitely not be a bad idea

src/main/java/fi/aalto/cs/intellij/services/PluginSettings.java
nikke234: No harm in logging it as well, so adding that :+1:

OlliKiljunen:
I refactored the code a bit to make unit tests work platform independently.

The issue was that `testPluginSettingsWithMissingCourseConfigurationFile` asserted that the notification's message contains string "File or directory not found" to make sure the notification was caused by `FileNotFoungException` which, on Linux, gives that message. On Windows, the message is however different, so the test failed.

I changed the code so that `CourseConfigurationError` gets the exception object itself instead of itse message. So the message shown is whatever the `FileNotFoungException` provides, and we just check that the exception is an instance of that class.

PR #56, opened 2020-02-03
Show beta manual script

superseacat:
# Description

add simple manual test case scenario + a tiny update to notification texts

# Testing

- [x] All unit tests pass
- [x] I have tested manually

- [x] Yes
- [ ] Not relevant


src/main/java/fi/aalto/cs/intellij/notifications/BetaVersionWarning.java
```diff
@@ -18,7 +18,7 @@ public BetaVersionWarning(@NotNull Version version) {
     super("A+",
         "A+ Courses plugin is under development",
         "You are using version " + version + " of A+ Courses plugin, "
-            + "which is a pre-release version of the plugin and still under development. "
+            + "which is a pre-release version of the plugin and still under development. \n "
```
OlliKiljunen: Should platform independent line separator (`System.lineSeparator()`) used instead of LF? Does IntelliJ platform have any instructions on that?

superseacat: seems so, good point, I'll fix, thanks:
https://github.com/JetBrains/intellij-community/blob/e6abc7158fee3734d7ec74bbc53744798ea4bc78/platform/util/src/com/intellij/util/LineSeparator.java

PR #57, opened 2020-02-04
Install dependencies

OlliKiljunen:
# Description
Double click a module on the list and it and its dependencies will be added to your project. **Note! Module zip files are not downloaded via HTTP (we don't have these files on the server yet). Instead they are copied from directory `../modules/`, that is, a directory that is a sibling of the project's base dir (project that is open in the "guest" IDE).** To create these zip files you can run the following command

```
cd o1matsku
git checkout s2020
cd projects/given
find . -maxdepth 1 -mindepth 1 -type d -exec zip -r HERE_YOUR_PATH_TO_THAT_DIR/modules/{}.zip {} \;
```

TODO:
- [x] Write unit tests
- [x] Write documentation
- [x] Some refactoring
- [ ] Write manual testing doc

# Testing

- [x] All unit tests pass
- [x] I have tested manually

- [x] Yes
- [ ] Not relevant


src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ImportModuleAction.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Override
+  public void actionPerformed(@NotNull AnActionEvent e) {
+    CourseViewModel courseViewModel =
+        mainViewModelProvider.getMainViewModel(e.getProject()).courseViewModel.get();
+    if (courseViewModel != null) {
+      List<Module> modules = courseViewModel.getModules().getSelectedElements()
+          .stream()
+          .map(BaseViewModel::getModel)
+          .collect(Collectors.toList());
```
OlliKiljunen: This, I think, is the main benefit of using Presentation Model (View Model) pattern: the state of the UI (in this case, the selected modules) is store separately from UI and can be queried without referencing UI classes (and without polluting model classes with UI state information).

src/test/java/fi/aalto/cs/apluscourses/presentation/module/ModuleListElementViewModelTest.java
```diff
@@ -0,0 +1,83 @@
...
+    assertEquals("Loaded", moduleViewModel.getStatus());
+    assertEquals(TextAttribute.WEIGHT_BOLD, moduleViewModel.getFontWeight(), delta);
+
+    module.stateMonitor.set(Module.WAITING_FOR_DEPS);
+    assertEquals("Waiting for dependencies...", moduleViewModel.getStatus());
+    assertEquals(TextAttribute.WEIGHT_BOLD, moduleViewModel.getFontWeight(), delta);
+
+    module.stateMonitor.set(Module.INSTALLED);
+    assertEquals("Installed", moduleViewModel.getStatus());
+    assertEquals(TextAttribute.WEIGHT_BOLD, moduleViewModel.getFontWeight(), delta);
```
OlliKiljunen: Another benefit of Presentation Model (= view model) pattern: being able to unit test presentation logic (such as here) without dealing with UI classes.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJCourseFactory.java
```diff
@@ -0,0 +1,33 @@
...
+import com.intellij.openapi.project.Project;
+import fi.aalto.cs.apluscourses.model.Course;
+import fi.aalto.cs.apluscourses.model.CourseFactory;
+import fi.aalto.cs.apluscourses.model.Module;
+import java.net.URL;
+import java.util.List;
+import java.util.Map;
+import org.jetbrains.annotations.NotNull;
+
+public class IntelliJCourseFactory implements CourseFactory {
```
superseacat: as it is capable of creating both a **course** and a **module** should it have a bit more descriptive (e.g. IntelliJCourseModuleFactory) or more generic (like IntelliJModelFactory)?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJCourseFactory.java
```diff
@@ -0,0 +1,33 @@
...
+
+  @Override
+  public Course createCourse(@NotNull String name,
+                             @NotNull List<Module> modules,
+                             @NotNull Map<String, String> requiredPlugins) {
+    return new IntelliJCourse(name, modules, requiredPlugins, project);
+  }
+
+  @Override
+  public Module createModule(@NotNull String name, @NotNull URL url) {
```
superseacat: to consider: does it make sense to create a module that is a part of the **project**, but is not tied to a **course**? In other words: [Project] 1 - * [Course] - * [Module]?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJCourseFactory.java
```diff
@@ -0,0 +1,33 @@
...
+  private final Project project;
+
+  public IntelliJCourseFactory(@NotNull Project project) {
+    this.project = project;
+  }
+
+  @Override
+  public Course createCourse(@NotNull String name,
+                             @NotNull List<Module> modules,
+                             @NotNull Map<String, String> requiredPlugins) {
```
superseacat: to consider: would it make sense to wrap `Map<String, String>` with a custom container class or are we still OK with tossing it around?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
```diff
@@ -0,0 +1,112 @@
...
+import org.w3c.dom.Node;
+import org.xml.sax.SAXException;
+
+class IntelliJModule extends Module {
+
+  @NotNull
+  private final Project project;
+
+  private static final String DEPENDENCY_NAMES =
+      "/module/component/orderEntry[@type='module']/@module-name";
```
superseacat: it might be easier if there would be a comment/link to what this is

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
```diff
@@ -0,0 +1,112 @@
...
+    return FileUtilRt.createTempFile(getName(), ".zip");
+  }
+
+  protected void extractZip(File file) throws IOException {
+    new ZipFile(file).extractAll(getBasePath());
+  }
+
+  protected void fetchZipTo(File file) throws IOException {
+    // Call HTTP client
+    throw new UnsupportedOperationException();
```
superseacat: I like the design approach, a lot :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJCourseFactory.java
OlliKiljunen: Agreed! ModelFactory and its subclass IntelliJModelFactory seem more suitable names here.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
```diff
@@ -0,0 +1,112 @@
...
+          .map(Node::getTextContent)
+          .collect(Collectors.toList());
+    } catch (IOException | SAXException e) {
+      throw new ModuleLoadException(this, e);
+    }
+  }
+
+  @Override
+  public void fetch() throws IOException {
+    File file = createTempFile();
```
superseacat: how about `File tempZipFile = createTempFile();` , otherwise made me re-read the block several times?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJCourseFactory.java
OlliKiljunen: I see your point. The problem is the order in which objects are created. Current course config file parser creates first `Module` objects and gives a list of them to the constructor of `Course`. Thus, `Module`s don't have links to `Course`. (We could add that link so that the `Course` would register itself calling `aModule.setCourse(this)` or something a like but currently there is no need for that -- besides this issue.) On the other hand, `Project` as a class of IntelliJ SDK is of course available already when `Module`s are created, so it is easy to give `Module` a link to `Project`.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJCourseFactory.java
OlliKiljunen: It certainly would. I didn't however refactor much `Module` class in this PR, so I left it unchanged.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
OlliKiljunen: I'll add that :+1: Perhaps constant name should also mention that it is XPath string.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
OlliKiljunen: A fair point, I'll change that.

src/main/java/fi/aalto/cs/apluscourses/intellij/activities/InitializationActivity.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Override
+  public void runActivity(@NotNull Project project) {
+    Course course;
+    try {
+      course = Course.fromResource(PluginSettings.COURSE_CONFIGURATION_FILE_PATH,
+          new IntelliJCourseFactory(project));
+    } catch (ResourceException | MalformedCourseConfigurationFileException e) {
+      course = null;
+      logger.info("Error occurred while trying to parse a course configuration file", e);
```
superseacat: `logger.`**error**`("Error occurred while trying to parse a course configuration file", e);` ?

OlliKiljunen: At this point, I agree: it is an error of our software because the course config file is currently bundled with the software. Once we start downloading the file from the web, we should change to consider it as user data, and handle malformation exceptions accordingly. I fix this now following your suggestion but I expect this whole class to change a lot very soon.

src/main/java/fi/aalto/cs/apluscourses/intellij/activities/InitializationActivity.java
```diff
@@ -0,0 +1,54 @@
...
+      notifier.notify(new CourseConfigurationError(e), null);
+    }
+    if (course != null) {
+      PluginSettings.getInstance()
+          .getMainViewModel(project).courseViewModel.set(new CourseViewModel(course));
+      ActionUtil.launch(RequiredPluginsCheckerAction.ACTION_ID,
+          new ExtendedDataContext().withProject(project));
+    }
+  }
```
superseacat: [Nit:](https://github.com/google/eng-practices/blob/master/review/reviewer/standard.md#mentoring)
```java
  @Override
  public void runActivity(@NotNull Project project) {
    Course course;
    try {
      course = Course.fromResource(PluginSettings.COURSE_CONFIGURATION_FILE_PATH,
          new IntelliJCourseFactory(project));
      PluginSettings.getInstance()
          .getMainViewModel(project).courseViewModel.set(new CourseViewModel(course));
      ActionUtil.launch(RequiredPluginsCheckerAction.ACTION_ID,
          new ExtendedDataContext().withProject(project));
    } catch (ResourceException | MalformedCourseConfigurationFileException e) {
      logger.info("Error occurred while trying to parse a course configuration file", e);
      notifier.notify(new CourseConfigurationError(e), null);
    }
  }
```
as an alternative...  then there is not `Course` initialization inside the `catch` block, however, then mode logix gets into the `try`

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ActionUtil.java
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * Returns an {@link ActionListener} that launches {@link AnAction} corresponding to the given
+   * id each time {@code actionPerformed()} is called.
+   *
+   * @param actionId Id of the action.
+   * @param source A {@link Component} that gives the {@link DataContext} for the action.
+   * @return An {@link ActionListener}
+   */
+  public static ActionListener onEventLauncher(String actionId, Component source) {
```
superseacat: smart

src/main/java/fi/aalto/cs/apluscourses/intellij/activities/InitializationActivity.java
OlliKiljunen: That's definitely better. If we want to keep `try` block small, we can also add `return` to `catch`. 

```java
  @Override
  public void runActivity(@NotNull Project project) {
    Course course;
    try {
      course = Course.fromResource(PluginSettings.COURSE_CONFIGURATION_FILE_PATH,
          new IntelliJCourseFactory(project));
    } catch (ResourceException | MalformedCourseConfigurationFileException e) {
      logger.info("Error occurred while trying to parse a course configuration file", e);
      notifier.notify(new CourseConfigurationError(e), null);
      return;
    }
    PluginSettings.getInstance()
        .getMainViewModel(project).courseViewModel.set(new CourseViewModel(course));
    ActionUtil.launch(RequiredPluginsCheckerAction.ACTION_ID,
        new ExtendedDataContext().withProject(project));
  }
```

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ActionUtil.java
OlliKiljunen: Let's consider the expanded form (without lambda):
```java
  public static ActionListener onEventLauncher(String actionId, Component source) {
    return new ActionListener() {
      @Override
      public void actionPerformed(ActionEvent actionEvent) {
        launch(actionId, DataManager.getInstance().getDataContext(source));
      }
    };
  }
```
So this creates a new `java.awt.event.ActionListener` object that just calls `launch()` function above on its `actionPerformed()` method. (`ActionListener` is used e.g. by many Swing classes as callback interface on certain UI events, like button clicked or such.)

src/main/java/fi/aalto/cs/apluscourses/ui/base/ComponentUtil.java
```diff
@@ -0,0 +1,38 @@
...
+import java.awt.Font;
+import java.awt.font.TextAttribute;
+import java.util.Collections;
+import java.util.Map;
+import javax.swing.JComponent;
+
+/**
+ * A helper class for {@link JComponent} objects.
+ */
+public class ComponentUtil {
```
nikke234: This class could maybe go one one directory "upwards" and reside in `ui`? (I could also be misunderstanding how `base` is used here)

OlliKiljunen: A good point. With "base", I just meant non-domain-specific base classes that domain-specific classes can inherit, so yeah, this should be in a different package.

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/ImportModuleActionTest.java
```diff
@@ -0,0 +1,84 @@
...
+    doReturn(presentation).when(e).getPresentation();
+    doReturn(project).when(e).getProject();
+
+    action.update(e);
+
+    assertFalse(presentation.isEnabledAndVisible());
+
+    mainViewModel.courseViewModel.get().getModules().getSelectionModel().addSelectionInterval(1,2);
+
+    action.update(e);
```
superseacat: would it make sense to have separate tests for ```action.update(e); ``` and ```action.actionPerformed(e); ```?

src/main/java/fi/aalto/cs/apluscourses/model/ModuleSource.java
```diff
@@ -0,0 +1,8 @@
+package fi.aalto.cs.apluscourses.model;
+
+import org.jetbrains.annotations.Nullable;
+
+public interface ModuleSource {
```
nikke234: A short explanatory comment here would be helpful (also since `getModule` isn't documented in `Course`)

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/ImportModuleActionTest.java
OlliKiljunen: Absolutely. :+1:  It was my laziness to put them together. I'll separate them.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ActionUtil.java
```diff
@@ -0,0 +1,43 @@
...
+
+public class ActionUtil {
+
+  /**
+   * Launches {@link AnAction}.
+   *
+   * @param actionId    Id of the action.
+   * @param dataContext A {@link DataContext} for the action.
+   */
+  public static void launch(String actionId, DataContext dataContext) {
```
nikke234: These could be annotated with `@NotNull`?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJCourse.java
```diff
@@ -0,0 +1,27 @@
...
+import org.jetbrains.annotations.NotNull;
+
+class IntelliJCourse extends Course {
+  @NotNull
+  private final Project project;
+
+  IntelliJCourse(@NotNull String name,
+                        @NotNull List<Module> modules,
+                        @NotNull Map<String, String> requiredPlugins,
+                        @NotNull Project project) {
```
nikke234: Nit: I think these lines are indented one level too much :smile:

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -182,7 +193,9 @@ private static String getCourseName(@NotNull JSONObject jsonObject, @NotNull Str
   }
 
   @NotNull
-  private static List<Module> getCourseModules(JSONObject jsonObject, String path)
+  private static List<Module> getCourseModules(JSONObject jsonObject,
+                                               String path,
+                                               @NotNull CourseFactory factory)
```
nikke234: These are leftovers from me but could also add `@NotNull` in this pr

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.concurrent.ConcurrentMap;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class PluginSettings implements MainViewModelProvider {
+
+  public static final String COURSE_CONFIGURATION_FILE_PATH = "o1_course_config.json";
+
+  @NotNull
+  private final ConcurrentMap<Project, MainViewModel> mainViewModels = new ConcurrentHashMap<>();
```
superseacat: can you hint me with the case where this map will store `MainViewModels` that are not related to the current `project`?

src/main/java/fi/aalto/cs/apluscourses/intellij/toolwindows/ExercisesToolWindowFactory.java
```diff
@@ -0,0 +1,14 @@
...
+import com.intellij.openapi.project.Project;
+import fi.aalto.cs.apluscourses.ui.ExerciseList;
+import javax.swing.JComponent;
+import org.jetbrains.annotations.NotNull;
+
+public class ExercisesToolWindowFactory extends BaseToolWindowFactory {
+
+  @Override
+  protected JComponent createToolWindowContentInternal(@NotNull Project project) {
+    return new ExerciseList().getBasePanel();
```
superseacat: this is slightly smaller than the `ModulesToolWindowFactory` because it is undone?

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
OlliKiljunen: If there are multiple projects open in IntelliJ, each of them have a separate MainViewModel (because each of the projects are mapped to some course separately so there are separate domain model objects too)

OlliKiljunen: There actually should be some cleanup: if a project is closed, its `MainViewModel` should be removed from the map, so it can be GC'ed. One possibility would be use `WeakHashMap` again, but I am not sure of the lifetime of `Project` objects. (It is even a bit risky to use them as hash keys in general -- I should study if that is something that the designer of `Project` have thought should be possible. I had this issue in mind but forgot it. Thanks for noting!)

OlliKiljunen: So the better way to implement cleanup is probably to listen to "project closed" event or something like that and then remove no-more-needed MainViewModels then. I'll check this.

src/main/java/fi/aalto/cs/apluscourses/intellij/toolwindows/ExercisesToolWindowFactory.java
OlliKiljunen: Yes (if I got your point right). We don't have Exercises tool window yet (just the placeholder) so this method just renders it. There are no viewmodel bindings, toolbar setups and such things here.

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
```diff
@@ -0,0 +1,103 @@
...
+   * @return Names of the dependencies, as a {@link List}.
+   * @throws ModuleLoadException If dependencies could not be read.
+   */
+  @NotNull
+  public List<String> getDependencies() throws ModuleLoadException {
+    return Collections.emptyList();
+  }
+  
+  public void fetch() throws IOException {
+    // Default implementation: do nothing
```
superseacat: is there a reason to leave it empty instead of `throw new UnsupportedOperationException();`?

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
```diff
@@ -0,0 +1,103 @@
...
+   * @throws org.json.JSONException If the jsonObject doesn't contain "name" and "url" keys with
+   *                                string values.
+   */
+  @NotNull
+  public static Module fromJsonObject(@NotNull JSONObject jsonObject, CourseFactory factory)
+      throws MalformedURLException {
+    String name = jsonObject.getString("name");
+    URL url = new URL(jsonObject.getString("url"));
+    return factory.createModule(name, url);
+  }
```
nikke234: Should factory be `NotNull`? Could also maybe add a javadoc @param for factory.

src/main/java/fi/aalto/cs/apluscourses/model/ModuleInstallerImpl.java
```diff
@@ -0,0 +1,143 @@
+package fi.aalto.cs.apluscourses.model;
+
+import fi.aalto.cs.apluscourses.utils.async.TaskManager;
+import java.io.IOException;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import org.jetbrains.annotations.NotNull;
+
+public class ModuleInstallerImpl<T> implements ModuleInstaller {
```
superseacat: could the IJ's [`ModuleManager`](https://github.com/JetBrains/intellij-community/blob/002b9b884ce17d5894eff6b00e9affb89f324ee0/platform/projectModel-impl/src/com/intellij/openapi/module/impl/ModuleManagerImpl.java) be of any help here?

OlliKiljunen: ModuleManager is finally called: see `IntelliJModule.Loader.load()` but before that we must download (fetch) the zip, unzip, read dependencies from IML file (and start their installations). So *loading* module to is just a small part of installation.

OlliKiljunen: Please note that I just changed this (outer) class to fix a bug (it became also simpler). I push that now if you don't mind.

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
OlliKiljunen: It seems that project can be used as hash key, see [here](https://upsource.jetbrains.com/idea-ce/file/idea-ce-e2d423f42b53eafd5c625b93149a0ebc7e49756c/platform/platform-impl/src/com/intellij/openapi/project/impl/DefaultProject.java?nav=3796:3826:focused&line=157&preview=false). (Some nerd humour there also a few lines above :D)

I consider using a WeakHahMap still too risky and instead, just do manual cleanup when project closes.

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -52,6 +56,12 @@ public static Course createEmptyCourse() {
     return new Course("", new ArrayList<>(), new HashMap<>());
```
superseacat: Nit:
I understand this but suggest to consider having a separate `CourseTestHelper` class to work with

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
OlliKiljunen: The honest reason seems to be my laziness again (avoid some overrides in unit tests). I'll change that.

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
OlliKiljunen: This method became actually obsolete in this PR. I'll remove it altogether. It wasn't originally intended (by Nikolas) for testing.

OlliKiljunen:
Bad news: I had to change `ModuleInstallerImpl` class a few times because my original implementation was a bit inconsistent in some aspects. Hopefully, you didn't review it before my changes.

Good news: the new version of the class is much simpler (and better in all aspects) than the earlier one.

src/main/java/fi/aalto/cs/apluscourses/utils/Resources.java
```diff
@@ -55,8 +55,14 @@ public Properties getProperties(@NotNull String resourceName) throws ResourceExc
...
 
+  /**
+   * Returns a resource as a stream.
+   * @param resourceName Name of the resource.
+   * @return An {@link InputStream} to the given resource.
+   * @throws ResourceException If the resource could not be found.
+   */
   @NotNull
-  private InputStream getStream(@NotNull String resourceName) throws ResourceException {
+  public InputStream getStream(@NotNull String resourceName) throws ResourceException {
```
superseacat: why this change occurs?

src/main/resources/META-INF/plugin.xml
```diff
@@ -20,24 +20,35 @@
...
-                    factoryClass="fi.aalto.cs.intellij.actions.ModulesToolWindow" />
+                    factoryClass="fi.aalto.cs.apluscourses.intellij.toolwindows.ModulesToolWindowFactory" />
         <toolWindow id="Exercises"
                     anchor="right"
-                    factoryClass="fi.aalto.cs.intellij.actions.ExercisesToolWindow" />
+                    factoryClass="fi.aalto.cs.apluscourses.intellij.toolwindows.ExercisesToolWindowFactory" />
     </extensions>
 
     <actions>
         <group id="A+.Menu" text="A+" description="A+ Menu">
```
superseacat: Nit: it would probably make sense to unify the approach to assigning IDs to `groups` (and `actions`)

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ImportModuleAction.java
```diff
@@ -0,0 +1,66 @@
...
+import fi.aalto.cs.apluscourses.presentation.CourseViewModel;
+import fi.aalto.cs.apluscourses.presentation.base.BaseViewModel;
+import fi.aalto.cs.apluscourses.utils.async.SimpleAsyncTaskManager;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.jetbrains.annotations.NotNull;
+
+public class ImportModuleAction extends AnAction {
+
+  public static final String ACTION_ID =
```
superseacat:
```java
public static final String ACTION_ID = ImportModuleAction.class.getCanonicalName();
```
produces exactly the same result ("fi.aalto.cs.apluscourses.intellij.actions.ImportModuleAction"), could this work?

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/RequiredPluginsCheckerAction.java
```diff
@@ -1,40 +1,40 @@
...
 
 /**
- * A startup activity that checks and hints on missing or disabled required for the course plugins.
+ * An action that checks and hints on missing or disabled required for the course plugins.
  */
-public class RequiredPluginsCheckerActivity implements StartupActivity {
+public class RequiredPluginsCheckerAction extends AnAction {
+
+  public static final String ACTION_ID =
+      "fi.aalto.cs.apluscourses.intellij.actions.RequiredPluginsCheckerAction";
```
superseacat: https://github.com/Aalto-LeTech/intellij-plugin/pull/57/files#r379591397

src/main/java/fi/aalto/cs/apluscourses/presentation/base/BaseListViewModel.java
```diff
@@ -0,0 +1,114 @@
...
+public class BaseListViewModel<E extends ListElementViewModel<?>> extends AbstractListModel<E>
+    implements SelectableListModel<E> {
+
+  @NotNull
+  private final transient ListSelectionModel selectionModel;
+
+  // Not actually something that should be transient but we are not doing serialization and Sonar
+  // requires fields be marked transient if they are not serializable.
+  @NotNull
+  private final transient List<E> elements;
```
superseacat: does this apply to all the `transient`-marked cases within the CL?

src/main/java/fi/aalto/cs/apluscourses/model/ModuleLoadException.java
```diff
@@ -0,0 +1,19 @@
+package fi.aalto.cs.apluscourses.model;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class ModuleLoadException extends Exception {
+  @NotNull
+  private final transient Module module;
```
superseacat: https://github.com/Aalto-LeTech/intellij-plugin/pull/57/files#r379608161 ?

src/main/java/fi/aalto/cs/apluscourses/utils/DomUtil.java
```diff
@@ -0,0 +1,121 @@
...
+    xPathCache = new ConcurrentHashMap<>();
+
+    xPathFactory = XPathFactory.newInstance();
+
+    documentBuilderFactory = DocumentBuilderFactory.newInstance();
+    try {
+      documentBuilderFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+      documentBuilderFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+    } catch (IllegalArgumentException e) {
+      logger.warn("Could not set XXE restrictions for DOM tools because the platform does not "
```
superseacat: [XXE defnition](https://www.acunetix.com/blog/articles/xml-external-entity-xxe-vulnerabilities/) :smiley_cat:

src/main/java/fi/aalto/cs/apluscourses/utils/CommonUtil.java
```diff
@@ -0,0 +1,32 @@
...
+import org.jetbrains.annotations.NotNull;
+
+public class CommonUtil {
+
+  private CommonUtil() {
+
+  }
+
+  /**
+   * Creates a new {@link List} of type {@link T} and size {@code length} populated with objects
```
superseacat: `{@link <T>}` this syntax should work better in JavaDoc

src/test/java/fi/aalto/cs/apluscourses/presentation/base/BaseListViewModelTest.java
```diff
@@ -0,0 +1,96 @@
...
+    List<Object> models = new ArrayList<>();
+    models.add(model1);
+    models.add(model2);
+    models.add(model3);
+
+    ListElementViewModel<?> listElementViewModel1 = new ListElementViewModel<>(model1);
+    ListElementViewModel<?> listElementViewModel2 = new ListElementViewModel<>(model2);
+    ListElementViewModel<?> listElementViewModel3 = new ListElementViewModel<>(model3);
+
+    Map<Object, ListElementViewModel<?>> elementMap = new HashMap<>();
```
superseacat: there is this clause when describing a HashMap:

> [This class makes no guarantees as to the order of the map; in particular, it does not guarantee that the order will remain constant over time.](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html)

could this "no guarantee" have an effect here, when accessing elements in assertions?

src/main/java/fi/aalto/cs/apluscourses/utils/StateMonitor.java
```diff
@@ -0,0 +1,125 @@
...
+  private boolean isError(int someState) {
+    return someState <= ERROR;
+  }
+
+  /**
+   * Returns the current state.
+   *
+   * @return The current state.
+   */
+  public int get() {
```
superseacat: I'd propose to extend the name of the method to point to what is being fetched, like `getStateCode();` or something alike; then [this comparison](https://github.com/Aalto-LeTech/intellij-plugin/blob/install-dependencies/src/main/java/fi/aalto/cs/apluscourses/model/Module.java#L101) would read cleaner.

src/main/java/fi/aalto/cs/apluscourses/utils/StateMonitor.java
```diff
@@ -0,0 +1,125 @@
...
+ *   All the states less than or equal to {@code ERROR} are considered error states.</li>
+ *   <li>Normally, the state can only increase, that is, on a single change, a new state can never
+ *   be less than the old state, except when either the old state or the new state is an error
+ *   state.</li>
+ * </ul>
+ *
+ * <p>The {@link StateListener}, given in the constructor is notified of the changes in state.</p>
+ */
+public class StateMonitor {
+  private int state = INITIAL;
```
superseacat: Nit: maybe these "state"s could be "stateCode"s?

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ActionUtil.java
superseacat: I've just probably misunderstood the method's purpose, as expected something related to an `Event`, and in that case, returning a `new Listener` on event launch is a bit confusing. So how then the name `onEventLauncher(...)`, is related to the `Event` then, as it seems to return a `Listener`?

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
superseacat: I meant a bit different thing, but it's good there is synergy!

so the point is, that I don't exactly understand how the projects work in the IJ. For example, I have our "intellij-plugin" opened in two IDEs at the same time (it's the same codebase), like here:
projects_in_IDE
does it make them separate projects then?

on the other hand, if I go to the list of projects in ToolBox:
projects?
then there is only one "intellij-plugin"

so what I was asking is, what `MainViewModels` will be associated with Projects in those cases?
also, in the first picture, are they modules or projects, sorry, help me out of this :laughing: ?

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
```diff
@@ -0,0 +1,103 @@
...
+   * </pre>
+   * @param jsonObject The JSON object containing information about a single module.
+   * @return A module constructed from the given JSON object.
+   * @throws MalformedURLException  If the URL of the module is malformed.
+   * @throws org.json.JSONException If the jsonObject doesn't contain "name" and "url" keys with
+   *                                string values.
+   */
+  @NotNull
+  public static Module fromJsonObject(@NotNull JSONObject jsonObject, ModelFactory factory)
+      throws MalformedURLException {
```
superseacat: Nit:
this is a bit controversial, as `JSONException` just happens to extend `RuntimeException`, but maybe it will improve the readability, if we at least add it to the `throw` clause like `throws MalformedURLException, JSONException`, thus, it does not really propagate further (as in [here](https://github.com/Aalto-LeTech/intellij-plugin/blob/e836952634484c6a0816cba4318a5a9c6d19d0c4/src/main/java/fi/aalto/cs/apluscourses/model/Course.java#L214), you just need to know, that the `JSONException` is coming). But this is tested, so I don't know...

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -197,7 +210,7 @@ private static String getCourseName(@NotNull JSONObject jsonObject, @NotNull Str
     for (int i = 0; i < modulesJsonArray.length(); ++i) {
       try {
         JSONObject moduleObject = modulesJsonArray.getJSONObject(i);
-        modules.add(Module.fromJsonObject(moduleObject));
+        modules.add(Module.fromJsonObject(moduleObject, factory));
       } catch (JSONException ex) {
```
superseacat: Nit: a tiny styling notice, sometimes we use `e` for `Exception` and sometimes `ex`, shall we unify (I suggest to `ex`, as we have `e` for `events` in other places)

src/main/java/fi/aalto/cs/apluscourses/utils/StateMonitor.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  @FunctionalInterface
+  public interface StateListener {
+    void onStateChanged();
+  }
+
+  public static class ErrorStateException extends Exception {
+
+  }
```
superseacat: is it intended to be empty (unused too, according to the IDE)?

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
```diff
@@ -0,0 +1,103 @@
...
+public class Module {
+
+  public static final int ERROR = StateMonitor.ERROR;
+  public static final int NOT_INSTALLED = StateMonitor.INITIAL;
+  public static final int FETCHING = NOT_INSTALLED + 1;
+  public static final int FETCHED = FETCHING + 1;
+  public static final int LOADING = FETCHED + 1;
+  public static final int LOADED = LOADING + 1;
+  public static final int WAITING_FOR_DEPS = LOADED + 1;
+  public static final int INSTALLED = WAITING_FOR_DEPS + 1;
```
superseacat: would it be of any help?
```java
  public enum States {
    ERROR(StateMonitor.ERROR),
    NOT_INSTALLED(StateMonitor.INITIAL),
    FETCHING(NOT_INSTALLED.code + 1);
    ...

    public final int code;

    private States(int code) {
      this.code = code;
    }
  }
```

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
```diff
@@ -0,0 +1,103 @@
...
+  public List<String> getDependencies() throws ModuleLoadException {
+    return Collections.emptyList();
+  }
+  
+  public void fetch() throws IOException {
+    // Default implementation: do nothing
+  }
+
+  public void load() throws ModuleLoadException {
+    // Default implementation: do nothing
```
superseacat: here too :pencil2:

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @NotNull
+  public URL getUrl() {
+    return url;
+  }
+
+  /**
+   * Returns the names of the modules on which this module is dependent.  This method should not be
+   * called unless the module is in FETCHED state or further.
```
superseacat: could there be some kind of check if the `state >= FETCHED` (maybe inside the method), or somehow... currently it feels like an assumption, that is hard to know about until the debug stage.

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
```diff
@@ -0,0 +1,103 @@
...
+   *
+   * @return Names of the dependencies, as a {@link List}.
+   * @throws ModuleLoadException If dependencies could not be read.
+   */
+  @NotNull
+  public List<String> getDependencies() throws ModuleLoadException {
+    return Collections.emptyList();
```
superseacat: this `return Collections.emptyList();` is because it still awaits implementation, right?

src/main/java/fi/aalto/cs/apluscourses/model/ModuleInstallerImpl.java
```diff
@@ -0,0 +1,133 @@
...
+        module.stateMonitor.set(Module.INSTALLED);
+      } else {
+        module.stateMonitor.waitUntil(Module.INSTALLED);
+      }
+    }
+
+    private List<Module> getDependencies() throws ModuleLoadException {
+      try {
+        if (dependencies == null) {
+          dependencies = module.getDependencies()
```
superseacat: `module.getDependencies()` currently returns an empty `List<String>`?

src/main/java/fi/aalto/cs/apluscourses/model/ModuleInstallerImpl.java
```diff
@@ -0,0 +1,133 @@
...
+              .stream()
+              .map(moduleSource::getModule)
+              .map(Objects::requireNonNull)
+              .collect(Collectors.toList());
+        }
+        return dependencies;
+      } catch (NullPointerException e) {
+        throw new ModuleLoadException(module, e);
```
superseacat: Nit: could the throwing of the `ModuleLoadException` occur earlier and avoid handling the NPE?

src/main/java/fi/aalto/cs/apluscourses/model/ModuleInstallerImpl.java
```diff
@@ -0,0 +1,133 @@
...
+      this.module = module;
+    }
+    
+    public void doIt() {
+      try {
+        fetch();
+        load();
+        waitForDependencies();
+      } catch (IOException | ModuleLoadException e) {
+        module.stateMonitor.set(Module.ERROR);
```
superseacat: are you sure we do not need those `Exceptions` themselves somehow?

src/main/java/fi/aalto/cs/apluscourses/presentation/module/ModuleListElementViewModel.java
```diff
@@ -0,0 +1,57 @@
...
+   */
+  public String getStatus() {
+    switch (getModel().stateMonitor.get()) {
+      case Module.NOT_INSTALLED:
+      case Module.FETCHED:
+        return "Not installed";
+      case Module.FETCHING:
+        return "Downloading...";
+      case Module.LOADING:
+        return "Installing...";
```
superseacat: Nit: could this be unified ("FETCHING -> "Fetching" or DOWNLOADING -> "Downloading")?

src/main/java/fi/aalto/cs/apluscourses/presentation/module/ModuleListElementViewModel.java
```diff
@@ -0,0 +1,57 @@
...
+      case Module.LOADED:
+        return "Loaded";
+      case Module.WAITING_FOR_DEPS:
+        return "Waiting for dependencies...";
+      case Module.INSTALLED:
+        return "Installed";
+      default:
+        return "Error";
+    }
+  }
```
superseacat: this `switch` is so cool! :astonished:

src/main/java/fi/aalto/cs/apluscourses/utils/DomUtil.java
```diff
@@ -0,0 +1,121 @@
...
+   */
+  @NotNull
+  public static List<Node> getNodesFromXPath(@NotNull String xpath, @NotNull Node node) {
+    NodeList nodeList;
+    try {
+      nodeList = (NodeList) getCachedXPathExpression(xpath).evaluate(node, XPathConstants.NODESET);
+    } catch (XPathExpressionException e) {
+      throw new IllegalArgumentException(e);
+    }
+    return CommonUtil.createList(nodeList.getLength(), nodeList::item);
```
superseacat:
```java
    return IntStream
        .range(0, nodeList.getLength())
        .mapToObj(nodeList::item)
        .collect(Collectors.toList());
```
?

src/main/java/fi/aalto/cs/apluscourses/utils/async/TaskManager.java
```diff
@@ -0,0 +1,57 @@
...
+ * An abstract interface for asynchronous execution following the fork-join model of task
+ * parallelism.
+ *
+ * <p>Note that it is not, however, required that the implementing classes actually use any
+ * concurrent computation.</p>
+ *
+ * @param <T> Type of the task objects (or futures) that are used as identifiers for asynchronous
+ *            tasks
+ */
+public interface TaskManager<T> {
```
superseacat: I have certain doubts, that we actually need our own Fork/Join ~~with blackjack and ...~~ for this task. If the idea is to parallelize some processes like installation and modules fetching, then maybe the [parallelStream()](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#parallelStream--) will do? 
What comes to the asynchrony, then a direct usage of `CompletableFuture` looks fit here.

I don't have much to say about the implementation - it looks cool, but I'm not sure if the tasks we work with actually require this level of complexity and the benefits gained from introducing such conceptual and maintenance load overweight the losses.

Curious reading [comparing Fork/Join Framework vs. Parallel Streams vs. ExecutorService](https://blog.overops.com/forkjoin-framework-vs-parallel-streams-vs-executorservice-the-ultimate-benchmark/) found on the internet.

src/main/java/fi/aalto/cs/apluscourses/utils/Resources.java
OlliKiljunen: Previously, **o1.json** was read using file system access. That caused some compatibility issues when developing on Windows, so, as a quick fix, I changed it to be read from resources. I reused this function which was already implemented. To be able to use it, I changed the method public.

The whole thing is of course a temporary solution and "for developing purposes only" because eventually the course config files are downloaded from the web. This method can then be changed back to private, if that feels appropriate.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ImportModuleAction.java
OlliKiljunen: I wasn't aware of that. A very good idea! I will change this.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/RequiredPluginsCheckerAction.java
OlliKiljunen: https://github.com/Aalto-LeTech/intellij-plugin/pull/57#discussion_r379923962

src/main/java/fi/aalto/cs/apluscourses/presentation/base/BaseListViewModel.java
OlliKiljunen: Yes. And now when I think about it, the model classes could be declared serializable (not now but at some point when the software and the whole world are otherwise ready).

So I remove the transient keywords and let Sonar keep reminding us of this (non-)issue which can then be "fixed" by declaring model classes serializable at some point.

src/main/java/fi/aalto/cs/apluscourses/model/ModuleLoadException.java
OlliKiljunen: https://github.com/Aalto-LeTech/intellij-plugin/pull/57#discussion_r379924549

src/main/java/fi/aalto/cs/apluscourses/utils/DomUtil.java
OlliKiljunen: Okay, to be honest, I am not an expert on these kind of security things. The XXE vulnerability (and how to fix it) was pointed out by Sonar. Do you think this solution is ok? (Both my Ubuntu machine and my Windows 10 machine at home fails to set those security measures thus falling to the catch block and logging the warning.)

As far as I understood it, this is not so crucial thing in our case, because users (students) should load XML files only from sources they trust (teachers), and as the result of the parse is not sent out from the users computer, there is no chance of information leaking. Only bad thing that could cause if a user parses a malicious XML (in this case *.iml) file they got from somewhere, would be their IDE to crash. Do you agree?

src/main/java/fi/aalto/cs/apluscourses/utils/CommonUtil.java
OlliKiljunen: Are you sure? Checkstyle google checks reports parse error when using `{@link <T>}`. 🤔

src/test/java/fi/aalto/cs/apluscourses/presentation/base/BaseListViewModelTest.java
OlliKiljunen: That map is not used in assertions. It is only used mapping model objects to view model objects in `BaseListViewModel` constructor.

src/main/java/fi/aalto/cs/apluscourses/utils/StateMonitor.java
OlliKiljunen: Hm... `StateMonitor` class is basically a representation of an integer value with certain properties (quite similar to AtomicInteger). So in that sense, `get()` and `set()` method names are, IMO, reasonable. Perhaps the problem is the name of the class. `StateMonitor` -> `MonitoredState` or `ThreadSafeProgressingState` or something.

src/main/java/fi/aalto/cs/apluscourses/utils/StateMonitor.java
OlliKiljunen: Related: https://github.com/Aalto-LeTech/intellij-plugin/pull/57#discussion_r379926959

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
OlliKiljunen: Sonar says `RuntimeException`s (like `JSONException`) should not be declared in `throws` statement. We certainly can mutiny against Sonar here, if you like, but I cannot say which way is better.

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
OlliKiljunen: It's a good idea to unify this. I started with `ex` but then, at some point, it struck my mind that `ex` must be C# convention and in Java, `e` is used instead, and so I changed to use `e` but was too lazy to revise those I wrote earlier.

You are probably right in that either can be used and your argument favoring `ex` is good. But I suggest we leave this issue out of this PR as there is already so much stuff in it.

src/main/java/fi/aalto/cs/apluscourses/utils/StateMonitor.java
OlliKiljunen: Good catch, this is an obsolete relic and shouldn't exist. I'll remove it.

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
OlliKiljunen: Probably, yes. I have to admit that when writing this I just didn't remember how Java enums work, and I was too lazy to start studying them. I'll check if enum can be used here, and change it, if it works well in this case.

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
OlliKiljunen: https://github.com/Aalto-LeTech/intellij-plugin/pull/57#discussion_r379406878

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
OlliKiljunen: You mean that the base class does that check in `getDependencies()` and then calls another method (`protected List<String> getDependenciesInternal()`) implemented by subclasses? Yeah, that's possible. I can change that if you feel it improves the code.

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
OlliKiljunen: Yes. Same as https://github.com/Aalto-LeTech/intellij-plugin/pull/57#discussion_r379406878. I'll change it to throw an `UnsupportedOperationException`.

OlliKiljunen: Or actually NO! I mean, it won't be implemented in this class. It's something that subclass(es) implement. It should be abstract. Whole this class should be abstract! But making it abstract would cause quite a bit changes to unit tests of this class (instead of creating concrete `Module` objects we should either create subclass objects or mocks).

So yes I see this is a bad solution to have class that is practically abstract but is not "officially" abstract. If you want me to change it in this PR, I can do that. It's an easy and straightforward thing to do, just more changed lines of code to this PR.

src/main/java/fi/aalto/cs/apluscourses/model/ModuleInstallerImpl.java
OlliKiljunen: Subclass (`IntelliJModule`) implements the method.

src/main/java/fi/aalto/cs/apluscourses/model/ModuleInstallerImpl.java
OlliKiljunen: Yep, you are right, we should show the user what went wrong and allow them to clean the modules state somehow and try again. But I'd suggest we implement that later (in particular after/during implementing #64 when we have more complete understanding on mapping between IDEA modules and `Module` objects and have designed policies like "what happens if user removes or unloads imported module", "what if user creates manually a module called GoodStuff and then try to import that from the grader").

src/main/java/fi/aalto/cs/apluscourses/presentation/module/ModuleListElementViewModel.java
OlliKiljunen: I used term "fetch" instead of "download" because I thought it is more general term which can refer either downloading or file copying (like the current implementation) or whatever. The `Module` (base) class doesn't care about *how* its content is fetched, so it uses this general term.

However, at the level of UI, we know that we are doing downloading when we fetch (in the final product), so we show the user more specific description on what is done.

src/main/java/fi/aalto/cs/apluscourses/utils/DomUtil.java
OlliKiljunen: Damn, you can't teach an old dog new tricks! 😆

src/main/java/fi/aalto/cs/apluscourses/utils/async/TaskManager.java
OlliKiljunen: My point here was to postpone the design decision which threading framework and thread pool we use by abstracting the task parallelism behind a single interface, which can easily be implemented with `CompleteableFuture`s, `Thread`s or, I'd guess, any task based parallel library, and allow execution to be dynamically changed to single-threaded for debugging and testing purposes.

Perhaps, when we are sure which threading framework we're gonna use and how, we can remove this extra layer of abstraction?

src/main/java/fi/aalto/cs/apluscourses/model/ModuleInstallerImpl.java
OlliKiljunen: Not sure what you mean... NPE here means that `moduleSource.getModule()` returned null... BUT WAIT! We were not allowed to return nulls! Oh... my bad. I should fix that... Not a difficult one, luckily. :P

src/main/java/fi/aalto/cs/apluscourses/model/ModuleSource.java
OlliKiljunen: I'll fix that to not return null. A very good point!

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
OlliKiljunen: My understanding: in one IntelliJ IDEA window there can be only one project open at any time. So the entirety of code you have open in one window = a PROJECT.

The project have 0..n modules. Symbol of a module is that folder with a cyan square (well, that marks the dir that is `$MODULE_DIR$` for that module). To see the list of all the modules in the project either
- see **<PROJECT_ROOT>/.idea/modules.xml**,  OR
- menu File->Project Structure...->Project Settings->Modules

One thing to note is that modules are not nested! **There is no tree structure of modules.** One module's `$MODULE_DIR$` can be inside another module's `$MODULE_DIR$` but that doesn't imply any kind of relationship between the modules. That's just how the files of there modules are located in the file system. The dependencies between the modules inside a project is a separate structure (basically, a directed graph) which is defined in the *.iml files of the modules and which is totally independent of the directory structure.

The situation in your picture (the right window) looks a bit weird. Is it possible that you have loaded **intellij-plugin** and **plugin-importer-exporter** as *modules* to **tmc-intellij** *project*.

image


I suggest that we take a look at this together on Monday.

Finally, I am not sure how IntelliJ works when the same *project* is opened in two windows. AFAIK, normally IntelliJ would not allow that but perhaps there can be cases where that happens (like having running both Ultimate and Community versions).

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
```diff
@@ -0,0 +1,118 @@
...
+  public void fetch() throws IOException {
+    File tempZipFile = createTempFile();
+    fetchZipTo(tempZipFile);
+    extractZip(tempZipFile);
+  }
+
+  @Override
+  public void load() throws ModuleLoadException {
+    try {
+      WriteAction.runAndWait(new Loader(getProject(), getImlFile())::load);
```
nikke234: good stuff! do you happen to have a link to some documentation on `runAndWait` (my googling so far has failed)?

src/main/java/fi/aalto/cs/apluscourses/intellij/toolwindows/BaseToolWindowFactory.java
```diff
@@ -1,21 +1,21 @@
...
 import com.intellij.openapi.wm.ToolWindowFactory;
 import com.intellij.ui.content.Content;
 import com.intellij.ui.content.ContentFactory;
-import fi.aalto.cs.intellij.ui.ModuleList;
+import javax.swing.JComponent;
 import org.jetbrains.annotations.NotNull;
 
-public class ModulesToolWindow implements ToolWindowFactory {
-
+public abstract class BaseToolWindowFactory implements ToolWindowFactory {
```
nikke234: removes some previous code duplication of mine :smile: nice :+1:

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
nikke234: If we want to have it in the `throws` clause, then I think we should catch `JSONException` here and throw our own exception that doesn't extend `RuntimeException`.

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -0,0 +1,10 @@
+package fi.aalto.cs.apluscourses.presentation;
+
+import fi.aalto.cs.apluscourses.utils.ObservableProperty;
+import org.jetbrains.annotations.NotNull;
+
+public class MainViewModel {
+
+  @NotNull
+  public final ObservableProperty<CourseViewModel> courseViewModel = new ObservableProperty<>(null);
```
nikke234: Nit (and not urgent to decide on for this pr): we could decide on some consistent way to choose between `public final` and getters (for an example, our exception classes have private finals with getters)

src/test/java/fi/aalto/cs/apluscourses/model/ModuleInstallerTest.java
```diff
@@ -0,0 +1,294 @@
+package fi.aalto.cs.apluscourses.model;
```
nikke234: `ModuleInstallerTest` seems to get stuck for some reason (which also fails travis builds)

nikke234: This test gets stuck for me locally (which I suspect is also why travis fails)

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ActionUtil.java
OlliKiljunen: Yes, the name is bad. My idea was something like "this method returns 'an object that launches anAction when an event occurs', in other words, it returns an 'on-event launcher'".

How about `getOnEventLauncher()`, `createAutomaticLauncher()` or something alike?

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
OlliKiljunen: I'll leave this for you, @nikke234, to decide as you are the original author of this method, right? In any case, I'd leave it out of this PR.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
OlliKiljunen: https://www.jetbrains.org/intellij/sdk/docs/basics/architectural_overview/general_threading_rules.html#readwrite-lock

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
OlliKiljunen: What makes (perhaps) the difference here is the nature of `ObservableProperty<>`. Consider:
```java
mainViewModel.getCourseViewModel().set(new CourseViewModel(...));
```
Here we want to *set* the course view model so calling *getter* for that is a bit confusing. That's why I thought
```java
mainViewModel.courseViewModel.set(new CourseViewModel(...));
```
is better although it is not that good OOP.

src/test/java/fi/aalto/cs/apluscourses/model/ModuleInstallerTest.java
OlliKiljunen: Hopefully this is now fixed.

src/main/java/fi/aalto/cs/apluscourses/utils/DomUtil.java
superseacat: I wasn't aware of this XXE thingy before, but I believe this definitely needs to stay!

src/main/java/fi/aalto/cs/apluscourses/utils/CommonUtil.java
superseacat: there were warning pointing to this when I tried to generate JavaDoc. I've actually created a new issue to generate and publish JavaDocs #73

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
superseacat: this is the case where C# approach is fully applicable :)

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
OlliKiljunen: Tried to change to enum, but conversions between int and enum values got overly complicated, so i'd suggest we keep it old-school.

OlliKiljunen:
Thank you for your valuable comments; once again they proved to be extremely helpful. I guess I have somehow addressed all the comments now. Please feel free to add more comments, if you like.

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
nikke234: Good point, the `get...set` is indeed somewhat confusing :smile:

src/main/java/fi/aalto/cs/apluscourses/model/ModelFactory.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.List;
+import java.util.Map;
+import org.jetbrains.annotations.NotNull;
+
+public interface ModelFactory {
+  Course createCourse(@NotNull String name,
+                      @NotNull List<Module> modules,
+                      @NotNull Map<String, String> requiredPlugins);
+
+  Module createModule(@NotNull String name, @NotNull URL url);
```
nikke234: Nit (can be fixed later in some other PR): return values could have `NotNull` as well

src/main/java/fi/aalto/cs/apluscourses/model/NoSuchModuleException.java
```diff
@@ -0,0 +1,10 @@
+package fi.aalto.cs.apluscourses.model;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class NoSuchModuleException extends Exception {
+  public NoSuchModuleException(@NotNull String moduleName, @Nullable Throwable cause) {
+    super("Module '" + moduleName + " was not found'.", cause);
```
nikke234: Closing single quote is in the wrong place :trollface:

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
superseacat: `This method should not be called unless the module is in FETCHED state or further.` should this method itself know when it can be called and notify back ("Sorry, bad state, won't even try to get(); ... ")

src/main/java/fi/aalto/cs/apluscourses/utils/DirAwareZipFile.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    for (String fileName : getFileHeaders()
+        .stream()
+        .map(FileHeader::getFileName)
+        .filter(Objects::nonNull)
+        .filter(fileName -> fileName.startsWith(prefix))
+        .toArray(String[]::new)) {
+      extractFile(fileName, destinationPath);
+    }
```
superseacat: `Exception` handling, eeehh? :crying_cat_face:

could this be maybe like this... streaming the part of `for` loops just feels so unconventional:
```java
    String[] fileNamesArray = getFileHeaders()
        .stream()
        .map(FileHeader::getFileName)
        .filter(Objects::nonNull)
        .filter(fileName -> fileName.startsWith(prefix))
        .toArray(String[]::new);

    for (String filename : fileNamesArray) {
      extractFile(filename, destinationPath);
    }
```

src/main/java/fi/aalto/cs/apluscourses/presentation/base/BaseListViewModel.java
```diff
@@ -0,0 +1,112 @@
...
+    return elements.size();
+  }
+
+  @Override
+  @Nullable
+  public E getElementAt(int i) {
+    try {
+      return elements.get(i);
+    } catch (IndexOutOfBoundsException e) {
+      return null;
```
superseacat: :stuck_out_tongue_winking_eye:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ActionUtil.java
superseacat: I'd go at least with the 1st option

src/main/java/fi/aalto/cs/apluscourses/presentation/base/BaseListViewModel.java
OlliKiljunen: The interface doesn't specify how it should work in those situations:
https://docs.oracle.com/javase/8/docs/api/javax/swing/ListModel.html#getElementAt-int-
I can't no longer remember what was my exact reasoning on this but if the client code in `JList` is something like
```java
int size = listModel.getSize();
for (int i = 0; i < size; i++) {
  E element = listModel.getElementAt(i);
  doSomething(element);
}
```
then there is certainly a possibility that the method is called with an out-of-bounds index (because of concurrency).

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ActionUtil.java
OlliKiljunen: I made a compromise: `createOnEventLauncer()`

src/main/java/fi/aalto/cs/apluscourses/utils/DirAwareZipFile.java
OlliKiljunen: I changed that as you suggested. Also wrote something like a unit test for the class.

PR #75, opened 2020-02-17
configure testing environment

superseacat:
# Description
this addition will trigger an A+ env to get established within the Travis builds + add a tiny change to the plugin description

# Testing

- [ ] All unit tests pass
- [x] I have tested manually

- [x] Yes
- [ ] Not relevant

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,5 @@ services:
   - docker
 
 script:
-  - ./omytest.sh
+  - ./start_local_a+_env.sh
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,6 @@ services:
   - docker
 
 script:
+  - chmod +x start_local_a+_env.sh
   - ./start_local_a+_env.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,25 @@ dist: trusty
 jdk:
   - oraclejdk11
 
-services:
-  - docker
+addons:
+  sonarcloud:
+    organization: "aalto-letech-intellij-plugin"
+    token: $SONARCLOUD_TOKEN
 
 script:
-  - chmod +x start_local_a+_env.sh
-  - ./start_local_a+_env.sh
+  - ./gradlew check
+  - chmod +x start_local_a+_env.sh && ./start_local_a+_env.sh
+  - sonar-scanner
 
+deploy:
+  - provider: script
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -10,6 +10,7 @@ addons:
 
 script:
   - ./gradlew check
+  - chmod +x start_local_a+_env.sh && ./start_local_a+_env.sh
```
OlliKiljunen: Just a note: perhaps all these steps should be joined with &&s so that if one step fails, subsequent steps are not run at all. But as there haven't be such functionality before, there is no need to change it in this PR.

.travis.yml
```diff
@@ -23,4 +24,4 @@ deploy:
       branch: test
 
 notifications:
-  slack: $SLACK
+  slack: $SLACK
```
OlliKiljunen: Nit: there is IntelliJ setting that auto-ensures LF in EOF.

start_local_a+_env.sh
```diff
@@ -0,0 +1,17 @@
...
+head -97 docker-up.sh > docker-up-custom.sh &&
+chmod +x docker-up-custom.sh &&
+ls -al &&
+docker ps &&
+netstat -tulpn &&
+./docker-up-custom.sh &&
+sleep 2m &&
+docker ps &&
+netstat -tulpn &&
+curl --user root:root http://localhost:8000/api/v2/courses/
```
OlliKiljunen: Nit: some comments here would be nice. Especially, why do we create a custom docker script with that `head` command (a bit hacky, not? ;) ). And `ls` is just for logging purposes, right? And what is the purpose of that 2m sleep. (I guess there was some discussion on that in Slack but can't remember exactly what.)

src/main/resources/META-INF/plugin.xml
```diff
@@ -12,7 +12,7 @@
     ]]></description>
 
     <change-notes><![CDATA[
-      Adding the initial skeleton of the plugin.<br>
+      Added some basic functionality for plugin configuration and basic course materials loading.<br>
```
OlliKiljunen: So, it is finally time for this. :O XD

start_local_a+_env.sh
superseacat: comments added

.travis.yml
superseacat: why not? :)

Issue #65, opened 2020-02-05
investigate how Scala REPL thing can be done


superseacat:
for installation of scala plugin:

- follow: https://www.jetbrains.org/intellij/sdk/docs/basics/plugin_structure/plugin_dependencies.html#gradle
- check this (optional): https://github.com/JetBrains/gradle-intellij-plugin#configuration

superseacat:
take into account options raised in the requirements (i, ii, iii)

superseacat:
here it starts: https://github.com/JetBrains/intellij-scala/tree/55ba27a93564021a7b60529424b726775965ee3e/scala/scala-impl/src/org/jetbrains/plugins/scala/console

superseacat:
right, so the proposed mechanism (to be verified with stakeholders):
1. user starts a REPL "the usual Scala plugin way"
2. a popup window shows up
 "the file that the REPL was called on belongs to the module X. If you wish to preserve this behavior - check [x] `this` checkbox. 

If you want to use some other module instead, then choose here [dropdown list of available modules]."


I'll demo on Wednesday (26.2)

superseacat:
some more details: https://github.com/Aalto-LeTech/intellij-plugin/pull/80

PR #80, opened 2020-02-21
NOT TO BE MERGED: Repl research

superseacat:
# Description

so this is just (probably) one way how to approach this. I'll leave some comments inside the code.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id 'java'
+    id 'scala'
     id 'checkstyle'
     id 'org.jetbrains.intellij' version '0.4.15'
     id "org.sonarqube" version "2.8"
@@ -19,6 +19,7 @@ repositories {
 dependencies {
     compile group: 'commons-io', name: 'commons-io', version: '2.6'
     compile group: 'net.lingala.zip4j', name: 'zip4j', version: '2.3.0'
+    compile group: 'org.scala-lang', name: 'scala-library', version: '2.13.1'
     testCompile group: 'junit', name: 'junit', version: '4.12'
     testCompile group: 'org.hamcrest', name: 'hamcrest', version: '2.2'
     implementation group: 'org.json', name: 'json', version: '20190722'
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,7 @@ dependencies {
 
 intellij {
     version '2019.3.2'
-    plugins = ['java']
+    plugins = ['java', 'org.intellij.scala:2019.3.26']
 }
 
 publishPlugin {
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -17,6 +17,7 @@
     </change-notes>
 
     <depends>com.intellij.modules.java</depends>
+    <depends>org.intellij.scala</depends>
 
     <extensions defaultExtensionNs="com.intellij">
 
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -53,7 +53,6 @@
                 text="Check Required Plugins"/>
         <action id="Scala.RunConsole" class="fi.aalto.cs.apluscourses.intellij.actions.REPLAction"
           overrides="true" text="Scala REPL..." description="Run Scala REPL">
-            <add-to-group group-id="RunContextPopupGroup" anchor="last"/>
             <add-to-group group-id="ToolsMenu" anchor="last"/>
             <keyboard-shortcut first-keystroke="control shift D" keymap="$default"/>
         </action>
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -51,5 +51,11 @@
         <action id="fi.aalto.cs.apluscourses.intellij.actions.RequiredPluginsCheckerAction"
                 class="fi.aalto.cs.apluscourses.intellij.actions.RequiredPluginsCheckerAction"
                 text="Check Required Plugins"/>
+        <action id="Scala.RunConsole" class="fi.aalto.cs.apluscourses.intellij.actions.REPLAction"
+          overrides="true" text="Scala REPL..." description="Run Scala REPL">
+            <add-to-group group-id="RunContextPopupGroup" anchor="last"/>
+            <add-to-group group-id="ToolsMenu" anchor="last"/>
+            <keyboard-shortcut first-keystroke="control shift D" keymap="$default"/>
+        </action>
     </actions>
 </idea-plugin>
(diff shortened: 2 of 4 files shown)


build.gradle
```diff
@@ -27,7 +28,7 @@ dependencies {
 
 intellij {
     version '2019.3.2'
-    plugins = ['java']
+    plugins = ['java', 'org.intellij.scala:2019.3.26']
```
superseacat: adding a particular version of the scala plugin (it seems to be updated once the IDE starts with `runIde`), additionally you need to complete this step: https://github.com/Aalto-LeTech/intellij-plugin/issues/65#issue-560340331

src/main/resources/META-INF/plugin.xml
```diff
@@ -17,6 +17,7 @@
     </change-notes>
 
     <depends>com.intellij.modules.java</depends>
+    <depends>org.intellij.scala</depends>
```
superseacat: exposing strict dependency on scala plugin (by ID), now our plugin won't even start wo Scala plugin installed/enabled

src/main/resources/META-INF/plugin.xml
```diff
@@ -50,5 +51,10 @@
         <action id="fi.aalto.cs.apluscourses.intellij.actions.RequiredPluginsCheckerAction"
                 class="fi.aalto.cs.apluscourses.intellij.actions.RequiredPluginsCheckerAction"
                 text="Check Required Plugins"/>
+        <action id="Scala.RunConsole" class="fi.aalto.cs.apluscourses.intellij.actions.REPLAction"
+          overrides="true" text="Scala REPL..." description="Run Scala REPL">
+            <add-to-group group-id="ToolsMenu" anchor="last"/>
+            <keyboard-shortcut first-keystroke="control shift D" keymap="$default"/>
+        </action>
```
superseacat: almost identical to the [actual declaration of the action in Scala Plugin](https://github.com/JetBrains/intellij-scala/blob/69a0f0d5d4dd2c9a637e64761ac7d8bba49d2040/scala/scala-impl/resources/META-INF/scala-plugin-common.xml#L1477-L1482), the most important change is `overrides="true"` - this [seems](https://intellij-support.jetbrains.com/hc/en-us/community/posts/207623085-Intellij-plugin-development-how-to-prevent-an-action-from-occuring-like-closing-a-tab) to be the way to override actions (assume this works for IDE's own actions too as it works here)

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
```diff
@@ -0,0 +1,25 @@
+package fi.aalto.cs.apluscourses.intellij.actions
+
+import com.intellij.openapi.actionSystem.{AnActionEvent, CommonDataKeys}
+import com.intellij.openapi.module.{ModuleManager, ModuleUtilCore}
+import org.jetbrains.plugins.scala.console.actions.RunConsoleAction
+
+class REPLAction extends RunConsoleAction {
```
superseacat: yepp, I could not find a simple way to extend [relevant Scala class](https://github.com/JetBrains/intellij-scala/blob/ca22e710b389b0eb58cdb5b69aae62bf3273d442/scala/scala-impl/src/org/jetbrains/plugins/scala/console/actions/RunConsoleAction.scala) in Java, so, well, welcome our first it's a Scala class :rofl: 
@nikke234 must be happy!

build.gradle
```diff
@@ -1,5 +1,5 @@
 plugins {
-    id 'java'
```
superseacat: this is changed in order the plugin would also be able to compile scala sources (scala plugin can do both: Java and Scala, whereas 'java' seems to work only for Java)

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
```diff
@@ -0,0 +1,25 @@
...
+
+  override def actionPerformed(e: AnActionEvent): Unit = {
```
superseacat: what happens in general:
we point our plugin to the new action class responsible for the processing. when the actual event is fired, it gets intercepted here, we can adjust the load of the event as we need and then pass it further to the actual implementation (of Scala plugin). with this approach, at least based on simple testing, the impact is minimal as well as dependency, we just reuse...

build.gradle
OlliKiljunen: How does this work, if a user has a different version of scala plugin already installed?

superseacat: **hm... now it does not upgrade...**
Screenshot from 2020-02-21 11-20-55

superseacat: Screenshot from 2020-02-21 11-27-54
**suggest an update, if the version is below the latest available for this build of IDEA**

src/main/resources/META-INF/plugin.xml
OlliKiljunen: It would be ideal if we could kinda "decorate"

build.gradle
OlliKiljunen: Do you know does it suggest downgrading Scala plugin if a user has a newer one? Or is it assumed that new versions are backward compatible?

src/main/resources/META-INF/plugin.xml
superseacat: and what if two plugins override the same action... :question:

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
nikke234: While I would love to have more Scala in our plugin, I'll look into doing this in java, so that our plugin is written in a single language :smile:

src/main/resources/META-INF/plugin.xml
nikke234: From discussion you [linked](https://intellij-support.jetbrains.com/hc/en-us/community/posts/207623085-Intellij-plugin-development-how-to-prevent-an-action-from-occuring-like-closing-a-tab):

> The idea is that ActionManager allows to get AnAction instance by its `id`. These `id` are registered in `.xml` files in `<actions>` block. If two plugins (or a plugin and platform) are trying to assign same `id` - it causes an error. Unless one of them passes `overrides` parameter.

So I assume/hope that if two plugins override the same action, it causes an error, which would be better than silently ignoring one

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
nikke234: This seems like a great approach! I assume we would insert some module selection stuff here before calling the Scala plugin's `actionPerformed`?

This way we avoid having to implement anything directly related to the REPL.

src/main/resources/META-INF/plugin.xml
nikke234: This is the only part I don't like about this approach... But having a working REPL is way higher in the priority list than making the plugin usable for non-scala-courses so I'm completely fine with using this approach for now

build.gradle
superseacat: I'll check

Issue #39, opened 2020-01-15
start creating a client to fetch the learning module from the API


superseacat:
take into account the possibility of some files/modules/course materials to get updated

superseacat:
https://docs.docker.com/install/linux/docker-ce/ubuntu/#install-using-the-repository

Issue #32, opened 2020-01-15
fetch the course configuration file from the API


superseacat:
https://grader.cs.hut.fi/static/O1_2020/projects/o1_course_config.json

this comes from (don't forget to update if changed): https://version.aalto.fi/gitlab/jsorva/o1matsku/blob/s2020/projects/o1_course_config.json

superseacat:
smth like a public HTTP server (A+)

superseacat:
props files (with URLs etc.)

superseacat:
once this is done the local conf file should be removed and the remote used only

PR #81, opened 2020-02-21
Fetch course configuration file and modules from server

nikke234:
# Description

Adds a client for fetching the course configuration file and modules from a server and integrates it with the current course configuration file and module importing features.

TODO: More unit tests.

# Testing

- [x] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant


src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
```diff
@@ -0,0 +1,171 @@
...
+   * @throws IOException                 If an error (e.g. network error) occurs while downloading
+   *                                     the file.
+   * @throws UnexpectedResponseException If the response isn't as expected (i.e. a status code other
+   *                                     than 2xx or different Content-Type header).
+   */
+  public static void fetchZip(@NotNull URL url, @NotNull File file)
+      throws IOException, UnexpectedResponseException {
+    fetch(url, "application/zip", entity -> {
+      FileUtils.copyInputStreamToFile(entity.getContent(), file);
+      return null; // The return value gets ignored here
```
OlliKiljunen: Nit: This looks a bit like abuse of mapper interface, as a programmer would usually expect mapping to be free of side-effects. Here, the mapper is more like a consumer, right? But fair enough. :)

src/test/java/fi/aalto/cs/apluscourses/utils/DirAwareZipFileTest.java
```diff
@@ -1,79 +0,0 @@
...
-import static org.mockito.Mockito.verifyNoMoreInteractions;
-
-import java.io.FileNotFoundException;
-import java.util.ArrayList;
-import java.util.List;
-import net.lingala.zip4j.exception.ZipException;
-import net.lingala.zip4j.model.FileHeader;
-import org.junit.Test;
-
-public class DirAwareZipFileTest {
```
OlliKiljunen: Info for other reviewers: I asked Nikolas to delete this test (instead of trying to "fix" it) written by me because it was bad, and shouldn't have been there in the first place.

src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Makes a GET request to the given URL and returns the value of the Last-Modified header of the
+   * response. Note that some servers don't add the Last-Modified header to the response.
+   * @return A string containing the value of the Last-Modified header.
+   * @throws IOException                 If an error occurs in the execution of the request.
+   * @throws UnexpectedResponseException If the status code of the response isn't 2xx, or if the
+   *                                     response doesn't contain the Last-Modified header
+   */
+  @NotNull
+  String getLastModified(@NotNull URL url) throws IOException, UnexpectedResponseException {
```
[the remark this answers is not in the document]
nikke234: Good catch! It should definitely be static :+1:

src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
nikke234: Yes, you are 100% correct, this was laziness on my part :smile: I'll probably add a consumer functional interface

nikke234:
> Did you try to make some of the logic in the client unit-testable? I guess there is something related to those http status checks and entity to object mappings that coukd be tested with mocked input streams? But not necessary, really.

Yes, I'm planning to add some unit tests for the client. At least those status checks etc. are easy to test, and the rest should be doable as well.

OlliKiljunen:
I merged this now in order to be able to test it in the tomorrow morning. I'd say that if this passes the manual testing, we can consider issues #32 and #39 done. If you want to add those unit tests, feel free to create a new PR for them.

Issue #95, opened 2020-03-07
configure checkstyle for Scala


superseacat:
- in Travis
- also upload the checkstyle file to our repo

PR #96, opened 2020-03-07
REPL implementation

superseacat:
# Description
so this should be enough to cover the first (no-UI) part or the REPL milestone. It's basically my first code in Scala, so I'd ask for some advice via comments ;-)

And some older tests might fail, but let's see

Created a task https://github.com/Aalto-LeTech/intellij-plugin/issues/95 to introduce some kind of checkstyle for Scala, any suggestions @oseppala @nikke234 

# Testing

- [ ] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant


superseacat:
a reference might be handy: https://github.com/Aalto-LeTech/intellij-plugin/pull/80

src/main/resources/META-INF/plugin.xml
```diff
@@ -17,6 +17,7 @@
     </change-notes>
 
     <depends>com.intellij.modules.java</depends>
+    <depends optional="true" config-file="plugin.xml">org.intellij.scala</depends>
```
superseacat: as agreed

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
```diff
@@ -0,0 +1,64 @@
...
+
+    val runManagerEx = RunManagerEx.getInstanceEx(project)
+    val configurationType = getMyConfigurationType
+    val settings = runManagerEx.getConfigurationSettingsList(configurationType).asScala
+
+    // get target module
+    val module = ModuleUtilCore.findModuleForFile(targetFileOrFolder, project)
+
+    //choose the configuration to run based on the condition if this a new configuration of not
+    def chooseConfigurationSettings: RunnerAndConfigurationSettings = {
```
superseacat: created a nested function just to ask someone (@nikke234) what's the point of this language feature :wink:

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  def adjustRunConfigurationSettings(@NotNull module: Module, @NotNull configuration: ScalaConsoleRunConfiguration): Unit = {
+    // adjust the configuration with: name, workDir and module
+    val moduleDirPath = ModuleUtilCore.getModuleDirPath(module)
+    configuration.setWorkingDirectory(moduleDirPath)
+    configuration.setModule(module)
+    configuration.setName("Scala REPL for module: " + module.getName)
+  }
+
+  def customDoRunAction(e: AnActionEvent): Unit = {
```
superseacat: this whole method is a mess, and any proposals on how to improve it are welcome :exclamation:

superseacat: one of the options I see could be cutting it into multiple smaller methods, thus, it will probably become harder to read, as there's gonna be plenty of joggling with same data

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
```diff
@@ -0,0 +1,64 @@
...
+    val targetFileOrFolder = CommonDataKeys.VIRTUAL_FILE.getData(dataContext)
+
+    if (project == null || checkFileOrFolderIsNull(targetFileOrFolder)) return
+
+    val runManagerEx = RunManagerEx.getInstanceEx(project)
+    val configurationType = getMyConfigurationType
+    val settings = runManagerEx.getConfigurationSettingsList(configurationType).asScala
+
+    // get target module
+    val module = ModuleUtilCore.findModuleForFile(targetFileOrFolder, project)
```
superseacat: I kinda get frustrated with all the `val`s, but is there a better alternative?

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
```diff
@@ -0,0 +1,64 @@
...
+
+import com.intellij.execution.{RunManager, RunManagerEx, RunnerAndConfigurationSettings}
+import com.intellij.openapi.actionSystem.{AnActionEvent, CommonDataKeys}
+import com.intellij.openapi.module.{Module, ModuleUtilCore}
+import com.intellij.openapi.vfs.VirtualFile
+import org.jetbrains.annotations.{NotNull, Nullable}
+import org.jetbrains.plugins.scala.console.actions.RunConsoleAction
+import org.jetbrains.plugins.scala.console.configuration.ScalaConsoleRunConfiguration
+
+import scala.collection.JavaConverters._
```
superseacat: this is for scala **2.12**, as I had to downgrade from **2.13**, because some terrible things started to happen during the compilation phase if the version was not the same as the one for [JB's Scala Plugin](https://plugins.jetbrains.com/plugin/1347-scala)

src/test/scala/REPLActionTest.scala
```diff
@@ -0,0 +1,73 @@
...
+    //given
+    val action = new REPLAction
+    val spyAction = spy(action)
+    val actionEvent = createActionEvent(action)
+
+    //when
+    spyAction.actionPerformed(actionEvent)
+
+    //then
+    verify(spyAction, times(1)).customDoRunAction(actionEvent)
```
superseacat: actually, testing the `customDoRunAction(...)` would be more relevant. I'm still thinking of how to approach this

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
OlliKiljunen: Hmm... An interesting question. I think nested functions are useful 1) when called multiple times, or, 2) for some recursive functions to keep the public interface cleaner (pseudo-code example):
```
int sum(list) {
  int sum_internal(s, list) {
    if (empty(list))
      return s;
    else
      return sum_internal(s + first(list), rest(list));
  }
  return sum_internal(0, list);
}
```

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
```diff
@@ -0,0 +1,64 @@
...
+
+    // get target module
+    val module = ModuleUtilCore.findModuleForFile(targetFileOrFolder, project)
+
+    //choose the configuration to run based on the condition if this a new configuration of not
+    def chooseConfigurationSettings: RunnerAndConfigurationSettings = {
+      var setting: RunnerAndConfigurationSettings = null
+      if (settings.nonEmpty) {
+        setting = settings.head
+      } else if (settings.isEmpty) {
```
OlliKiljunen: This should be just `else`? Is there some third option besides `nonEmpty` and `isEmpty`?

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
OlliKiljunen: Not that bad IMO. It's quite straightforward after all. I counted three different paths of execution. I'd suggest either inlining the nested function or  moving it outside of this function.

src/test/scala/REPLActionTest.scala
OlliKiljunen: The last two lines of `customDoRunAction(...)` function are those that *do* something; all before that is setup, right? Perhaps, we could extract those two lines behind and interface which `customDoRunAction(...)` would then call like:
```
consoleFacade.showConsole(module, setting, whatever);
```
and in unit-tests, it is verified that the method is called with correct parameters.

Okay, but there would still be a lot of nasty mocking, so perhaps we leave that method out of testing; also it contains only a little logic so that's fine, I guess.

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
nikke234: Some helper functions or tail-recursive inner functions as in Olli's pseudo code are some good use cases. The public interface can be kept clean with `private`, but I think one advantage nested functions is the ability to access variables of the outer function:
```
def outer(x: Int): Int = {
  val y = x + 5
  def inner(a: Int): Int = {
    // Can read (and even modify) x and y here
  }
}
```

jsorva: Yes, I strongly agree with Nikolas's point about closures (i.e., helper functions "capturing" variables from the surrounding context). Once you get used to the idea of local, nested functions, this sort of thing is often handy and feels very natural.   

And yes, it's true that `private` is enough for keeping the public interface clean, but nested (i.e., _local_ functions) still add to that. This is in many respects analogous to how we use local variables even though we could use private variables instead. If we adopt the stance that functions are a primary means of abstraction just like variables are (i.e., functions are "first-class citizens" of the programming language, as people say), then it makes sense that functions, just like variables, can be restricted to a local scope.

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
```diff
@@ -0,0 +1,64 @@
...
+
+class REPLAction extends RunConsoleAction {
+
+  override def actionPerformed(e: AnActionEvent): Unit = {
+    customDoRunAction(e)
+  }
+
+  def checkFileOrFolderIsNull(@Nullable fileOrFolder: VirtualFile): Boolean = {
+    fileOrFolder == null
+  }
```
nikke234: Nit: I would probably just remove the curly brackets and write this as
```
def checkFileOrFolderIsNull(@Nullable fileOrFolder: VirtualFile): Boolean = fileOrFolder == null
```

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
```diff
@@ -0,0 +1,64 @@
...
+      if (settings.nonEmpty) {
+        setting = settings.head
+      } else if (settings.isEmpty) {
+        val factory = configurationType.getConfigurationFactories.apply(0)
+        setting = RunManager.getInstance(project).createConfiguration("Scala REPL for module: " + module.getName, factory)
+      }
+      setting
+    }
+
+    val setting = chooseConfigurationSettings
```
nikke234: The inner function is nice, but since it's only called once, I'd write this as 
```
val setting = {
  var setting: RunnerAndConfigurationSettings = null
  if (settings.nonEmpty) {
    setting = settings.head
  } else if (settings.isEmpty) {
    val factory = configurationType.getConfigurationFactories.apply(0)
    setting = RunManager.getInstance(project).createConfiguration("Scala REPL for module: " + module.getName, factory)
  }
  setting
}
```
Two alternative (perhaps more Scala-like) ways:
```
val setting = settings.headOption.getOrElse {
  val factory = configurationType.getConfigurationFactories.apply(0)
  RunManager.getInstance(project).createConfiguration(s"Scala REPL for module: ${module.getName}", factory)
}
```
or
```
val setting = settings.headOption match {
  case Some(head) => head
  case None       => {
    val factory = configurationType.getConfigurationFactories.apply(0)
    RunManager.getInstance(project).createConfiguration(s"Scala REPL for module: ${module.getName}", factory)
  }
}
```

nikke234: The inner function is nice, but since it's only called once, I'd write the function body directly: 
```
val setting = {
  var setting: RunnerAndConfigurationSettings = null
  if (settings.nonEmpty) {
    setting = settings.head
  } else if (settings.isEmpty) {
    val factory = configurationType.getConfigurationFactories.apply(0)
    setting = RunManager.getInstance(project).createConfiguration("Scala REPL for module: " + module.getName, factory)
  }
  setting
}
```
An alternative (perhaps more idiomatic Scala) way:
```
val setting = settings.headOption.getOrElse {
  val factory = configurationType.getConfigurationFactories.apply(0)
  RunManager.getInstance(project).createConfiguration(s"Scala REPL for module: ${module.getName}", factory)
}
```
Or yet another possibility (but I prefer the `getOrElse` one)
```
val setting = settings.headOption match {
  case Some(head) => head
  case None       => {
    val factory = configurationType.getConfigurationFactories.apply(0)
    RunManager.getInstance(project).createConfiguration(s"Scala REPL for module: ${module.getName}", factory)
  }
}
```

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
nikke234: I think it's perfectly fine as is :+1:

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
superseacat: not really :laughing:

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/REPLAction.scala
superseacat: I see now, however, for OO languages, Classes (representing objects) often allow some way to see into them, and therefore unit test, whereas the idea of nested does not seem to provide this option

src/test/scala/REPLActionTest.scala
superseacat: I've considered the interface approach option as well (regardless of the way it works with Scala), thus it just so much mocking, that I would not maybe unit test the whole thing

PR #97, opened 2020-03-09
IntelliJ module state update

### LINKED ISSUE #64: IDEA should remember that some modules were previously installed
persisted over restart?

nikke234:
# Description

This resolves #64, such that installed modules are shown as installed after restarting the IDE, and removing modules or the module files updates the modules states to uninstalled. 

# Testing

- [x] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant


src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -231,21 +190,20 @@ private static String getCourseName(@NotNull JSONObject jsonObject, @NotNull Str
...
       }
     }
     return requiredPlugins;
   }
 
-
   @NotNull
   @Override
-  public Module getModule(@NotNull String moduleName) throws NoSuchModuleException {
+  public Optional<Module> getModuleOptional(@NotNull String moduleName) {
```
OlliKiljunen: Not sure what is an advantage of using `Optional` instead of plain nullable reference, but it seems to be customary in Java so I guess that's OK. :)

nikke234: Just my instinct to avoid nulls :smile: But since we're consistently using the `@Nullable` and `@NotNull` annotations, I think a nullable reference would be completely fine here as well

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
```diff
@@ -22,11 +28,43 @@ public IntelliJModelFactory(@NotNull Project project) {
...
+        }
+      }
+    });
+    project.getMessageBus().connect().subscribe(VirtualFileManager.VFS_CHANGES,
+        new BulkFileListener() {
+          @Override
+          public void after(@NotNull List<? extends VFileEvent> events) {
+            for (VFileEvent event : events) {
+              if (event instanceof VFileDeleteEvent) {
+                String deletedFileName = event.getFile().getName();
```
superseacat: Method invocation `getName()` may produce NPE

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
```diff
@@ -22,11 +28,43 @@ public IntelliJModelFactory(@NotNull Project project) {
   public Course createCourse(@NotNull String name,
                              @NotNull List<Module> modules,
                              @NotNull Map<String, String> requiredPlugins) {
-    return new IntelliJCourse(name, modules, requiredPlugins, project);
+    IntelliJCourse course = new IntelliJCourse(name, modules, requiredPlugins, project);
+    // Add a module change listener with the created course instance to the project
+    project.getMessageBus().connect().subscribe(ProjectTopics.MODULES, new ModuleListener() {
```
superseacat: can this chained call can be extracted into a separate method (then there could be a call to it on line 43 as well)?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
```diff
@@ -22,11 +28,43 @@ public IntelliJModelFactory(@NotNull Project project) {
...
+      @Override
+      public void moduleRemoved(@NotNull Project project,
+                                @NotNull com.intellij.openapi.module.Module projectModule) {
```
superseacat: is it possible to overall unit test methods of this class?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
```diff
@@ -22,11 +28,43 @@ public IntelliJModelFactory(@NotNull Project project) {
...
   }
 
   @Override
   public Module createModule(@NotNull String name, @NotNull URL url) {
-    return new IntelliJModule(name, url, project);
+    Module module = new IntelliJModule(name, url, project);
+    // IntelliJ modules may already be present in the project or file system, so we update the state
+    // at module creation here.
+    module.updateState();
+    return module;
```
superseacat: I slightly disagree with semantics here... the name of the method does not exactly reflect the contents: _create_ vs. _update_. I'd propose a simpler renaming from `createModule()` into something like `updateOrOtherwiseCreate()`

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -231,21 +191,21 @@ private static String getCourseName(@NotNull JSONObject jsonObject, @NotNull Str
...
     }
     return requiredPlugins;
   }
 
-
-  @NotNull
+  @Nullable
   @Override
-  public Module getModule(@NotNull String moduleName) throws NoSuchModuleException {
+  public Module getModuleOpt(@NotNull String moduleName) {
```
superseacat: what does `Opt` part stand for in `getModuleOpt(...)`: optional, options?

superseacat: could you please also JavaDoc this method :smirk_cat:

nikke234: It stands for optional. Originally I named the method `getModuleOptional`, but I thought that would cause the expectation of `Optional<Module>`. Would `getModuleOptional` be a better name?

nikke234: Good point about Javadoc!

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
nikke234: Good point - it should be. I'll look into it :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
nikke234: You're right, I'll add a check

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
```diff
@@ -22,11 +28,43 @@ public IntelliJModelFactory(@NotNull Project project) {
...
+                String deletedFileName = event.getFile().getName();
+                Module module = course.getModuleOpt(deletedFileName);
+                if (module != null) {
+                  course.onModuleFilesDeletion(module);
+                }
+              }
+            }
+          }
+        }
+    );
```
superseacat: I'm re-reading the use-cases of the [`Course.getModuleOpt(...)`](https://github.com/Aalto-LeTech/intellij-plugin/blob/c42d7a5b2fdd0b5c5c70e6ac8e88905bd902be46/src/main/java/fi/aalto/cs/apluscourses/model/Course.java#L201-L209). 

In case of [`moduleRemoved(...)`](https://github.com/Aalto-LeTech/intellij-plugin/blob/c42d7a5b2fdd0b5c5c70e6ac8e88905bd902be46/src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java#L37) it seems OK, we take a projectModule's name and compare it (inside `getModuleOpt` with names of existing modules)

However, in case of [`after()`](https://github.com/Aalto-LeTech/intellij-plugin/blob/1691959d898f8d1d612cb132773aae36d0366e59/src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java#L50) I'm a bit puzzled. So overall, for the cases, where there is a bulk change of files occur, we check all the events (is there by accident a narrower _topic_ to listen?) and then pass it down to the `getModuleOpt`. So far OK, but then, inside the `getModuleOpt(...)` the **deletedFileName** (name of the file affected) being [compared to the **name of the module**](https://github.com/Aalto-LeTech/intellij-plugin/blob/1691959d898f8d1d612cb132773aae36d0366e59/src/main/java/fi/aalto/cs/apluscourses/model/Course.java#L206), so I wonder if it will ever work?

Could there be a test for such a use case or something?

nikke234: Ideally we would only listen to file deletion events here since we only care about those, but I haven't found a way to do that. However, I'll double-check. 

Comparing the deleted file name to the name of the module works, because in our case the directory of the module always has the same name as the module. For an example, going to the project directory and deleting the folder `GoodStuff` causes a file deletion event with the filename `"GoodStuff"` (same as the name of the module). 

However, deleting files from the `GoodStuff` directory won't work. For an example, deleting `GoodStuff.iml` doesn't affect the modules in anyway, because no module has the name `"GoodStuff.iml"`. This could cause problems, but I'm not sure if we need to deal with cases where students delete individual files from modules.

Issue #66, opened 2020-02-05
implement REPL thingy (ii)


superseacat:
scala plugin deps = ooptional

* The plugin runs the REPL in whichever module is currently open for editing or otherwise selected within IntelliJ.

superseacat:
take this into accout: https://github.com/Aalto-LeTech/intellij-plugin/pull/80

superseacat:
https://plugins.jetbrains.com/plugin/13634-a-courses/update/80794

PR #98, opened 2020-03-13
implement REPL thingy (ii) #66

superseacat:
and fix for stackoverflow at 'verifyPlugin', update some dependancies, add scala-java-docs

# Description
just minor fixes

# Testing

- [ ] All unit tests pass
- [ ] I have tested manually

- [ ] Yes
- [ ] Not relevant

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -17,7 +17,8 @@
     </change-notes>
 
     <depends>com.intellij.modules.java</depends>
-    <depends optional="true" config-file="plugin.xml">org.intellij.scala</depends>
+    <depends optional="true" config-file="">org.intellij.scala</depends>
+<!--    <depends optional="true" config-file="plugin.xml">org.intellij.scala</depends>-->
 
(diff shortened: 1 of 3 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -17,7 +17,7 @@
     </change-notes>
 
     <depends>com.intellij.modules.java</depends>
-    <depends optional="true" config-file="plugin.xml">org.intellij.scala</depends>
+    <depends optional="true" config-file="">org.intellij.scala</depends>
```
OlliKiljunen: Was that `config-file` attribute unnecessary in the first place or does this change something?

superseacat: surprisingly it does... with the attribute non-empty, the `verifyPlugin` task of Gradle IJ plugin failed with stackOverFlow error, whereas this way it works (tests pass, `runIde` starts, _etc._). It's a bit weird that this appeared only when I merged to `test`, however, this change solves the issue.

superseacat: magic...

PR #99, opened 2020-03-17
Sdk install

OlliKiljunen:
# Description

Please tell me if you want unit tests and for which units.

# Testing

- [x] All unit tests pass
- [ ] I have tested manually

- [ ] Yes
- [x] Not relevant


src/main/java/fi/aalto/cs/apluscourses/intellij/model/CommonLibraryProvider.java
```diff
@@ -0,0 +1,37 @@
...
+    this.project = project;
+    commonLibraries = new ConcurrentHashMap<>();
+  }
+
+  @NotNull
+  @Override
+  public Component getComponent(@NotNull String componentName) throws NoSuchModuleException {
+    Library library = commonLibraries.computeIfAbsent(componentName, this::createLibrary);
+    if (library == null) {
+      throw new NoSuchModuleException(componentName, null);
```
superseacat: a bit misleading... why a component `throw new NoSuchModuleException`?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CommonLibraryProvider.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+    return library;
+  }
+
+  private Library createLibrary(String libraryName) {
+    if (libraryName.startsWith("scala-sdk-")) {
```
superseacat: :innocent:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CommonLibraryProvider.java
```diff
@@ -0,0 +1,37 @@
...
+  @NotNull
+  @Override
+  public Component getComponent(@NotNull String componentName) throws NoSuchModuleException {
+    Library library = commonLibraries.computeIfAbsent(componentName, this::createLibrary);
+    if (library == null) {
+      throw new NoSuchModuleException(componentName, null);
+    }
+    return library;
+  }
+
```
superseacat: `@Nullable`?

src/test/java/fi/aalto/cs/apluscourses/model/ModelExtensions.java
```diff
@@ -71,13 +76,19 @@ public void updateState() {
...
     }
 
     @Override
     public Module createModule(@NotNull String name, @NotNull URL url) {
       return new TestModule(name, url);
     }
+
+    @Override
+    public Library createLibrary(@NotNull String name) {
+      return null;
```
superseacat: https://github.com/Aalto-LeTech/intellij-plugin/blob/0df3dabd8ba864a7f5a6661ade669bf704448b68/src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java#L76

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CommonLibraryProvider.java
```diff
@@ -0,0 +1,37 @@
...
+  private final ConcurrentMap<String, Library> commonLibraries;
+
+  public CommonLibraryProvider(Project project) {
+    this.project = project;
+    commonLibraries = new ConcurrentHashMap<>();
+  }
+
+  @NotNull
+  @Override
+  public Component getComponent(@NotNull String componentName) throws NoSuchModuleException {
```
superseacat: `getLibrary(...)`?

src/main/java/fi/aalto/cs/apluscourses/model/InstallerImpl.java
```diff
@@ -0,0 +1,139 @@
...
+    this.componentSource = componentSource;
+    this.taskManager = taskManager;
+  }
+
+  protected T waitUntilLoadedAsync(Component component) {
+    return taskManager.fork(() -> component.stateMonitor.waitUntil(Component.LOADED));
+  }
+
+  protected T installInternalAsync(Component component) {
+    Installation moduleInstallation = new Installation(component);
```
superseacat: `componentInstallation`?

src/main/java/fi/aalto/cs/apluscourses/model/InstallerImpl.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+    
+    public void doIt() {
+      try {
+        fetch();
+        initDependencies();
+        load();
+        waitForDependencies();
+      } catch (IOException | ComponentLoadException | NoSuchModuleException e) {
+        logger.info("A module could not be installed", e);
```
superseacat: maybe tweak a message a bit?

src/main/java/fi/aalto/cs/apluscourses/model/InstallerImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        component.stateMonitor.waitUntil(Component.INSTALLED);
+      }
+    }
+
+    private void initDependencies() throws ComponentLoadException, NoSuchModuleException {
+      if (dependencies != null) {
+        return;
+      }
+      List<String> dependencyNames = component.getDependencies();
+      dependencies = new ArrayList<>(dependencyNames.size());
```
superseacat: :open_mouth:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
```diff
@@ -0,0 +1,87 @@
...
+  @Override
+  public void load() throws ComponentLoadException {
+    try {
+      WriteAction.runAndWait(this::loadInternal);
+    } catch (MalformedURLException e) {
+      throw new ComponentLoadException(this, e);
+    }
+  }
+
+  private void loadInternal() throws MalformedURLException {
```
superseacat: IDEA says never throws

src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
```diff
@@ -0,0 +1,87 @@
...
+    com.intellij.openapi.roots.libraries.Library.ModifiableModel modifiableModel =
```
superseacat: static import?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
```diff
@@ -0,0 +1,87 @@
...
+  private void fetchZipTo(File file) throws IOException {
+    FileUtils.copyURLToFile(new URL(URL + getFileName() + ".zip"), file);
+  }
+
+  private void extractZip(File file) throws IOException {
+    new DirAwareZipFile(file).extractDir(getFileName() + "/lib", getBasePath());
+  }
+
+  private String getFileName() {
+    return getName().replace("scala-sdk", "scala");
```
superseacat: why?

superseacat: `@NotNull`?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CommonLibraryProvider.java
```diff
@@ -0,0 +1,37 @@
...
+    Library library = commonLibraries.computeIfAbsent(componentName, this::createLibrary);
+    if (library == null) {
+      throw new NoSuchModuleException(componentName, null);
+    }
+    return library;
```
superseacat: not a big fan of ternaries, 
`return module != null ? module : commonLibraryProvider.getComponent(componentName);`
how do you @OlliKiljunen @nikke234 feel?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJCourse.java
```diff
@@ -1,23 +1,39 @@
...
+                        @NotNull Map<String, String> requiredPlugins,
+                        @NotNull Project project) {
+    super(name, modules, libraries, requiredPlugins);
 
     this.project = project;
+    this.commonLibraryProvider = new CommonLibraryProvider(project);
+    this.modulesByNames = modules
+        .stream()
+        .map(IntelliJModule.class::cast)
+        .collect(Collectors.toMap(IntelliJModule::getName, Function.identity()));
```
superseacat: a note for myself: [Function._identity()_](https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html#identity--)

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CommonLibraryProvider.java
OlliKiljunen: Forgot to rename that. I'll do that

src/test/java/fi/aalto/cs/apluscourses/model/ModelExtensions.java
OlliKiljunen: Good point! That would be better here too.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CommonLibraryProvider.java
OlliKiljunen: Is this pointing to a wrong line-of-code. I don't see how...?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CommonLibraryProvider.java
OlliKiljunen: Possibly, yes...

src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
OlliKiljunen: "Why?" Scala plugin names Scala SDK libraries `scala-sdk-*`. When, in this approach, ZIPs are downloaded from scala-lang.org they are structured like
```
+/
+-+scala-12.3.1
  +-+lib
  | +--scala-compiler.jar
  | +--scala-library.jar
  | +--scala-reflect.jar
  +-+other
  +-+dirs
```
so there is the difference

src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
OlliKiljunen: Yeah, forgot that there. Good catch!

src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
OlliKiljunen: I thought that Java would have something similar to C#'s `using System.Console=SysConsole` (if I remember the syntax right). Turned out there is no importing with aliases in Java. Perhaps we should rename our `Module` and `Library` to `APlusModule` and `APlusLibrary` or something. Perhaps not now though.

src/main/java/fi/aalto/cs/apluscourses/model/InstallerImpl.java
OlliKiljunen: Oops!

src/main/java/fi/aalto/cs/apluscourses/model/InstallerImpl.java
OlliKiljunen: Yeah, you're absolutely right again.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
jsorva: Disclaimer: I did not actually read any of the code or the earlier messages, so I'm just replying to what I'm seeing in Olli's latest message, which I happened to notice.

In Scala, you can do `import foo.{bar => fubar}`, which may or may not be what is needed here.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CommonLibraryProvider.java
superseacat: no, the code is perfectly fine, just proposed using a ternary operator

src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
superseacat: sorry, was not clear enough
> Scala plugin names Scala SDK libraries `scala-sdk-*` 

didn't know that there is this part in lib names

src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
superseacat: I support the idea, as there are already `lang.Module` and IJ's own `Module` class (standing for IDEA module)

TESTING.md
```diff
@@ -149,13 +153,7 @@ a manual testing manual. Marked with a github issue label "manual testing".
...
         than 10&nbsp;seconds, consider it an error.</sub>
       </li>
-      <li>Select <em>GoodStuff</em> from the list and double click it.</li>
-      <li>
-        Ensure that <em>GoodStuff</em> and <em>O1Library</em> appear as loaded modules in the
-        project tree, and their contents match the image below:<br/>
-        <img src="images/30_module_loaded.png" alt="GoodStuff and O1Library contents" /><br/>
-        <sub>It shouldn't take longer than 5 seconds for modules to be imported.</sub>
-      </li>
+      <li>Select <em>GoodStuff</em> from the list and double click it to install the module.</li>
```
OlliKiljunen, line 156: I suggest that we do not re-ensure things that are already checked in other parts of the test procedure.

superseacat: I agree. another idea I had was to create some kind of a single longer script that would test all the features wo repeatings (thus, this will need to logically restructure things and thinking a bit on this).

build.gradle
```diff
@@ -21,11 +21,12 @@ repositories {
 dependencies {
     compile group: 'commons-io', name: 'commons-io', version: '2.6'
     compile group: 'net.lingala.zip4j', name: 'zip4j', version: '2.3.0'
-    compile group: 'org.scala-lang', name: 'scala-library', version: '2.12.10'
+    compileOnly group: 'org.scala-lang', name: 'scala-library', version: '2.12.10'
```
superseacat, line 24: I'm a bit skeptical when the configs have to be changed due to a "does not work on my machine"

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJLibrary.java
```diff
@@ -0,0 +1,58 @@
...
+import com.intellij.openapi.roots.impl.libraries.LibraryEx;
+import com.intellij.openapi.roots.libraries.LibraryProperties;
+import com.intellij.openapi.roots.libraries.LibraryTable;
+import com.intellij.openapi.roots.libraries.PersistentLibraryKind;
+import fi.aalto.cs.apluscourses.model.Library;
+import java.nio.file.Path;
+import org.jetbrains.annotations.CalledWithWriteLock;
+import org.jetbrains.annotations.NotNull;
+
+public abstract class IntelliJLibrary<
```
superseacat, line 14: a nasty line wrap

build.gradle
OlliKiljunen: There is definitely an issue if Scala libraries are loaded both by our plugin and Scala plugin. The dubious part of the original config is `compile` keyword which is deprecated in Gradle in favor of keywords `api`/`implementation`/`compileOnly` whose behavior is more easily understandable and straightforward. I guess that in some setups, `compile` works here but causes troubles in others. Instead of trying to understand the quirks of `compile`, it is better to replace it with a more specific keyword.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJLibrary.java
OlliKiljunen: What would be the correct place? Written on one line it reads:
```java
public abstract class IntelliJLibrary<K extends PersistentLibraryKind<? extends LibraryProperties<S>>, S> extends Library {
//                                                                                                  ^ line wrap should be before this char
```
Perhaps just:
```java
public abstract class IntelliJLibrary
    <K extends PersistentLibraryKind<? extends LibraryProperties<S>>, S> extends Library {
```

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJLibrary.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public IntelliJLibrary(@NotNull String name, @NotNull APlusProject project, int state) {
+    super(name, state);
+    this.project = project;
+  }
+
+  @CalledWithWriteLock
+  private void loadInternal() {
+    LibraryTable.ModifiableModel libraryTable = project.getLibraryTable().getModifiableModel();
+    com.intellij.openapi.roots.libraries.Library.ModifiableModel library = libraryTable
```
superseacat, line 28: not something to address now, but I suggest we consider codebase-wide change, where we somehow (by convention) distinct entities that are special to A+ from the same-named IntellijJ's ones (or framework ones, like Swing's):
`Project` (IJ) & `APlusProject`
`Module` (IJ) & `APlusModule`
`Component` (Swing) & `APlusComponent`
`Library` (IJ) & `APlusLibrary`
we could also take into account the fact, that we have **Model** entities, that are specific to IJ platform (like `IntelliJModule` etc.), so to have a uniform approach to this matter

src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
```diff
@@ -0,0 +1,104 @@
...
+    extractZip(tempZipFile);
+  }
+
+  @NotNull
+  private File createTempFile() throws IOException {
+    return FileUtilRt.createTempFile(getFileName(), ".zip");
+  }
+
+  private void fetchZipTo(File file) throws IOException {
+    FileUtils.copyURLToFile(new URL(URL + getFileName() + ".zip"), file);
```
superseacat, line 62: this might be a good addition to some of the methods here: https://github.com/Aalto-LeTech/intellij-plugin/blob/7b9f14b1efe246bd1da75f921d3d1dca0eac07bd/src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java#L50

src/main/java/fi/aalto/cs/apluscourses/intellij/model/APlusProject.java
```diff
@@ -0,0 +1,125 @@
...
+
+  public int resolveLibraryState(String name) {
+    return new LibraryStateResolver(name).resolveState();
+  }
+
+  public int resolveModuleState(String name) {
+    return new ModuleStateResolver(name).resolveState();
+  }
+
+  private abstract class ComponentStateResolver {
```
nikke234, line 59: In this case this abstraction adds quite a bit of complexity as opposed to just having the state resolving code in `resolveLibraryState` and `resolveModuleState` with some minor code duplication. Could be useful though if we add different types of components later

OlliKiljunen: Yeah, now that I look this, it's terribly complex. As far as I can see, there ain't gonna be other components than libraries and modules, so following YAGNI principle, I'll adopt your suggestion here.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/APlusProject.java
```diff
@@ -0,0 +1,125 @@
...
+import java.nio.file.Paths;
+import java.util.Objects;
+import org.jetbrains.annotations.NotNull;
+
+public class APlusProject {
+  private final Project project;
+
+  public APlusProject(Project project) {
+    this.project = project;
+  }
```
nikke234, line 19: `project` could be annotated with `@NotNull`

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
```diff
@@ -3,43 +3,52 @@
...
                              @NotNull List<Module> modules,
+                             @NotNull List<Library> libraries,
                              @NotNull Map<String, String> requiredPlugins,
                              @NotNull Map<String, URL> resourceUrls) {
-    IntelliJCourse course
-        = new IntelliJCourse(name, modules, requiredPlugins, resourceUrls, project);
+    IntelliJCourse course =
+        new IntelliJCourse(name, modules, libraries, requiredPlugins, resourceUrls, project);
     // Add a module change listener with the created course instance to the project
+    // TODO: These should be handled in those classes; also, renaming issue?
```
nikke234, line 39: What should be handled and in which classes? Do you mean the listeners should be added in the classes themselves, e.g. `IntelliJCourse` or...?

OlliKiljunen: Yeah, that was my intention. I think it's problematic if the responsibilities to listen all these thing are loaded to this single method. Other ideas?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CommonLibraryProvider.java
```diff
@@ -0,0 +1,43 @@
...
+import fi.aalto.cs.apluscourses.model.Component;
+import fi.aalto.cs.apluscourses.model.ComponentSource;
+import fi.aalto.cs.apluscourses.model.Library;
+import fi.aalto.cs.apluscourses.model.NoSuchComponentException;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+class CommonLibraryProvider implements ComponentSource {
```
nikke234, line 12: What is a common library?

src/main/java/fi/aalto/cs/apluscourses/model/Installer.java
```diff
@@ -0,0 +1,16 @@
+package fi.aalto.cs.apluscourses.model;
+
+import java.util.List;
+import org.jetbrains.annotations.NotNull;
+
+public interface Installer {
```
nikke234, line 6: `ComponentInstaller` might be clearer name for this

src/main/java/fi/aalto/cs/apluscourses/model/InstallerImpl.java
nikke234: Just adding a reminder :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CommonLibraryProvider.java
OlliKiljunen: A library that is not provided by the course but it is assumed that the plugin knows where to get it. Our only library `ScalaSdk` is a common library.

Generally, if a course-wide library (which is not a module like O1Lib but binary) is included in the project, the course configuration file should hold the information (URL) how to get it (we don't currently have support to this, but it is taken into account in design). Some libraries, i.e. common libraries, are though handled separately: it is enough that they are referenced in IML files, and our plugin knows where to get them. Scala SDK is (so far only) this kind of library.

src/main/java/fi/aalto/cs/apluscourses/model/NoSuchComponentException.java
```diff
@@ -3,8 +3,8 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-public class NoSuchModuleException extends Exception {
-  public NoSuchModuleException(@NotNull String moduleName, @Nullable Throwable cause) {
+public class NoSuchComponentException extends Exception {
+  public NoSuchComponentException(@NotNull String moduleName, @Nullable Throwable cause) {
     super("Module '" + moduleName + " was not found'.", cause);
```
nikke234, line 8: Could change the parameter to `componentName` and the message to `"Component '" + copmonentName + ...`

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CommonLibraryProvider.java
nikke234: Makes sense, thanks for the explanation! And is the `Library` class the common base class for both course libraries and common libraries? If so, should it also have a URL?

OlliKiljunen: Yes, `Library` is a base class but... as common libraries "do not exist" in course configuration file there is not common/course library separation outside the IntelliJ specific part of our code.

`IntelliJLibrary` is also a base class for both common and course libraries; it is there in-between. (See the UML chart made by Nikolai. It is also included in this PR but it can be only viewed with IDEA Ultimate.)

It is not yet designed (as it is not really needed in O1) how course libraries are represented in course configuration file (one URL or many URLs or something else?) so I wouldn't add URL field to `Library` class. Also, it might be that in future there are common libraries whose binaries are e.g. shipped with our plugin so they wouldn't have a URL at all. For these reasons, I would keep URL out of the base class.

Annotation_2020-04-15_154026

nikke234: Okay, sounds good :+1: By the way great chart @superseacat!

superseacat: thanks, but it's auto-generated, take a look at: [IntelliJ IDEA's manual](https://www.jetbrains.com/help/idea/class-diagram.html) (the main challenge is to pick the right scope to display) :stuck_out_tongue:

PR #116, opened 2020-04-01
REPL with configuration dialog

superseacat:
# Description
`ReplAction.scala` and the corresponding tests are still being in progress

# Testing

- [x] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant


src/main/java/fi/aalto/cs/apluscourses/ui/repl/ModuleComboBoxListRenderer.java
```diff
@@ -0,0 +1,24 @@
...
+import javax.swing.JList;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Custom renderer for items (modules in this case) stored in {@link javax.swing.JComboBox}.
+ */
+public class ModuleComboBoxListRenderer extends SimpleListCellRenderer<String> {
+
+  //todo add the icon path when developing locally
+  public static final String iconPath = "/u/39/denissn2/unix/IdeaProjects/intellij-plugin"
```
superseacat, line 14: right, this breaks!

src/test/java/fi/aalto/cs/apluscourses/TestHelper.java
```diff
@@ -17,7 +22,7 @@
 /**
  * A helper class to simplify testing plugin manipulation logics.
  */
-public abstract class PluginsTestHelper extends BasePlatformTestCase {
+public abstract class TestHelper extends BasePlatformTestCase {
```
superseacat, line 25: renamed and joined. there is now one aggregate test helper class that extends the `lightweight` test class of IJ, hopefully you don't mind :wink:

src/main/java/fi/aalto/cs/apluscourses/ui/repl/ReplConfigurationForm.java
```diff
@@ -0,0 +1,128 @@
...
+
+public class ReplConfigurationForm extends JPanel {
+
+  public static final String INFOLABEL_TEXT = "<html>"
+      + "This is an A+ Scala REPL configuration window. By default, the working "
+      + "directory and the dependencies, loaded to the REPL classpath belong to the Module, that "
+      + "it was started on. To change the behavior, use the checkbox at the bottom of the window."
+      + "</html>";
+
+  private ReplConfigurationFormModel model;
```
superseacat, line 24: @OlliKiljunen can you take a look here and at the overall composition. I'm using a model for data inside the UI (kinda Presentation), but is this a correct way?

OlliKiljunen: To me this seems very good.

I'd argue that there is no Model-View-Presenter/ViewModel pattern here, but that is by no means a problem because this view does not actually handle our domain model classes. Instead, the pattern used here is sometimes called an "Autonomous View" and as far as I understand, this implementation of yours could be a textbook example of it done right. :)

superseacat: thanks...
so this sh*t (implementation) even has a name :rofl: , I'll look it up

superseacat:
Olli's changes are in, code is now more or less ready

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/ReplAction.scala
```diff
@@ -86,7 +86,7 @@ class ReplAction extends RunConsoleAction {
     //  virtual file is working for both: files and folders
     val targetFileOrFolder = CommonDataKeys.VIRTUAL_FILE.getData(dataContext)
 
-    if (project == null || checkFileOrFolderIsNull(targetFileOrFolder)) return
+    if (project == null || checkFileOrFolderIsNull(targetFileOrFolder)) return // scalastyle:ignore
```
OlliKiljunen, line 89: I think `return` statement is fine here. I guess these kind of `void`/`Unit` type methods are rarely used in (functional) Scala so need for `return`s does not occur often when coding pure Scala. As the design of this class is, however, based on its Java base class, it is natural that the code is structured Java/imperative way. Luckily, Scala supports this coding paradigm too (even though it might not be that "Scala-like")

src/main/java/fi/aalto/cs/apluscourses/ui/repl/ReplConfigurationForm.java
```diff
@@ -0,0 +1,137 @@
...
+
+  private TextFieldWithBrowseButton workingDirectoryField;
+  private ComboBox moduleComboBox;
```
nikke234, line 28: This causes some compiler warnings from other lines in this file. Changing it to `ComboBox<String>` fixes the issue I think.

src/main/java/fi/aalto/cs/apluscourses/ui/repl/ReplConfigurationForm.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  public JLabel getInfoTextLabel() {
+    return infoTextLabel;
+  }
+
+  public void setInfoTextLabel(JLabel infoTextLabel) {
+    this.infoTextLabel = infoTextLabel;
+  }
+}
```
nikke234, line 137: Could remove all of these `setX` methods since they are unused (unless they are used by the GUI designer).

src/main/java/fi/aalto/cs/apluscourses/ui/base/DialogBaseHelper.java
```diff
@@ -0,0 +1,46 @@
...
+import javax.swing.JButton;
+import javax.swing.JComponent;
+import javax.swing.JDialog;
+import javax.swing.JPanel;
+import javax.swing.KeyStroke;
+
+/**
+ * The contents of this class are fully auto-generated by the IntelliJ IDEA.
+ */
+public class DialogBaseHelper extends JDialog {
```
nikke234, line 15: I assume there's no point in fixing sonar's complaints here if this is auto-generated?

src/main/java/fi/aalto/cs/apluscourses/presentation/ReplConfigurationFormModel.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.stream.Collectors;
+import org.jetbrains.annotations.NotNull;
+
+public class ReplConfigurationFormModel {
+
+  private Project project;
+  private String moduleWorkingDirectory;
+  private String targetModuleName;
+  private List<String> moduleNames;
+  public boolean startRepl = true;
```
nikke234, line 18: Nit: this should probably be private since there is already a `setStartRepl` method

nikke234, line 18: Nit: this should probably be private since there is a `setStartRepl` method

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/ReplAction.scala
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    val configuration = setting.getConfiguration.asInstanceOf[ScalaConsoleRunConfiguration]
+    try {
+      setConfigurationConditionally(project, module, configuration)
+    } catch {
+      case _: CancelReplStartException => return // scalastyle:ignore
+    }
+
+    RunConsoleAction.runExisting(setting, runManagerEx, project)
```
nikke234, line 111: I would personally avoid using exceptions for control flow. I have added a PR with an alternative possibility, but feel free to disagree of course :smile:

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/ReplAction.scala
nikke234: Maybe we need a third style checker that looks for ignore-comments :trollface:

src/main/java/fi/aalto/cs/apluscourses/ui/base/DialogBaseHelper.java
superseacat: not really

superseacat:
@nikke234 , @OlliKiljunen , thanks for all the comments, taken into account

PR #120, opened 2020-04-07
Wrote missing tests for Resources class, minor refactoring

OlliKiljunen:
This is how I'd write these tests.


superseacat:
and I hope you don't mind me incorporating this approach

PR #122, opened 2020-04-07
Refactor ReplAction#setConfigurationConditionally

nikke234:
An alternative possibility without exceptions


src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/ReplAction.scala
```diff
@@ -43,25 +42,30 @@ class ReplAction extends RunConsoleAction {
...
     if (ReplConfigurationFormModel.showREPLConfigWindow) {
-      val configModel = new ReplConfigurationFormModel(project,
-        getModuleWorkDir(module), module.getName)
+      val configModel = new ReplConfigurationFormModel(project, getModuleWorkDir(module), module.getName)
 
       createAndShowReplConfigurationDialog(configModel)
 
       if (!configModel.isStartRepl) {
-        throw new CancelReplStartException
+        false
```
superseacat, line 58: I agree, this is a good approach as well :pushpin:

PR #125, opened 2020-04-08
Import IDE Settings Tests

nikke234:
# Description
Adds unit tests for the IDE settings import action.

# Testing

- [x] I created new unit tests
- [x] All unit tests pass
- [x] I have tested manually

# Have you updated the README or other relevant documentation?

- [ ] Yes
- [ ] Not yet. I will do it next.
- [x] Not relevant

PR #126, opened 2020-04-08
Import Project Settings

nikke234:
# Description
Adds a project settings import action. This is based off of #125, which should be merged first.

# Testing

- [x] All unit tests pass
- [x] I have tested manually

- [x] Yes
- [ ] Not relevant


src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ImportProjectSettingsAction.java
```diff
@@ -0,0 +1,134 @@
...
+  public ImportProjectSettingsAction() {
+    this(PluginSettings.getInstance(), ImportProjectSettingsAction::doImport,
+        new IntelliJDialogs());
+  }
+
+  @Override
+  public void actionPerformed(@NotNull AnActionEvent e) {
+    Project project = e.getProject();
+    if (project == null) {
+      // TODO: override AnAction#update instead and only enable this action if a project is loaded
```
nikke234, line 57: Oops, will fix this

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/AboutAction.java
```diff
@@ -2,10 +2,11 @@
 
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.project.DumbAware;
 import fi.aalto.cs.apluscourses.ui.AboutDialog;
 import org.jetbrains.annotations.NotNull;
 
-public class AboutAction extends AnAction {
+public class AboutAction extends AnAction implements DumbAware {
```
OlliKiljunen, line 9: This `DumbAware` thing is something that I have been thinking I should study more about because I guess there are some other parts of our code that should be `DumbAware`. If you have robust understanding how does it work and how it should be used, I'd be keen to get some insights. :)

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ImportProjectSettingsAction.java
```diff
@@ -0,0 +1,133 @@
...
+    URL settingsUrl = course.getResourceUrls().get("projectSettings");
+    if (settingsUrl == null) {
+      informNoSettings(course);
+      return;
+    }
+
+    try {
+      projectSettingsImporter.doImport(project, settingsUrl);
+    } catch (IOException | UnexpectedResponseException ex) {
+      informErrorOccurred();
```
superseacat, line 79: Nit: we probably need to come our with a common approach for exception handling (especially taking into account recent discoveries when debugging the IDE) apart from users notifications

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ImportProjectSettingsAction.java
```diff
@@ -0,0 +1,133 @@
...
+        .stream()
+        .filter(file -> !file.isDirectory())
+        .map(FileHeader::getFileName)
+        .collect(Collectors.toList());
+
+    for (String fileName : fileNames) {
+      Path path = Paths.get(fileName);
+      Path pathWithoutRoot = path.subpath(1, path.getNameCount());
+      zipFile.extractFile(path.toString(), settingsPath.toString(), pathWithoutRoot.toString());
+    }
```
superseacat, line 107: Nit: this can't last forever, we need to talk... https://www.vavr.io/vavr-docs/ ?

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ImportProjectSettingsAction.java
```diff
@@ -0,0 +1,133 @@
...
+      informErrorOccurred();
+    }
+  }
+
+  /**
+   * Downloads the project settings ZIP file from the given URL to a temporary file. After that the
+   * files from the .idea directory of the ZIP file are extracted to the .idea directory of the
+   * given project, after which the project is reloaded.
+   */
+  private static void doImport(@NotNull Project project, @NotNull URL settingsUrl)
```
superseacat, line 88: I wonder if this could be split into smaller methods?

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/ImportProjectSettingsActionTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+  private Project project;
+  private MainViewModel mainViewModel;
+  private AnActionEvent anActionEvent;
+
+  /**
+   * Called before each test method call.  Initializes private fields.
+   */
+  @Before
+  public void createMockObjects() throws MalformedURLException {
```
superseacat, line 36: there are some minor complaints on "duplication" from Sonar, but they can easily be pushed to that refactoring stage

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ImportProjectSettingsAction.java
OlliKiljunen: Though I agree with you on that, this implementation seems mostly appropriate to me. It is in a method that is called by the platform, so it cannot be thrown any further. Of course, some information of the exception could be used to tailor a more informative/specific error message and stack trace could be printed to the log. I also like `Dialogs` interface and `showErrorDialog()` method there. Perhaps that approach could be abstracted even further so that we had some kind of `ErrorHandler` interface, an implementation of which would handle both logging and informing user.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ImportProjectSettingsAction.java
OlliKiljunen: Can you elaborate? I understand that `fileNames` could be processed e.g. via Streams API to contain the data in the expected form but that seems at most marginal improvement. Is there more that you see problematic here?

superseacat: the problem is, that `fileNames` actually can't be processed within the same Stream, because I think the `....extractFile(...)` throws an exception. and exceptions can neither be handled in Streams nor be pushed to method signature, so one has to break the Stream and introduce a old-fashioned for-loop to handle the situation. I've been into this several times myself, so that's why the vague comment, sorry. so there are some tools (like Vavr) to work with those cases, but here it's more of a discussion-starter...

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ImportProjectSettingsAction.java
superseacat: the implementation is good, indeed, and this idea would probably make it perfect:
> Perhaps that approach could be abstracted even further so that we had some kind of `ErrorHandler` interface, an implementation of which would handle both logging and informing user.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/AboutAction.java
nikke234: Implementing `DumbAware` makes the action available during indexing (that long process that happens when you start the IDE). As long as the action doesn't access the indices, it can/should implement `DumbAware`. The problem is that I'm not 100% sure which methods from the API need the indices and which don't, but the good news is that an exception is thrown if the indices are accessed before they are ready :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ImportProjectSettingsAction.java
nikke234: Good idea :+1:

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/ImportProjectSettingsActionTest.java
nikke234: Sounds good, I'll add a note of this

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/AboutAction.java
OlliKiljunen: I created a new ticket of this #129

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ImportProjectSettingsAction.java
nikke234: Excellent discussion starter - I constantly run into this issue of checked exceptions making streams a pain in the ... to use :smile: Luckily in this case the result is still fine. Thanks for the interesting link :+1:

OlliKiljunen: For this discussion: #131

PR #128, opened 2020-04-15
tiny enhancement proposal to Sdk install

superseacat:
just a simple diagram, as it's a quicker way to figure out the structure


OlliKiljunen:
Thanks, this is great!

PR #131, opened 2020-04-15
Do-Not-Merge: Test implementation of iterating streams with exceptions

OlliKiljunen:
# Description

Just a quick POC for this discussion https://github.com/Aalto-LeTech/intellij-plugin/pull/126#discussion_r408089430

I don't actually feel this is necessary: for loops are much cleaner. What do you think?


nikke234:
This only works if the function throws at most one type of checked exception right?

OlliKiljunen:
> This only works if the function throws at most one type of checked exception right?

Well, kinda. `E` must be a supertype for all the exceptions that are thrown by the consumer function. So if in the test `interruptIfZero()` could also throw an IOException then the third parameter to `StreamUtil.forEach()` should be just `Exception.class`

PR #132, opened 2020-04-15
sample tests for part of sdk-install 

superseacat:
most of the tests I've been making recently are not actually unit tests, but of this weird platform test type, thus, they somehow still avail testing what I wanted.


superseacat:
if you can wait just a bit more, I'll review this tomorrow with a fresh eye and remove the "do-not-merge" so you could re-review :wink:

superseacat:
good enough, thus some issues externalized to `sdk-install-test-samples-shelved`

PR #135, opened 2020-04-21
Mark modules invalid on

superseacat:
# Description

# Testing

- [x] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant


superseacat:
anyone knows where those PR templates are coming from :confused: ? @OlliKiljunen @nikke234

OlliKiljunen:
> anyone knows where those PR templates are coming from :confused: ? @OlliKiljunen @nikke234 

I thought that you configured it. 😃 I have no idea. 🤷‍♂️

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
```diff
@@ -79,4 +88,48 @@ public Library createLibrary(@NotNull String name) {
...
+  public List<String> getCourseModuleDependencies(@NotNull Module courseModule) {
+    List<String> courseModuleDependencies = null;
+    try {
+      courseModuleDependencies = courseModule.getDependencies();
+    } catch (ComponentLoadException ex) {
+      logger.error(ex.getMessage(), ex);
+    }
+    return courseModuleDependencies;
+  }
 }
```
OlliKiljunen, line 135: This is fine :+1: though I am thinking that `Module#getDependecies` should actually be changed so that it is called only after the module is loaded and instead of parsing IML file, it would use IDEA platform API to access dependency info. In that way, we could get rid of the exception altogether.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
```diff
@@ -79,4 +88,48 @@ public Library createLibrary(@NotNull String name) {
...
+   * marks their {@link Module}s {@link StateMonitor} state.
+   *
+   * @param course            a {@link Course} which {@link Module}s to check
+   * @param removedModuleName a {@link String} name of the removed
+   * {@link com.intellij.openapi.module.Module}
+   */
+  public void markDependentModulesInvalid(@NotNull IntelliJCourse course,
+      @NotNull String removedModuleName) {
+    APlusProject project = course.getProject();
+    com.intellij.openapi.module.Module[] projectModules = project.getModuleManager().getModules();
```
OlliKiljunen, line 104: Could (and should) this use our domain model classes instead IDEA platform API?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
```diff
@@ -29,19 +37,20 @@ public IntelliJModelFactory(@NotNull Project project) {
...
         new IntelliJCourse(name, modules, libraries, requiredPlugins, resourceUrls, project);
     // Add a module change listener with the created course instance to the project
     // TODO: These should be handled in those classes; also, renaming issue?
     project.getMessageBus().connect().subscribe(ProjectTopics.MODULES, new ModuleListener() {
       @Override
       public void moduleRemoved(@NotNull Project project,
-                                @NotNull com.intellij.openapi.module.Module projectModule) {
-          course.onComponentRemove(course.getComponentIfExists(projectModule.getName()));
+          @NotNull com.intellij.openapi.module.Module projectModule) {
+        markDependentModulesInvalid(course, projectModule.getName());
```
OlliKiljunen, line 52: Alternative approach would be that all modules are checked whenever there is a change in module configuration and also on start up. A bit more idiot proof, I'd say. But no need to change now. And how you implemented this is also a more elegant way, so it's fine.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
superseacat: but it does, isn't it?
`project`**`.getModuleManager()`** is a method of the APlusProject class ("our API") or what do you mean? :confused:

OlliKiljunen: Yeah, I meant it could ask `Course` of the modules that are known to it (that is, defined in the course config file). I don't know if that would be much better solution. It could be implemented in `Module` class though as it wouldn't be IDEA-specific.

superseacat:
2h buried here :skull:  (these calls: `getModuleDependentModules` and `getDependencyModuleNames` of the IJ's API simply do not work):
```java
    logger.error("before");
    com.intellij.openapi.module.Module moduleByName = ModuleManager.getInstance(project)
        .findModuleByName(removedModuleName);
    if (moduleByName != null) {
      List<com.intellij.openapi.module.Module> moduleDependentModules = ModuleManager
          .getInstance(project).getModuleDependentModules(moduleByName);
      for (com.intellij.openapi.module.Module module : moduleDependentModules) {
        logger.error(module.getName());
      }
    }
    Collection<ModuleDescription> allModuleDescriptions = ModuleManager.getInstance(project)
        .getAllModuleDescriptions();
    for (ModuleDescription moduleDescription : allModuleDescriptions) {
      List<String> dependencyModuleNames = moduleDescription.getDependencyModuleNames();
      for (String moduleName : dependencyModuleNames) {
        logger.error(moduleName);
      }
    }
    logger.error("after");

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
superseacat: I like the idea but suggest to postpone it a bit

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
superseacat: I'm a bit confused, as there are currently two different entities (thus, under the same name):
* `com.intellij.openapi.module.Module` and 
* `fi.aalto.cs.apluscourses.intellij.model.IntelliJModule`

ideally, those two gentlemen would know about each other somehow :cupid:

PR #136, opened 2020-04-23
setting for starting a REPL configuration window is persistent over r…

superseacat:
# Description

so I suggest to introduce an additional concept:
* IDE settings (remote, persistent)
* Project settings (remote, persistent)
* **NEW** Local settings (local, persistent) **NEW** - some settings that don't really need to be fetched from A+

otherwise, please review and comment. I'm not sure about not having a local representation of the `A+.showReplConfigDialog`, thus it kinda replaces the listener in this case (but I'm still thinking if the listener is required here).

# Testing

- [x] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
@@ -1,5 +1,8 @@
 package fi.aalto.cs.apluscourses.intellij.services;
 
+import static fi.aalto.cs.apluscourses.intellij.services.PluginSettings.LocalSettingsNames.A_PLUS_SHOW_REPL_CONFIGURATION_DIALOG;
+
+import com.intellij.ide.util.PropertiesComponent;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.project.ProjectManager;
@@ -12,8 +15,14 @@ import org.jetbrains.annotations.Nullable;
 
 public class PluginSettings implements MainViewModelProvider {
 
+  public interface LocalSettingsNames {
+
+    String A_PLUS_SHOW_REPL_CONFIGURATION_DIALOG = "A+.showReplConfigDialog";
+  }
+
   public static final String COURSE_CONFIGURATION_FILE_URL
       = "https://grader.cs.hut.fi/static/O1_2020/projects/o1_course_config.json";
+  private static final PropertiesComponent propertiesManager = PropertiesComponent.getInstance();
 
   @NotNull
   private final ConcurrentMap<Project, MainViewModel> mainViewModels = new ConcurrentHashMap<>();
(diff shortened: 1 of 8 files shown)


src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
```diff
@@ -12,8 +15,14 @@
 
 public class PluginSettings implements MainViewModelProvider {
 
+  public interface LocalSettingsNames {
```
OlliKiljunen, line 18: IIRC, Sonar would prefer this to be a (static) class. For me, it doesn't matter.

superseacat: hm... but the report is clean (at least it does not complain about it, but some other things), and it's an **interface**, does it make any difference?

OlliKiljunen: I meant that "Move constants to a class or enum" code smell. But surely it's a minor issue.

superseacat: I dislike this `ENUM.BLAAAAH.name()` approach to get the value and also the amount of code for `enum` exceeds so much the same for the `interface` approach, that I'd prefer to keep it for now :pleading_face:

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
```diff
@@ -46,8 +56,27 @@ public MainViewModel getMainViewModel(@Nullable Project project) {
     return mainViewModels.computeIfAbsent(project, this::createNewMainViewModel);
   }
 
+  @NotNull
   private MainViewModel createNewMainViewModel(@NotNull Project project) {
     ProjectManager.getInstance().addProjectManagerListener(project, projectManagerListener);
     return new MainViewModel();
   }
+
+  public static boolean isShowReplConfigurationDialog() {
```
nikke234, line 65: Nit: I would personally prefer something like `shouldShowReplConfigurationDialog`

superseacat: that's what you get when generating getter/setter for boolean with IJ, not optimal. I'll take into account in the future.

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
OlliKiljunen: Totally agree here.

I also kinda dislike enums in Java so let's forget them. What I was saying is that Sonar (for some imo-not-so-valid reason) likes it more if `LocalSettingsNames` is a class instead of interface (and it probably should have a private constructor too). Personally I see interface better here, just as you coded.

PR #138, opened 2020-04-23
Windows repl naming bug

superseacat:
# Description

it was a paths related problem, worked on Windows

# Testing

- [x] All unit tests pass
- [x] I have tested manually


src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/ReplAction.scala
```diff
@@ -27,7 +28,8 @@ class ReplAction extends RunConsoleAction {
 
   def getModuleWorkDir(@NotNull module: Module): String = {
     //  for project "root" this points to .../.../.idea folder
-    ModuleUtilCore.getModuleDirPath(module).replace("/.idea", "")
+    toSystemIndependentName(ModuleUtilCore.getModuleDirPath(module)
+      .replace("/.idea", ""))
```
OlliKiljunen, line 32: Should `replace()` be called for the system-independent name? I'm afraid `"/.idea"` does not match to any path in Windows because they don't have forward-slashes there. Or did I misunderstood?

PR #139, opened 2020-04-24
Update deps

superseacat:
# Description
so this now works for me and on clean Windows for regular scenarios. please try it too...

- `gradle assemble`
- `runIde` and click around

# Testing

- [x] All unit tests pass
- [x] I have tested manually


src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ActionGroups.java
```diff
@@ -3,6 +3,8 @@
 public class ActionGroups {
```
superseacat, line 3: could/should: `extends ActionGroup {` ?

Issue #110, opened 2020-03-20
Renaming/removing module files leads to incorrect states in module list


nikke234:
Removing/renaming module files causes `.idea/modules.xml` to be out of date. The IDE still thinks the modules are there, which makes the modules show as installed in the course modules list.

superseacat:
tried & worked:
- remove a module folder from OS -> 'error'
- delete a module from project view -> 'download to install'
- remove a library other modules depend on -> 'error in dependencies'
- reinstalling a library other modules depend on after deletion from project view -> makes depending modules valid again ('installed')

tried and unsure if I understand the idea:
- reinstalling a library other modules depend on after deletion from the OS -> makes dependants 'errored in deps', module itself (in module view) 'error'ed, does not reinstall; **a question:** should it reinstall?
- install some modules that depend on O1 (O1 is listed in modules' `.iml`'s); remove O1 (module statuses go into 'error in deps'), O1 itself goes 'error'; reinstall O1; module statuses get back to 'installed'. **a question:** should O1 return to modules' `.iml`? @OlliKiljunen 

non-related observation: `module.xml` takes time to refresh/reload

superseacat:
- delete a single `.jar` inside the scala SDK folder; things just stay as they are. **a question:** is it possible to make them maybe 'readOnly'?
- delete the whole scala SDK folder; O1 and other modules go 'error in deps'

PR #141, opened 2020-04-27
Module files removal

OlliKiljunen:
# Description

Issue #110. When removing a module/library (either by unloading it or deleting the directory), its status (along with the statuses of modules depending on it) gets updated in the modules list.

# Testing

- [x] All unit tests pass
- [x] I have tested manually

- [x] Yes
- [ ] Not relevant


src/main/java/fi/aalto/cs/apluscourses/intellij/model/APlusProject.java
```diff
@@ -40,43 +48,38 @@ public LibraryTable getLibraryTable() {
...
-  }
-
-  private int resolveComponentState(@Nullable Object componentObj, @NotNull Path dirPath) {
+  /**
+   * Returns a state in which the given component should be.
+   * @param component An IntelliJ specific component.
+   * @param <T> Type of the component.
+   * @return Component state.
+   */
+  public <T extends Component & IntelliJComponent<?>> int resolveComponentState(
```
superseacat, line 57: so...

1. `<?>` is https://docs.oracle.com/javase/tutorial/extra/generics/wildcards.html
2. `<T extends Component & ... > is https://docs.oracle.com/javase/tutorial/java/generics/boundedTypeParams.html

OlliKiljunen: I wanted the parameter `component` to be both `Component` and `IntelliJComponent`. If `IntelliJComponent` was a subtype of `Component` (as it semantically should be) it would be enough to have `public int resolveComponentState(IntelliJComponent<?> component)`.

The problem is that `Component` is an abstract class and `IntelliJComponent` is an interface, and Java doesn't allow interface to inherit an abstract class (that is, restrict interface to be implemented only by subclasses of `Component`, which actually would kinda make sense).

I guess the "most correct" way here would be making `Component` an interface too (and having a separate `ComponentBase` base class for implementations). Then `IntelliJComponent` could just extend `Component`.

OlliKiljunen: But yeah this solution (which abuses type parameters) is awkward, especially if there will be more methods like this in the future.

superseacat:
could you please merge `test` here and briefly verify :adventures_ahead:

superseacat:
can't find reference neither in sources nor in some conf files, probably missed some discussion, where (and why) we have a pleasure to meet those two ladies:

- jansi-1.12.jar
- jline-2.14.6.jar

superseacat:
added comment/question on behavior into the original issue: https://github.com/Aalto-LeTech/intellij-plugin/issues/110#issuecomment-625651479

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJCourse.java
```diff
@@ -16,27 +21,45 @@
...
+  public Component getComponentIfExists(@NotNull String name) {
+    Component component = super.getComponentIfExists(name);
+    return component != null ? component : commonLibraryProvider.getComponentIfExists(name);
+  }
+
   @Nullable
   public Component getComponentIfExists(VirtualFile file) {
     Component component = getComponentIfExists(file.getName());
-    return component != null && component.getPath().toString().equals(file.getPath())
+    return component != null && component.getFullPath().equals(Paths.get(file.getPath()))
```
superseacat, line 53: do you think there could be ways to improve the readability of `component.getFullPath().equals(Paths.get(file.getPath()))`?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJLibrary.java
```diff
@@ -58,4 +86,10 @@ public Path getPath() {
 
   @CalledWithWriteLock
   protected abstract void initializeLibraryProperties(LibraryProperties<S> properties);
+
+  @Override
+  @Nullable
+  public com.intellij.openapi.roots.libraries.Library getPlatformObject() {
```
superseacat, line 92: this probably could have a more explaining name, like `get...Library()`?, as it returns a `Library` and that's what's happening inside the method too.
also the part that is `...Platform...` why is it platform, what is sematical link between it and the `IntelliJLibrary` class?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/APlusProject.java
```diff
@@ -30,6 +32,12 @@ public ModuleManager getModuleManager() {
     return ModuleManager.getInstance(project);
   }
 
+  @Nullable
+  public ModuleRootManager getModuleRootManager(@NotNull String moduleName) {
```
superseacat, line 36: could this have a JavaDoc, please :wink:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/APlusProject.java
```diff
@@ -40,43 +48,38 @@ public LibraryTable getLibraryTable() {
...
+     *   0. The component is in the project but the files are missing -> ERROR.
      *   1. The component is in the project, so its state should be INSTALLED.
      *   2. The component is not in the project but the module files are present in the file
      *      system, so its state should be FETCHED.
      *   3. The component files aren't present in the file system (and by extension the component
      *      isn't in the project), so its state should be NOT_INSTALLED.
      */
-    if (componentObj != null) {
-      return Component.INSTALLED;
+    boolean loaded = component.getPlatformObject() != null;
```
superseacat, line 68: this is maybe a bit too abstract `component.getPlatformObject()`? :dizzy_face:

superseacat:
[](https://sonarcloud.io/component_measures?id=Aalto-LeTech_intellij-plugin&pullRequest=141&metric=new_coverage&view=list) [55.1% Coverage](https://sonarcloud.io/component_measures?id=Aalto-LeTech_intellij-plugin&pullRequest=141&metric=new_coverage&view=list)  

do you think improving test coverage for some methods of `intellij/model` package classes would be possible?

OlliKiljunen:
> can't find reference neither in sources nor in some conf files, probably missed some discussion, where (and why) we have a pleasure to meet those two ladies:
> 
>     * jansi-1.12.jar
> 
>     * jline-2.14.6.jar
> 
> 
> especially jansi 😉

I got problems with Scala REPL without them. Well, perhaps it was only jline that was required but that made me think that it's best to include them all to avoid other unforeseeable issues. They are small libraries though,

OlliKiljunen:
> 
> 
> [](https://sonarcloud.io/component_measures?id=Aalto-LeTech_intellij-plugin&pullRequest=141&metric=new_coverage&view=list) [55.1% Coverage](https://sonarcloud.io/component_measures?id=Aalto-LeTech_intellij-plugin&pullRequest=141&metric=new_coverage&view=list)
> 
> do you think improving test coverage for some methods of `intellij/model` package classes would be possible?

Yes. I'll work with them next.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJCourse.java
OlliKiljunen: `Paths.get(file.getPath()).equals(component.getFullPath())` ?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/APlusProject.java
OlliKiljunen: I'll add that.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJLibrary.java
OlliKiljunen: The point here is that all `IntelliJComponent<T>` objects have some "platform object" (of type `T`) that is their correspondent in the IntelliJ platform's object hierarchy.

IntelliJLibrary is a component of ours. Its existence is based in the in information in the course configuration file (hypothetically, as we haven't really implemented those kind of libraries yet, but they are kind of supported by our design) or information "hard-coded" to our plugin ("common libraries" as, and currently only, Scala SDK). IntelliJLibrary knows where to download its binaries and where to store them and such things.

The platform object in this case is a `com.intellij.openapi.roots.libraries.Library` object. It's an object created and managed by IntelliJ platform and its existence is tied to that specific library being loaded in the project.

So if our IntelliJModule is in LOADED state, there should be a "platform object" for it. That object is returned by this method.

Hope this makes sense.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/APlusProject.java
OlliKiljunen: I agree that the name "platform object" is a bit lame. Could we come up with a better word?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/APlusProject.java
superseacat, line 57: the methods has a dendroid logics and feels important, can it be somehow unit-tested?

OlliKiljunen: With some mocking, yes. I'll do! Perhaps even refactor this to another class, as it feels a bit out of place here, or whaddya think?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
```diff
@@ -41,23 +39,30 @@ public Course createCourse(@NotNull String name,
       @NotNull List<Library> libraries,
```
superseacat, line 39: keeps growing :heart_eyes:, do you think the whole method or the at least the methods introduced in the anonymous classes can be tested (as this might become **very** expensive to debug)?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
```diff
@@ -89,17 +71,36 @@ public void load() throws ComponentLoadException {
...
 
+  @Override
+  public void unload() {
+    super.unload();
+    Optional.ofNullable(getPlatformObject()).ifPresent(project.getModuleManager()::disposeModule);
+  }
+
+  @NotNull
+  @Override
+  protected List<String> computeDependencies() {
```
superseacat, line 82: `ctrl + shift + t` :honeybee: ?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModelFactory.java
OlliKiljunen: This calls for big refactoring IMO. Such refactoring that makes it logic unit testable. Now it's debt.

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -141,7 +142,12 @@ public String getName() {
    */
   @NotNull
   public List<Library> getLibraries() {
-    return libraries;
+    return Collections.unmodifiableList(libraries);
```
superseacat, line 145: do you think we should put more attention to the _modifiability_ of the collections we work with overall?

src/main/java/fi/aalto/cs/apluscourses/utils/FunctionUtil.java
```diff
@@ -0,0 +1,13 @@
+package fi.aalto.cs.apluscourses.utils;
+
+import java.util.function.IntPredicate;
+
+public class FunctionUtil {
```
superseacat, line 5: we probably have more custom predicates across the code, so far they were located close to where used, do you think we shall extract them too :question:

src/main/java/fi/aalto/cs/apluscourses/utils/StateMonitor.java
```diff
@@ -36,37 +37,47 @@ public StateMonitor(int state, @NotNull StateListener stateListener) {
...
+   * @return Whether or not the current state in the beginning was equal to any of
+   *         {@code expectedStates}.
+   * @throws IllegalArgumentException If {@code newState} is less than any of {@code expectedStates}
+   *                                  (and is not an error state).
    */
-  public boolean setConditionally(int expectedState, int newState) {
-    if (newState < expectedState) {
+  public boolean setConditionallyTo(int newState, int... expectedStates) {
+    if (!isError(newState)
+        && !Arrays.stream(expectedStates).allMatch(FunctionUtil.lessThanOrEqualTo(newState))) {
```
superseacat, line 65: can this `stream` part be extracted-tested?

src/test/java/fi/aalto/cs/apluscourses/utils/StateMonitorTest.java
```diff
@@ -61,7 +59,7 @@ public void testSetConditionallyReturnsFalse() {
     int expectedState = 20;
     int newState = 220;
     assertFalse("setConditionally() should return false",
-        stateMonitor.setConditionally(expectedState, newState));
+        stateMonitor.setConditionallyTo(newState, expectedState));
```
superseacat, line 62: should the method name here and across the class also be renamed into `testSetConditionally`**To**`ReturnsFalse()`?

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/ListDependenciesPolicy.java
```diff
@@ -0,0 +1,26 @@
+package fi.aalto.cs.apluscourses.intellij.utils;
+
+import com.intellij.openapi.roots.LibraryOrderEntry;
+import com.intellij.openapi.roots.ModuleOrderEntry;
+import com.intellij.openapi.roots.RootPolicy;
+import java.util.List;
+import org.jetbrains.annotations.NotNull;
+
+public class ListDependenciesPolicy extends RootPolicy<List<String>> {
```
superseacat, line 9: could you please JavaDoc the whole class with contents :pleading_face:

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/ListDependenciesPolicy.java
```diff
@@ -0,0 +1,26 @@
...
+
+  @Override
+  public List<String> visitModuleOrderEntry(@NotNull ModuleOrderEntry moduleOrderEntry,
```
superseacat, line 12: I understand what happens but can't relate to the method name, why a _visit_?

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/ListDependenciesPolicy.java
```diff
@@ -0,0 +1,26 @@
...
+                                            List<String> value) {
+    value.add(moduleOrderEntry.getModuleName());
+    return value;
+  }
+
+  @Override
+  public List<String> visitLibraryOrderEntry(@NotNull LibraryOrderEntry libraryOrderEntry,
+                                             List<String> value) {
+    if (!libraryOrderEntry.isModuleLevel()) {
+      value.add(libraryOrderEntry.getLibraryName());
```
superseacat, line 22: will the `value`'s check for `null` make sense here (or `@NotNull`)?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJComponent.java
```diff
@@ -0,0 +1,8 @@
+package fi.aalto.cs.apluscourses.intellij.model;
+
+import jdk.internal.jline.internal.Nullable;
```
nikke234, line 3: Could use jetbrains nullable annotation here for consistency

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
OlliKiljunen: Not necessarily, IMO. It is important that the design (or ultimately documentation) makes it clear which collections are meant to be modifiable by the client code and which are not. Wrapping them inside `Collections.unmodifiableSomething()` is just an extra runtime assertion for that, but programmers should be aware of it already when coding. (Well, using `Collections.unmodifiableList()` is of course one way to tell people looking at the code that "hey, btw, don't modify this list".)

src/main/java/fi/aalto/cs/apluscourses/utils/FunctionUtil.java
OlliKiljunen: No. Already this may be an overkill here. The logic where this was used was so complex that I tried to make it easier to read by putting parts of it in separate functions. I suggest that if, after a few sprints, this class still exists only for that single purpose, let's inline it. But perhaps there will be more methods/uses before that. So let's see.

src/test/java/fi/aalto/cs/apluscourses/utils/StateMonitorTest.java
OlliKiljunen: You're right.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJCourse.java
superseacat: I had some extraction and renaming in mind

src/main/java/fi/aalto/cs/apluscourses/model/ComponentInitializationCallback.java
```diff
@@ -0,0 +1,5 @@
+package fi.aalto.cs.apluscourses.model;
+
+public interface ComponentInitializationCallback {
```
nikke234, line 3: `@FunctionalInterface`?

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -42,6 +43,8 @@
   @NotNull
   private final Map<String, URL> resourceUrls;
 
+  private final AtomicBoolean resolved = new AtomicBoolean(true);
```
nikke234, line 46: This seems to be unused

src/main/java/fi/aalto/cs/apluscourses/utils/FunctionUtil.java
nikke234: In this case I would actually favor having the predicate inline in the code, `y -> y <= x` is very readable and one doesn't have to check that `lessThanOrEqualTo` doesn't have any surprises

src/main/java/fi/aalto/cs/apluscourses/model/ComponentInitializationCallback.java
nikke234: Could also consider moving this inside `Component` as `InitializationCallback`

src/main/java/fi/aalto/cs/apluscourses/utils/FunctionUtil.java
nikke234: Oops posted my above comment before refreshing the page so your reply wasn't visible, your plan of coming back to this later sounds good :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/ListDependenciesPolicy.java
OlliKiljunen: That's a naming decision of JetBrains as this is an overridden method. I guess they named that because the design of this class is a bit similar to the [Visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern).

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/ListDependenciesPolicy.java
OlliKiljunen: We cannot use `@NotNull` because the method is overridden.  We could add `if (value == null) return null` but I think that makes very little sense. Throwing an `ArgumentException` would be ok, but is that much more informative in stack trace than an `NullPointerException`? And then there would be a lot of redundant checks for null arguments.

Issue #121, opened 2020-04-07
setting for starting a REPL configuration window is persistent over restart


superseacat:
as a course setting

superseacat:
does not work!

superseacat:
`Exception "java.lang.ClassNotFoundException: com/intellij/codeInsight/editorActions/ReferenceData"`

superseacat:
`WARN - com.intellij.util.xmlb.Binding - no accessors for class org.jetbrains.kotlin.idea.core.script.configuration.utils.ScriptClassRootsStorage`

superseacat:
hm... [super unset](https://github.com/JetBrains/intellij-community/blob/78a91459c4d8454b2d5df36ade108807ef772373/platform/core-api/src/com/intellij/ide/util/PropertiesComponent.java#L50-L55) :confounded: 
```java
  /**
   * Set value or unset if equals to false
   */
  public final void setValue(@NotNull String name, boolean value) {
    setValue(name, value, false);
  }
```

PR #149, opened 2020-05-05
setting for starting a REPL configuration window is persistent over restart #121

superseacat:
# Description

this now should fix

# Testing

- [x] All unit tests pass
- [x] I have tested manually

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
@@ -62,26 +62,28 @@ public class PluginSettings implements MainViewModelProvider {
     return new MainViewModel();
   }
 
-  public static String shouldShowReplConfigurationDialog() {
-    return propertiesManager.getValue(A_PLUS_SHOW_REPL_CONFIGURATION_DIALOG);
+  public boolean shouldShowReplConfigurationDialog() {
+    return Boolean.parseBoolean(propertiesManager.getValue(A_PLUS_SHOW_REPL_CONFIGURATION_DIALOG));
   }
 
   /**
    * Method to set property, responsible for showing REPL configuration window.
    *
-   * @param showReplConfigDialog a {@link String} representing a boolean value of the flag.
+   * @param showReplConfigDialog a boolean value of the flag.
    */
-  public static void setShowReplConfigurationDialog(String showReplConfigDialog) {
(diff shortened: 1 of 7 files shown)


src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
```diff
@@ -62,12 +62,16 @@ private MainViewModel createNewMainViewModel(@NotNull Project project) {
...
   }
 
-  //  todo consider to create a listener
-  public static void setShowReplConfigurationDialog(boolean showReplConfigDialog) {
+  /**
+   * Method to set property, responsible for showing REPL configuration window.
+   *
+   * @param showReplConfigDialog a {@link String} representing a boolean value of the flag.
+   */
+  public static void setShowReplConfigurationDialog(String showReplConfigDialog) {
```
OlliKiljunen, line 74: Is there a reason why we stick with IntelliJ's (somewhat awkward) "boolean stored as string" design solution? Could this method just take a boolean value and hide that conversion inside itself? Especially, since the getter returns boolean, it'd be nice if they were symmetrical.

superseacat: I've totally overlooked this. Fully agree with the idea of symmetry here, this is now fixed to `String`s.

I had the opposite idea here actually. AFAIK, the current approach of the platform for `boolean` properties provides only two options: `true` and **unset** (so the value is completely erased instead of setting it to `false` if 
```java 
setValue(false)
```
 is called. This would be acceptable if, say, they've used `Boolean` or some other non-dual-value-based structure. So my point is to **explicitly** emphasize (and propose for the future) the idea of having all the properties as `String`s.

superseacat: good idea, will do this and also translate some of the points from the comments on **nikke**'s open PR.

moreover, a little concern here... IDEA seems to write many configuration files like this one`build/idea-sandbox/config/options/other.xml` only after the first (the very first) run of the IDEA has **successfully** ended. In other words, `.xml`'s get created in some `onDestroyActivity` function (haven't checked where in particular). This means, that if the initial run of the IDE (simulate with `gradle clean` -> `gradle runIde` and just kill it from the host IDE), won't terminate properly, these files won't even be created.

OlliKiljunen: Nice observation! :+1: Certainly something that could confuse the hell out of me when debugging, so it's good to be aware of it.

PR #175, opened 2020-05-18
Improved dialog

nikke234:
# Description of the PR

Replaces the current IDE settings information dialog with an improved dialog that includes the (strongly discouraged) option to opt out of course IDE settings.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [x] Yes
- [ ] Not relevant


src/main/java/fi/aalto/cs/apluscourses/ui/courseproject/CourseProjectActionDialogs.java
```diff
@@ -1,9 +1,9 @@
...
+package fi.aalto.cs.apluscourses.ui.courseproject;
 
+import com.intellij.openapi.project.Project;
 import org.jetbrains.annotations.NotNull;
 
-public interface Dialogs {
-  void showInformationDialog(@NotNull String message, @NotNull String title);
+public interface CourseProjectActionDialogs {
 
   void showErrorDialog(@NotNull String message, @NotNull String title);
```
superseacat, line 8: Nit: please consider unifying the API by making all the methods returning say `boolean`

src/main/java/fi/aalto/cs/apluscourses/ui/courseproject/CourseProjectActionDialogs.java
```diff
@@ -1,9 +1,9 @@
-package fi.aalto.cs.apluscourses.ui.base;
+package fi.aalto.cs.apluscourses.ui.courseproject;
 
+import com.intellij.openapi.project.Project;
 import org.jetbrains.annotations.NotNull;
 
-public interface Dialogs {
-  void showInformationDialog(@NotNull String message, @NotNull String title);
+public interface CourseProjectActionDialogs {
```
superseacat, line 6: as with most programming challenges, one can relatively easily find good arguments backed up [by some blogpost](https://octoperf.com/blog/2016/10/27/impl-classes-are-evil/) to support its' points. but it's not my point here :smile: 

I'd like to draw attention to a higher-level matter: in several places, we have been utilizing the approach of `Interface` + `Impl` class. this approach has certain benefits in certain cases because of (... blablabla ..., just google). thus, I'd like us to question if we are really the subject of those cases? 

as @OlliKiljunen already mentioned once, there is [YAGNI principle](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and it makes a lot of sense to me. Now the actual questions (I have no answers myself): 
+ what is the benefit of having an `Interface` here? 
+ would that make more sense to create a concrete `Class` in this case and introduce an `Interface` only when there's a clear need to use `Dialogs` elsewhere?
+ is it already known, that these `Dialogs` will be reused (have another implementation)?
+ a class just moved from `Dialogs` to become `CourseProjectActionDialogs`, could this be a sign, that this is now a conceptually same entity (`Interface` + `Imlp`)?

@nikke234 @OlliKiljunen :grey_question:

superseacat: and yes, maybe I just don't know how to cook them (`Interfaces`) properly :stuck_out_tongue_closed_eyes:

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectActionTest.java
```diff
@@ -20,6 +21,46 @@
 
 public class CourseProjectActionTest {
 
+  private class TestDialogs implements CourseProjectActionDialogs {
```
superseacat, line 24: does not it make test cases a bit dependent on this implementation? :octocat:

OlliKiljunen: Perhaps there is some confusion over the fact that, in this PR, `Dialogs` interface doesn't exist anymore but it is replaced with more specific-use interface `CourseProjectActionDialogs`. In the current design, if the old `Dialogs`-like interface are required elsewhere, there is nothing here that should be reused. This is only for `CourseProjectAction`, as the name suggests.

`CourseProjectActionDialogs`'s design is perhaps a bit sub-optimal because it's methods are in different levels of abstraction (compare `showOkCancelDialog` with `showImportIdeSettingsDialog`). One possibility would be keep old `Dialogs` interface and make `CourseProjectActionDialogs` an extension (subtype) to it. Similarly, `CourseProjectActionDialogsImpl` would inherit old `IntelliJDialogs`.

src/main/java/fi/aalto/cs/apluscourses/ui/courseproject/CourseProjectActionDialogs.java
superseacat: `false`? :roll_eyes:

superseacat: I understand it as a representation of the state of the `showErrorDialog`

checkstyle/google_checks.xml
```diff
@@ -27,6 +27,8 @@
         <property name="eachLine" value="true"/>
     </module>
 
+    <module name="SuppressWarningsFilter" />
```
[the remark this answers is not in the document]
nikke234: good point :+1:

src/main/java/fi/aalto/cs/apluscourses/ui/courseproject/CourseProjectActionDialogs.java
nikke234: Interesting blog post :+1: As Olli mentioned, the main reason for the separation here is unit-testability. I do agree that the `Impl` naming is somewhat annoying, but I think the benefits of testability outweigh the negatives here :smile: In this case I removed the original more generic `Dialogs` interface which could be used elsewhere as well, we could consider having a generic `Dialogs` interface again.

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectActionTest.java
nikke234: Yes, very good point about `showImportIdeSettingsDialog` being on a different level of abstraction than the other more generic dialog methods.

I actually really like your idea about keeping the old `Dialogs` interface and making `CourseProjectActionDialogs` a subtype of it, as the `Dialogs` interface could be useful in other places as well.

PR #176, opened 2020-05-18
Add module id to the course file

nikke234:
# Description of the PR
Adds a version ID to the `Module` class, which gets added to the course file (`a-plus-project.json`) when a module is imported. The module id is optional in the course configuration file and the modules default to an empty string id.

The current implementation uses a static lock for all instances of `APlusProject`, as I am not sure if we only have one instance of `APlusProject` per IntelliJ project. The current implementation also parses the course file again every time a new entry is added. Caching the previously parsed JSON object should be easy, as long as the `APlusProject` instance is the same.  The current locking is pretty hacky and should be improved.  

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [ ] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/model/APlusProject.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/model/APlusProject.java
@@ -6,9 +6,12 @@ import com.intellij.openapi.roots.ModuleRootManager;
 import com.intellij.openapi.roots.libraries.LibraryTable;
 import com.intellij.openapi.roots.libraries.LibraryTablesRegistrar;
 import com.intellij.util.messages.MessageBus;
+
 import fi.aalto.cs.apluscourses.model.Component;
+import fi.aalto.cs.apluscourses.model.Module;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
@@ -20,7 +23,9 @@ import java.util.Optional;
 import org.apache.commons.io.FileUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.json.JSONException;
 import org.json.JSONObject;
+import org.json.JSONTokener;
 
 public class APlusProject {
   @NotNull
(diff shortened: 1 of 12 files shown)


src/main/java/fi/aalto/cs/apluscourses/intellij/model/APlusProject.java
```diff
@@ -29,7 +42,75 @@ public Path getBasePath() {
...
+      JSONObject modulesObject = jsonObject.optJSONObject("modules");
+      if (modulesObject == null) {
+        modulesObject = new JSONObject();
+      }
+
+      JSONObject entry = new JSONObject().put("id", module.getVersionId());
+      modulesObject.put(module.getName(), entry);
+      jsonObject.put("modules", modulesObject);
+      FileUtils.writeStringToFile(courseFile, jsonObject.toString(), StandardCharsets.UTF_8);
+    }
```
OlliKiljunen, line 113: This is an interesting design choice. 🤔 And by no means I'd say it's bad. It's actually kinda cool but let's discuss it because it's something that I wouldn't come up with.

If I understand this correctly, the things that are stored in this JSON file are never kept in the main memory but instead, they are read from / written to the file each time they are accessed.

I'd find more conventional a solution where the JSON file is parsed to some domain-specific objects in startup and then those objects are modified (concurrency control happening there) and then the file is rewritten when some `save()` method is called. Did you consider this?

I can see there might be some advantages in your design but I'd really like to hear how you reasoned this out.

superseacat: hm... it seems to be the same place that attracted my attention as well (thus, in different PR)
https://github.com/Aalto-LeTech/intellij-plugin/pull/177#discussion_r427133533 :yum:

nikke234: A quick recap of what was discussed in the meeting for reference:

nikke234: A recap of what was discussed in the meeting for reference:
  + The state of the course file could be stored and maintained in `IntelliJCourse`
  + Either way, we still want to write to the file after each module import
    + In the distant future, we could consider "bulk writes" when multiple modules are imported at the same time
  + This stateless approach removes the possibility of the course file state in memory being inconsistent with the file on the disk
  + For now we'll go with this (assuming that synchronizing is sound, i.e. no deadlocks etc.) and refactor this later as needed

src/test/java/fi/aalto/cs/apluscourses/intellij/model/APlusProjectTest.java
```diff
@@ -55,18 +68,67 @@ public boolean doesDirExist(@NotNull Path relativePath) {
...
+        project.resolveComponentState(notInstalledComponent));
 
     IntelliJModelExtensions.TestComponent errorComponent =
         new IntelliJModelExtensions.TestComponent(errorComponentName, new Object());
-    assertEquals(Component.ERROR, project.resolveComponentState(errorComponent));
+    Assert.assertEquals(Component.ERROR, project.resolveComponentState(errorComponent));
+  }
+
+  @Test
+  public void testAddCourseFileEntry() throws IOException, InterruptedException {
```
superseacat, line 88: :+1:

PR #179, opened 2020-05-19
Add custom settings to workspace.xml with project settings import

nikke234:
# Description of the PR

Edits `workspace.xml` when project settings are imported. Adds a component with the name `CompilerWorkspaceConfiguration`, that contains an option with the name `AUTO_SHOW_ERRORS_IN_EDITOR` and sets its value to `false`. The custom settings are completely hard coded, and we should explore how this information could be included in the course configuration file.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [x] Yes
- [ ] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
@@ -27,6 +27,12 @@ import org.xml.sax.SAXException;
 
 public class SettingsImporterImpl implements SettingsImporter {
 
+  // Some hard coded custom settings for workspace.xml. This information should eventually come from
+  // the course configuration file in some way.
+  private static final String WORKSPACE_XML_COMPONENT_NAME = "CompilerWorkspaceConfiguration";
+  private static final String WORKSPACE_XML_OPTION_NAME = "AUTO_SHOW_ERRORS_IN_EDITOR";
+  private static final String WORKSPACE_XML_OPTION_VALUE = "false";
+
   /**
    * Downloads the course IDE settings ZIP file to a temporary file. Also adds IDEA startup actions
    * that unzip the temporary file to the IDEA configuration path after which the temporary file is
@@ -95,13 +101,7 @@ public class SettingsImporterImpl implements SettingsImporter {
 
   private static void extractZipTo(@NotNull ZipFile zipFile, @NotNull Path target)
       throws IOException {
-    List<String> fileNames = zipFile
-        .getFileHeaders()
-        .stream()
-        .filter(file -> !file.isDirectory())
-        .map(FileHeader::getFileName)
-        .collect(Collectors.toList());
-
+    List<String> fileNames = getZipFileNames(zipFile);
     for (String fileName : fileNames) {
       Path path = Paths.get(fileName);
       // The ZIP contains a .idea directory with all of the settings files. We want to extract the
@@ -112,14 +112,32 @@ public class SettingsImporterImpl implements SettingsImporter {
     }
   }
 
-  private static final String WORKSPACE_XML_COMPONENT_NAME = "CompilerWorkspaceConfiguration";
-  private static final String WORKSPACE_XML_OPTION_NAME = "AUTO_SHOW_ERRORS_IN_EDITOR";
-  private static final String WORKSPACE_XML_OPTION_VALUE = "false";
+
+  /**
+   * Returns the names of the files inside the given ZIP file. Directories are not included, but
+   * files inside directories are included.
+   * @param zipFile The ZIP file from which the file names are read.
+   * @return A list of names of the files inside the given ZIP file.
+   * @throws IOException If an IO error occurs.
+   */
+  @NotNull
+  private static List<String> getZipFileNames(@NotNull ZipFile zipFile) throws IOException {
+    return zipFile
+        .getFileHeaders()
+        .stream()
+        .filter(file -> !file.isDirectory())
+        .map(FileHeader::getFileName)
+        .collect(Collectors.toList());
+  }
 
   /**
    * Parses the XML file (usually the workspace.xml file stored in .idea) at the given path and
    * returns a {@link Document} with file contents and an additional setting added.
    * @param workspaceXmlPath The path pointing to the XML file.
+   * @throws IOException           If an IO error occurs while parsing the XML file.
+   * @throws IllegalStateException If the existing XML in the given file is malformed. In practice
+   *                               this would mean that the {@code workspace.xml} file is malformed,
+   *                               which shouldn't happen.
    */
   @NotNull
   public static Document createCustomWorkspaceXml(@NotNull Path workspaceXmlPath) throws
@@ -152,6 +170,5 @@ public class SettingsImporterImpl implements SettingsImporter {
       throw new IllegalStateException();
     }
   }
-
 
 }
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
@@ -124,14 +124,17 @@ public class SettingsImporterImpl implements SettingsImporter {
       Node projectNode = document.getDocumentElement();
 
       // Check if a component with the given name already exists
-      Node compilerConfigurationNode = DomUtil.findChildNodeWithAttribute(projectNode,
-          "component", "name", WORKSPACE_XML_COMPONENT_NAME);
-
-      if (compilerConfigurationNode == null) {
+      List<Node> componentMatches = DomUtil.getNodesFromXPath(
+          "//component[@name=\"" + WORKSPACE_XML_COMPONENT_NAME + "\"]", projectNode);
+      Node compilerConfigurationNode;
+      if (componentMatches.isEmpty()) {
         Element compilerConfigurationElement = document.createElement("component");
         compilerConfigurationElement.setAttribute("name", WORKSPACE_XML_COMPONENT_NAME);
         compilerConfigurationNode = compilerConfigurationElement;
         projectNode.appendChild(compilerConfigurationNode);
+      } else {
+        // There should only be one match, otherwise the workspace.xml file is in a weird state
+        compilerConfigurationNode = componentMatches.get(0);
       }
 
       Element option = document.createElement("option");
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
@@ -3,12 +3,12 @@ package fi.aalto.cs.apluscourses.intellij.model;
 import com.intellij.ide.startup.StartupActionScriptManager;
 import com.intellij.openapi.application.PathManager;
 import com.intellij.openapi.project.Project;
-import com.intellij.openapi.project.ProjectManager;
 import com.intellij.openapi.updateSettings.impl.UpdateSettings;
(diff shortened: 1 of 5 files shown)


src/main/java/fi/aalto/cs/apluscourses/utils/DomUtil.java
```diff
@@ -107,6 +115,39 @@ private DomUtil() {
...
+        continue;
+      }
+
+      Node attribute = attributes.getNamedItem(attributeName);
+      if (attribute != null && attribute.getNodeValue().equals(attributeValue)) {
+        return child;
+      }
+    }
+    return null;
+  }
```
OlliKiljunen, line 149: Could this method be replaced with a simple XPath query?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
```diff
@@ -91,11 +104,46 @@ public void importProjectSettings(@NotNull Project project, @NotNull Course cour
...
+          "component", "name", WORKSPACE_XML_COMPONENT_NAME);
+
+      if (compilerConfigurationNode == null) {
+        Element compilerConfigurationElement = document.createElement("component");
+        compilerConfigurationElement.setAttribute("name", WORKSPACE_XML_COMPONENT_NAME);
+        compilerConfigurationNode = compilerConfigurationElement;
+        projectNode.appendChild(compilerConfigurationNode);
+      }
+
+      Element option = document.createElement("option");
```
OlliKiljunen, line 137: Is it possible that this element already exists? Are there then two of them and does that matter?

src/main/java/fi/aalto/cs/apluscourses/utils/DomUtil.java
nikke234: Good idea, done :+1: Although now the method is open to XPath injections, but in this case all input comes from us (i.e. a trusted source), any opinions?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
nikke234: Nice catch, there are indeed two of them. IntelliJ does some sort of `workspace.xml` "cleanup" regularly (for an example when you save the file or restart the IDE). During that process the duplicate gets removed. Another curiosity:
```<option name="AUTO_SHOW_ERRORS_IN_EDITOR" value="true" />```
gets removed entirely during the cleanup, so I assume it also removes options that have their default value.

I have just observed this cleanup process, not found it documented or anything like that, so maybe I should instead explicitly check that the option isn't there yet.

File: src/main/java/fi/aalto/cs/apluscourses/utils/DomUtil.java (hunk 1 above)
Comment (nikke234) [reply src/main/java/fi/aalto/cs/apluscourses/utils/DomUtil.java]: After some more refactoring, it now uses a XPath query without any issues

File: src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java (hunk 2 above)
Comment (OlliKiljunen) [reply src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java:140]: We should refactor this hardcoded thing at some point (perhaps not until the autumn though), so if you are confident that your solution works with current version of IDEA, I'd say it's unnecessary to further improve this now.

File: src/main/java/fi/aalto/cs/apluscourses/utils/DomUtil.java [hunk 3]
```diff
@@ -118,9 +124,42 @@ private DomUtil() {
...
+   * @param document The document that is written to the file.
+   * @param out      The file to which the document is written.
+   * @throws IOException If an IO error occurs.
+   */
+  public static void writeDocumentToFile(@NotNull Document document, @NotNull File out)
+      throws IOException {
+    Source source = new DOMSource(document);
+    StreamResult result = new StreamResult(out);
+    try {
+      Transformer transformer = TransformerFactory.newInstance().newTransformer();
```
OlliKiljunen, line 159: Sonar points a vulnerability here (see SonarCloud report). Could we have a static field `transformerFactory` (cf. `documentBuilderFactory` in the same class) which is initialized with these two attributes
```java
transformerFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");

transformerFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_STYLESHEET, ""); 
```

nikke234: Done, good solution, thanks :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
```diff
@@ -82,6 +85,16 @@ public void importProjectSettings(@NotNull Project project, @NotNull Course cour
...
+
+    Path workspaceXmlPath = settingsPath.resolve("workspace.xml");
+    Document workspaceXml = createCustomWorkspaceXml(workspaceXmlPath);
+    DomUtil.writeDocumentToFile(workspaceXml, workspaceXmlPath.toFile());
+  }
+
+  private static void extractZipTo(@NotNull ZipFile zipFile, @NotNull Path target)
+      throws IOException {
     List<String> fileNames = zipFile
         .getFileHeaders()
```
superseacat, line 99: is there a chance `getFileHeaders()` return `null`?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
```diff
@@ -82,6 +85,16 @@ public void importProjectSettings(@NotNull Project project, @NotNull Course cour
...
+    extractZipTo(zipFile, settingsPath);
+
+    Path workspaceXmlPath = settingsPath.resolve("workspace.xml");
+    Document workspaceXml = createCustomWorkspaceXml(workspaceXmlPath);
+    DomUtil.writeDocumentToFile(workspaceXml, workspaceXmlPath.toFile());
+  }
+
+  private static void extractZipTo(@NotNull ZipFile zipFile, @NotNull Path target)
+      throws IOException {
     List<String> fileNames = zipFile
```
superseacat, line 98: can this part be a separate method?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
```diff
@@ -91,11 +104,54 @@ public void importProjectSettings(@NotNull Project project, @NotNull Course cour
...
+      // .idea/.idea/<settings_files>.
       Path pathWithoutRoot = path.subpath(1, path.getNameCount());
-      zipFile.extractFile(path.toString(), settingsPath.toString(), pathWithoutRoot.toString());
+      zipFile.extractFile(path.toString(), target.toString(), pathWithoutRoot.toString());
     }
+  }
+
+  private static final String WORKSPACE_XML_COMPONENT_NAME = "CompilerWorkspaceConfiguration";
+  private static final String WORKSPACE_XML_OPTION_NAME = "AUTO_SHOW_ERRORS_IN_EDITOR";
+  private static final String WORKSPACE_XML_OPTION_VALUE = "false";
```
superseacat, line 117: not a big deal, but I'd suggest moving the declaration to the top of the file, to follow the convention

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
```diff
@@ -91,11 +104,54 @@ public void importProjectSettings(@NotNull Project project, @NotNull Course cour
...
 
-    ProjectManager.getInstance().reloadProject(project);
+  /**
+   * Parses the XML file (usually the workspace.xml file stored in .idea) at the given path and
+   * returns a {@link Document} with file contents and an additional setting added.
+   * @param workspaceXmlPath The path pointing to the XML file.
+   */
```
superseacat, line 123: could you please write a brief note about why and what is being `thrown` "in case of :fire: "?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
```diff
@@ -91,11 +104,54 @@ public void importProjectSettings(@NotNull Project project, @NotNull Course cour
...
+      Node compilerConfigurationNode;
+      if (componentMatches.isEmpty()) {
+        Element compilerConfigurationElement = document.createElement("component");
+        compilerConfigurationElement.setAttribute("name", WORKSPACE_XML_COMPONENT_NAME);
+        compilerConfigurationNode = compilerConfigurationElement;
+        projectNode.appendChild(compilerConfigurationNode);
+      } else {
+        // There should only be one match, otherwise the workspace.xml file is in a weird state
+        compilerConfigurationNode = componentMatches.get(0);
+      }
```
superseacat, line 143: can this be, for example, a separate method?

src/test/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImplTest.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.commons.io.FileUtils;
+import org.junit.Assert;
+import org.junit.Test;
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+
+public class SettingsImporterImplTest {
+
+  @Test
+  public void testCreateCustomWorkspaceXml() throws IOException {
```
superseacat, line 18: :sunglasses: if only there could be more cases covered, as the original method seems to have quite a branching logics?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
nikke234: I will check :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
nikke234: Good idea, will do :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
nikke234: I checked, and it returns an empty list if the ZIP file doesn't exist, so this is safe

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
nikke234: Yup, will do :+1:

src/test/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImplTest.java
nikke234: Added a test case for when `worspace.xml` already contains the setting

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporterImpl.java
nikke234: In this case I think it would probably add more complexity. We add `compilerConfigurationNode` to `projectNode` if it doesn't exist yet, and I think it's easier to follow the logic this way

Issue #171, opened 2020-05-18
IDE settings installation bug


superseacat:
after IDE settings installation, the zip file with settings is created into `tmp`. if the file is deleted before the restart, it breaks the IDE. so there must be some mechanism (a check on start-up?) to ensure the file exists, otherwise do nothing/something. might have or not a notification (like "unsuccessful installation" or so)

PR #189, opened 2020-05-25
Ide settings installation bug

OlliKiljunen:
# Description of the PR

Hugely disappointed, **I suggest that we close the issue #171 as "WON'T FIX".**

If you see the commit log of this PR, you'll see how I tried to solve this bug. Can you see also why it didn't work. For me, it took quite a lot of time until I realized that the startup script is executed before the plugin code are loaded into runtime. That's it. If our code doesn't exist when the startup actions are executed, there is no way we can affect what happens then.

I think there are good reasons why plugins are loaded only after the startup script has been run. So probably JetBrains is not gonna change that. What we could do instead, is try to create a PR to IntelliJ's code base with `ActionCommandDecorator` and `SilentActionCommandDecorator` classes (see the commit log of this PR) and wish they could get through. I don't know though how long it would take until we could see those changes in the production.

**This PR now contains only minor code cleaning.**

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


superseacat:
@OlliKiljunen thanks for the effort with this... seems that we can't fix the world (yet :wink:)

PR #190, opened 2020-05-25
Regularly check modules for updates

nikke234:
Adds a background thread that parses the course configuration file again every 5 minutes and updates the main view model. The background thread also assembles a list of updatable modules, but I'm not sure what should actually be done with the list, any ideas @superseacat @OlliKiljunen?

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [ ] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
@@ -12,9 +12,7 @@ import fi.aalto.cs.apluscourses.intellij.model.SettingsImporterImpl;
 import fi.aalto.cs.apluscourses.intellij.services.MainViewModelProvider;
 import fi.aalto.cs.apluscourses.intellij.services.PluginSettings;
 import fi.aalto.cs.apluscourses.model.Course;
-import fi.aalto.cs.apluscourses.model.MainViewModelUpdater;
 import fi.aalto.cs.apluscourses.model.MalformedCourseConfigurationFileException;
-import fi.aalto.cs.apluscourses.presentation.CourseViewModel;
 import fi.aalto.cs.apluscourses.ui.courseproject.CourseProjectActionDialogs;
 import fi.aalto.cs.apluscourses.ui.courseproject.CourseProjectActionDialogsImpl;
 import java.io.IOException;
@@ -28,9 +26,6 @@ import org.slf4j.LoggerFactory;
 public class CourseProjectAction extends AnAction {
 
   private static final Logger logger = LoggerFactory.getLogger(CourseProjectAction.class);
-
-  @NotNull
-  private MainViewModelProvider mainViewModelProvider;
 
   @NotNull
   private CourseFactory courseFactory;
@@ -49,13 +44,11 @@ public class CourseProjectAction extends AnAction {
   /**
    * Construct a course project action with the given main view model provider and dialogs.
    */
-  public CourseProjectAction(@NotNull MainViewModelProvider mainViewModelProvider,
-                             @NotNull CourseFactory courseFactory,
+  public CourseProjectAction(@NotNull CourseFactory courseFactory,
                              boolean createCourseFile,
                              @NotNull SettingsImporter settingsImporter,
                              @NotNull IdeRestarter ideRestarter,
                              @NotNull CourseProjectActionDialogs dialogs) {
-    this.mainViewModelProvider = mainViewModelProvider;
     this.courseFactory = courseFactory;
     this.createCourseFile = createCourseFile;
     this.settingsImporter = settingsImporter;
@@ -68,7 +61,6 @@ public class CourseProjectAction extends AnAction {
    */
   public CourseProjectAction() {
     this(
-        PluginSettings.getInstance(),
         (url, project) -> Course.fromUrl(url, new IntelliJModelFactory(project)),
         true,
         new SettingsImporterImpl(),
@@ -107,8 +99,10 @@ public class CourseProjectAction extends AnAction {
       }
     }
 
-    MainViewModelUpdater mainViewModelUpdater = new MainViewModelUpdater(project);
-    new Thread(mainViewModelUpdater::run).start();
+    if (createCourseFile) {
+      // The course file not created in testing.
+      PluginSettings.getInstance().createUpdatingMainViewModel(project);
+    }
   }
 
   @Override
@@ -154,10 +148,6 @@ public class CourseProjectAction extends AnAction {
   private Course tryInitializeCourse(@NotNull Project project, @NotNull URL courseUrl) {
     try {
       Course course = courseFactory.fromUrl(courseUrl, project);
-      mainViewModelProvider
-          .getMainViewModel(project)
-          .courseViewModel
-          .set(new CourseViewModel(course));
       return course;
     } catch (IOException e) {
       notifyNetworkError();
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
@@ -16,10 +16,7 @@ import fi.aalto.cs.apluscourses.model.MalformedCourseConfigurationFileException;
 import fi.aalto.cs.apluscourses.presentation.CourseViewModel;
 import fi.aalto.cs.apluscourses.ui.courseproject.CourseProjectActionDialogs;
 import fi.aalto.cs.apluscourses.ui.courseproject.CourseProjectActionDialogsImpl;
-import fi.aalto.cs.apluscourses.utils.CoursesClient;
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
 import java.net.MalformedURLException;
 import java.net.URL;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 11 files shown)


src/main/java/fi/aalto/cs/apluscourses/intellij/model/APlusProject.java
```diff
@@ -80,9 +87,11 @@ public URL getCourseFileUrl() throws IOException {
...
-    JSONObject jsonObject = new JSONObject(tokenizer);
-    return new URL(jsonObject.getString("url"));
+    synchronized (courseFileLock) {
+      JSONTokener tokenizer = new JSONTokener(new FileInputStream(courseFile));
+      JSONObject jsonObject = new JSONObject(tokenizer);
+      return new URL(jsonObject.getString("url"));
+    }
   }
 
   private static final Object courseFileLock = new Object();
```
superseacat, line 97: can you more it to the top?

src/main/java/fi/aalto/cs/apluscourses/model/MainViewModelUpdater.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class MainViewModelUpdater {
```
superseacat, line 24: could you please review the access modifiers for the methods inside this class, please?

src/main/java/fi/aalto/cs/apluscourses/model/MainViewModelUpdater.java
```diff
@@ -0,0 +1,172 @@
...
+    // project is still actually open.
+    try {
+      return Course.fromUrl(courseUrl, new IntelliJModelFactory(aplusProject.getProject()));
+    } catch (IOException | MalformedCourseConfigurationFileException e) {
+      return null;
+    }
+  }
+
+  @NotNull
+  Set<String> getProjectModules() throws InterruptedException {
```
superseacat, line 94: I do not thoroughly understand, but could those methods (previous and the next one as well) be located elsewhere (like in the APlusProject class)?

superseacat:
I'll merge it into my own branch, if you don't mind :yum:

src/main/java/fi/aalto/cs/apluscourses/model/MainViewModelUpdater.java
```diff
@@ -0,0 +1,172 @@
...
+    Map<String, String> localModuleIds;
+    try {
+      localModuleIds = aplusProject.getCourseFileModuleIds();
+    } catch (IOException e) {
+      return updatableModules;
+    }
+
+    for (Module module : course.getModules()) {
+      // An updatable module must be in the project and it's ID in the local
+      // course file must be different from the ID in the course configuration file.
```
superseacat, line 126: but why? @nikke234

OlliKiljunen: If the version IDs are equal in local and remote, then there's no updates, so the module is not updatable. Would it be clearer if the wording was "To be updatable, a module must be in the project an its ID [...] must be different from [...]"?

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
```diff
@@ -112,6 +106,9 @@ public void actionPerformed(@NotNull AnActionEvent e) {
         ideRestarter.restart();
       }
     }
+
+    MainViewModelUpdater mainViewModelUpdater = new MainViewModelUpdater(project);
+    new Thread(mainViewModelUpdater::run).start();
```
OlliKiljunen, line 111: Java threads are on quite low level of abstraction. Did you take a look if IntelliJ platform provides some "background worker" objects to do things on background. On the other hand, as this thread is forever-running, there's no need for thread-pooling and such. But still this makes me a bit nervous. :)

Btw, is this thread (silently) terminated, if the plugin is disabled/uninstalled?

src/main/java/fi/aalto/cs/apluscourses/intellij/activities/InitializationActivity.java
```diff
@@ -70,6 +64,8 @@ public void runActivity(@NotNull Project project) {
     }
     PluginSettings.getInstance()
         .getMainViewModel(project).courseViewModel.set(new CourseViewModel(course));
+    MainViewModelUpdater mainViewModelUpdater = new MainViewModelUpdater(project);
+    new Thread(mainViewModelUpdater::run).start();
```
OlliKiljunen, line 68: Are there two threads of this kind? 😕 Is it possible to make it so that there is only one place in the code where this updater is spawned so that is easier to understand when does that happen?

src/main/java/fi/aalto/cs/apluscourses/model/MainViewModelUpdater.java
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  @NotNull
+  URL getCourseUrl() throws InterruptedException {
+    return ReadAction.compute(() -> {
+      interruptIfProjectClosed();
+      try {
+        URL url = aplusProject.getCourseFileUrl();
+        if (url == null) {
+          throw new InterruptedException();
```
OlliKiljunen, line 71: Is this correct way to handle this? Where is this caught? How do we recover from this? (The same question applies other places where this exception is thrown.)

src/main/java/fi/aalto/cs/apluscourses/model/MainViewModelUpdater.java
nikke234: Yes that's a clearer wording :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/activities/InitializationActivity.java
nikke234: There should be only one thread, but there are two different cases:
  + The project isn't a course project, so no thread created by the initialization activity. A thread may be created later by the course project action.
  + The project is a course project, so a thread is created in the initialization activity.

I agree it would be simpler if there's only one place. I'll see if I can refactor it.

src/main/java/fi/aalto/cs/apluscourses/model/MainViewModelUpdater.java
nikke234: The exception here is perhaps confusing, but there is nothing special about `InterruptedException`. There is a single catch clause for `InterruptedException`, which is outside the infinite loop, so the thread terminates after that. The catch clause also calls `Thread.currentThread().interrupt()`, which makes it "visible" that this thread was interrupted (currently that information isn't used anywhere).

`Thread.sleep` throws an `InterruptedException`, so I decided to have all the possible interruptions (e.g. project was closed) throw an `InterruptedException` and handle it in a single catch clause.

https://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
nikke234: Well it's not strictly forever-running, since it terminates if the project associated with it gets closed. But yes, I agree that Java threads are very low level. I did find IntelliJ's `ProgressManager`, but that seemed to have a lot to do with UI related to background processes (e.g. progress bars and the like), but I'll check again if IntelliJ has some convenient abstractions for this.

I overlooked plugin uninstallation, so I'll have to test that :smile:

nikke234:
> "Update loop" recreates the whole object hierarchy (Course, Module etc. objects as well as their view model correspondents). This is not necessarily bad design but I'd say then the creation of these objects (in the first place) should also happen in this "updater" method (because this kind of update =create new and replace old).
@OlliKiljunen This is a good point, and it should actually be pretty easy to implement.

src/main/java/fi/aalto/cs/apluscourses/model/MainViewModelUpdater.java
superseacat: I've totally forgot about the IDs changing as well, sorry :bow:

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
```diff
@@ -135,25 +131,31 @@ Course getCourse(@NotNull URL courseUrl) {
...
+     * loaded course has any "active" components, and only update the course view model if it
+     * doesn't.
+     */
+    boolean hasActiveComponents = current.getComponents()
+        .stream()
+        .anyMatch(component -> {
+          int state = component.stateMonitor.get();
+          int depState = component.dependencyStateMonitor.get();
+          return state == Component.FETCHING || state == Component.LOADING
+              || depState == Component.DEP_WAITING;
```
OlliKiljunen, line 155: Nit: This could be a method, e.g. `Component::isActive`, so it could be unit tested.

PR #193, opened 2020-05-28
Improved Course Project Dialog

nikke234:
This resolves the issue related to the unintuitive behavior of the current IDE settings dialog. The two dialogs (IDE settings, restart) are replaced with one dialog, which changes its look depending on the currently imported IDE settings, and allows the user to cancel the entire action.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [ ] Not relevant


superseacat:
this is (going to be) reflected in the testing.md, right? @nikke234 @OlliKiljunen

src/main/java/fi/aalto/cs/apluscourses/ui/courseproject/CourseProjectDialog.java
```diff
@@ -0,0 +1,95 @@
...
+import javax.swing.Action;
+import javax.swing.JCheckBox;
+import javax.swing.JComponent;
+import javax.swing.JLabel;
+import javax.swing.JPanel;
+import javax.swing.SwingConstants;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class CourseProjectDialog extends DialogWrapper {
```
superseacat, line 16: Swing components are totally unit-testable, shall this be tried?

nikke234: I'll try adding some tests :+1:

nikke234:
> This is very good! +1 I wonder if some logic could have been out of `CourseProjectDialog` to further separate it from the UI but that is by no means necessary (and it can be later refactored if we want to).

@OlliKiljunen Do you mean for an example a presentation layer between the action and the actual UI? I'm pretty much a complete novice when it comes to this so any pointers are welcome :smile: In any case I agree that it would be nice to make the UI "dumber" and move some of the logic elsewhere

OlliKiljunen:
> @OlliKiljunen Do you mean for an example a presentation layer between the action and the actual UI? I'm pretty much a complete novice when it comes to this so any pointers are welcome 😄 In any case I agree that it would be nice to make the UI "dumber" and move some of the logic elsewhere

Yes, something like that. If there was a something of a view model object that had fields like `userWantsSettings`, `userWantsRestart`, `isRestartAvailable`, and `areSettingsOptOutAvailable` and the checked- and enabled-states of the checkboxes were somehow bound to the corresponding fields, that is, those properties are set according to the fields's values and when a checkbox is (un)checked, it updates the value of the field and when a field's value is changed, the checkbox is notified and refreshes itself.

If you want I can try to demonstrate this idea here but it is not needed really.

nikke234:
@OlliKiljunen I made an attempt to implement the view model object you mentioned (the name `CourseProjectViewModel` is perhaps confusing; suggestions welcome). It works mostly as you specified, except that instead of the view model telling the checkboxes to refresh, the UI simply naively updates both checkboxes with values from the view model when either of them is selected/deselected.

Feel free to suggest changes if any are needed, otherwise I'll add manual testing instructions and merge.

src/main/java/fi/aalto/cs/apluscourses/ui/courseproject/CourseProjectView.java
```diff
@@ -0,0 +1,96 @@
...
+  private ItemListener createCheckboxListener(BooleanSetter setter) {
+    return event -> {
+      if (event.getStateChange() == ItemEvent.SELECTED) {
+        setter.set(true);
+      } else if (event.getStateChange() == ItemEvent.DESELECTED) {
+        setter.set(false);
+      }
+      updateCheckboxes();
+    };
+  }
```
OlliKiljunen, line 78: Nit: this could perhaps be moved somewhere else as it more generic thing (and that's good of course). Subclass of `JCheckbox` being an easy option?

src/main/java/fi/aalto/cs/apluscourses/presentation/CourseProjectViewModel.java
```diff
@@ -0,0 +1,123 @@
...
+   * @param currentlyImportedIdeSettings The name of the course for which the IDE settings have
+   *                                     currently been imported.
+   */
+  public CourseProjectViewModel(@NotNull Course course,
+                                @Nullable String currentlyImportedIdeSettings) {
+    cancel = false;
+
+    String courseName = course.getName();
+    informationText = "<html><body>The currently opened project will be turned into a project for "
+        + "the course <b>" + courseName + "</b>.</body></html>";
```
OlliKiljunen, line 31: Not even nit, just a remark: At some point we could think about whether we want all UI text to be concentrated somewhere for easy access. At the same time we could consider if they should be maintained in Markdown format and only converted to HTML either in build time (may be too difficult) or run time (but in a reasonably cached way).

nikke234: Agreed, that would good. Markdown converted to HTML at compile time would be amazing, as long as we find a solution that doesn't cause IDEA's syntax highlighting to show errors at the places where the generated HTML get's inserted :smile: But even paying a small runtime cost of converting MD to HTML is worth it for the maintainability benefits of having all our UI text in one place in an easy format

PR #195, opened 2020-05-29
Determine module files change

superseacat:
# Description of the PR

+ I'll rebase and merge @nikke234 branch here once it's done
+ a bit unsure about the 
the correctness of the behavior implemented, but let's discuss
+ currently only notifies

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [ ] Not relevant


src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
```diff
@@ -127,7 +133,32 @@ public APlusProject getProject() {
...
   }
 
+  @Override
+  public boolean hasLocalChanges() {
+    VirtualFile virtualFile = VfsUtil.findFile(getFullPath(), true);
+    final boolean[] hasChanges = {false};
+    long downloadedAt = getDownloadedAt().toInstant().toEpochMilli();
+
+    if (virtualFile != null) {
+      VfsUtilCore.visitChildrenRecursively(virtualFile, new VirtualFileVisitor<Object>() {
```
superseacat, line 143: do you think this should be deanonymized and tested on the lower lever (`visitFile(...)`)?

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
```diff
@@ -120,12 +117,14 @@ Course getCourse(@Nullable URL courseUrl) {
       if (!projectModules.contains(module.getName())) {
         continue;
       }
-      if (!module.getVersionId().equals(localModuleIds.get(module.getName()))) {
+      if (!module.getVersionId().equals(localModuleIds.get(module.getName()))
+          // todo: a bit unsure here
+          && module.hasLocalChanges()) {
```
superseacat, line 122: is this the correct place for this check to happen? :confused:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
OlliKiljunen: A good thing about that approach would be that `hasChanges` could be then a (`volatile`?) field of the de-anonymized (that single-cell array is a bit hacky, IMO, and also data race prone if `visitChildrenRecursively` use parallelism now or in future).
 
Another thing: I appreciate the effort you have put to it to make the traversal fast-quitting if there's a single modified file. I'd, however, like if more of the logic of this method could be pulled up to `Module` class. What would you like about such a design where `Module` class implemented `hasLocalChanges` but called subclass method `getLastModifiedTimestamp` that would simply calculate the maximum of file timestamps? Taking into account how seldom this method is called, I'd consider it okay to sacrifice fast-quit for clearer design.

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
```diff
@@ -13,32 +14,43 @@
   @NotNull
   private final String versionId;
 
+  @NotNull
+  private final ZonedDateTime downloadedAt;
```
OlliKiljunen, line 18: Shouldn't this be changed when the module is downloaded. Now it is only set when the object is created.

(This is my biggest concern about this PR.)

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
superseacat: I would gladly delegate plugging this in for a person who will be constructing the whole thing (this feature) together :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
```diff
@@ -41,8 +41,10 @@ public String getName() {
   public static final String COURSE_CONFIGURATION_FILE_URL
       = "https://grader.cs.hut.fi/static/O1_2020/projects/o1_course_config.json";
 
-  // 15 minutes in milliseconds
+  //  15 minutes in milliseconds
   public static final long MAIN_VIEW_MODEL_UPDATE_INTERVAL = 15L * 60L * 1000L;
+  //  1 minute in milliseconds
+  public static final long REASONABLE_DELAY_FOR_MODULE_INSTALLATION = 60L * 1000;
```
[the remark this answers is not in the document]
nikke234: I suggest we reduce the tolerance to 30 or even 15 seconds. A lot of changes can be done in a minute and a false positive is preferable to silently overwriting changes :thinking:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
```diff
@@ -127,7 +133,32 @@ public APlusProject getProject() {
...
+        @Override
+        public boolean visitFile(@NotNull VirtualFile file) {
+          boolean proceedVisiting = !hasChanges[0] && file.getTimeStamp()
```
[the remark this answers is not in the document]
nikke234: Nevermind, `visitChildrenRecursively` seems to work differently

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
```diff
@@ -120,12 +117,14 @@ Course getCourse(@Nullable URL courseUrl) {
...
         updatableModules.add(module);
       }
     }
 
-    return updatableModules;
+    return updatableModules.stream().distinct().collect(Collectors.toList());
```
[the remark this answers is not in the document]
superseacat: [true](https://github.com/Aalto-LeTech/intellij-plugin/blob/e595090ca809326d7efb5ead069d29b58e379cb7/src/test/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdaterTest.java#L71) ... fixed now :wink:

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
```diff
@@ -13,32 +14,43 @@
...
-   * object don't cause any errors. Example of a valid JSON object:
+   * contain a version id. Additional members in the object don't cause any errors. Example of a
+   * valid JSON object:
    * <pre>
    * {
    *   "name": "My Module",
    *   "url": "https://example.com",
-   *   "id": "abc"
+   *   "id": "abc",
+   *   "downloadedAt": "2020-05-29T11:20:42.513813+03:00[Europe/Helsinki]"
```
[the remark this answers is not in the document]
superseacat: oops!

superseacat:
> Great stuff! First time I see `VfsUtilCore.visitChildrenRecursively`, looks really useful 

I'll be glad if we can reuse this, as it took a bit of experimenting to find this API endpoint :joy:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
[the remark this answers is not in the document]
superseacat: yes, the behavior is a bit tricky(not according to my expectations)

https://github.com/JetBrains/intellij-community/blob/1b16fac6ad111e08e713c110d09dd423d500aadd/platform/core-api/src/com/intellij/openapi/vfs/VirtualFileVisitor.java#L89-L98

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
superseacat: yes, and initially I had an extra `update()` method in `Module`, but then realized it's better to let this open, so the person who'll implement the actual updating logics could decide. or what did you mean :question:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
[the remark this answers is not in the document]
superseacat: this is now changed a bit...

src/test/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModuleTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    File tempDirectory = FileUtilRt.createTempDirectory("first", "", true);
+    File tempFileOne = FileUtilRt.createTempFile(tempDirectory, "test1", ".json", true);
+    File tempFileTwo = FileUtilRt.createTempFile(tempDirectory, "test2", ".json", true);
+    tempFileTwo.setLastModified(Instant.now().toEpochMilli());
+    long oneSecond = 1000L;
+    sleep(PluginSettings.REASONABLE_DELAY_FOR_MODULE_INSTALLATION + oneSecond);
+
+    long now = Instant.now().toEpochMilli();
+    tempFileOne.setLastModified(now);
```
OlliKiljunen, line 36: Sonar sees vulnerabilities where `setLastModified` is called while ignoring the return value. As a solution to this, I suggest surrounding each of these with `assertTrue(...)`. Should it happen that the file metadata cannot be updated, the test results would become meaningless, so it is reasonable to fail anyways. Well, it's not probable that would ever happen but at least Sonar would be happy.

src/test/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdaterTest.java
```diff
@@ -0,0 +1,139 @@
...
+  @Test
+  public void testAddSameModuleTwiceFails() throws MalformedURLException {
+    //  given, when, then, whatever
+    Course course = getDummyCourse();
+    /**
+     *  Check {@link MainViewModelUpdaterTest#getDummyCourse()} method.
+     **/
+    fi.aalto.cs.apluscourses.model.Module module2 = new ModelExtensions.TestModule("Module2");
+    assertThrows(UnsupportedOperationException.class, () -> course.getModules().add(module2));
+  }
```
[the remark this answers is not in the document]
nikke234: But for sure we can't have modules with duplicate names (and what weird course would have modules with the same name), because the local course file uses module names as keys, but I'm not sure if that's testable.

src/test/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModuleTest.java
```diff
@@ -0,0 +1,61 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
+import java.time.Instant;
+import org.junit.Test;
+
+public class IntelliJModuleTest extends HeavyPlatformTestCase implements TestHelper {
+
+  @Test
+  public void testHasLocalChangesReturnsTrue() throws IOException, InterruptedException {
```
nikke234, line 20: I didn't expect this to be testable :smile: very nice :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
superseacat: I would not call this hack a "fast-quit" because of [this](https://github.com/Aalto-LeTech/intellij-plugin/pull/195#discussion_r433709296), so it's more of a "do not traverse" option. 

I've now extracted the anonymous class into a regular one, would it be OK if I leave this part "as is" now? You are naturally free to tweak it into a better design (or otherwise I'd need some more explanations on how to do that myself) :pray:

OlliKiljunen: You're right, my assumption on how that class works was misguided.

And yes, this is very good now. Thank you for your refactoring effort. ❤️ I appreciate that much!

src/test/java/fi/aalto/cs/apluscourses/model/CourseTest.java
```diff
@@ -64,9 +66,22 @@ public void testGetModuleWithMissingModule() throws NoSuchComponentException {
...
   private static String modulesJson = "\"modules\":[{\"name\":\"O1Library\",\"url\":"
       + "\"https://wikipedia.org\"},{\"name\":\"GoodStuff\",\"url\":\"https://example.com\"}]";
+  private static String modulesFaultyJson = "\"modules\":[{\"name\":\"O1Library\",\"url\":"
+      + "\"https://wikipedia.org\"},{\"name\":\"GoodStuff\",\"url\":\"https://example.com\"},"
+      + "{\"name\":\"GoodStuff\",\"url\":\"https://example.com\"}]";
   private static String resourcesJson = "\"resources\":{\"abc\":\"http://example.com\","
       + "\"def\":\"http://example.org\"}";
 
+  @Test
+  public void testFromConfigurationFileWithSameModule() {
```
superseacat, line 76: this will fail even earlier @nikke234 (it seems)

src/test/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdaterTest.java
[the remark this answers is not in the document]
superseacat: this will fail even earlier @nikke234 ([it seems](https://github.com/Aalto-LeTech/intellij-plugin/blob/f32d04cb8043ee6874eece41006469714463bad8/src/test/java/fi/aalto/cs/apluscourses/model/CourseTest.java#L76))

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
superseacat: OK, so now when started debugging I understand what you meant, sorry :student: . Clearly, the initially proposed approach with `Module.downloadAt` is incorrect. I'll probably try @nikke234 's proposed approach (analogous to `id` extraction).

:question: noticed one thing here... is it intended, that the `Module`s are being recreated every time project reloads? @OlliKiljunen @nikke234

nikke234: Well... yes and no.

It's intended in the sense that it is intentional and not a bug.

It's not intended in the sense that I'm not completely happy with the fact that they are recreated every time.

If someone can come up with a design that...
  + Doesn't recreate the module objects every time
  + Doesn't have any data races
  + Updates the tool window with new modules
  + Is relatively quick to implement and doesn't require massive refactoring

...then I would 100% support switching to that design. :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/APlusProject.java
```diff
@@ -163,7 +168,11 @@ public void addCourseFileEntry(@NotNull Module module) throws IOException {
       Iterable<String> moduleNames = modulesObject::keys;
       for (String moduleName : moduleNames) {
         String moduleId = modulesObject.getJSONObject(moduleName).getString("id");
-        modules.put(moduleName, moduleId);
+        ZonedDateTime downloadedAt = ZonedDateTime
+            .parse(modulesObject.getJSONObject(moduleName).getString("downloadedAt"));
+        IntelliJModuleMetadata intelliJModuleMetadata = new IntelliJModuleMetadata(moduleId,
+            downloadedAt);
+        modules.put(moduleName, intelliJModuleMetadata);
```
nikke234, line 175: Possibility to avoid getting the module object twice:
```
JSONObject moduleObject = modulesObject.getJSONObject(moduleName);
String moduleId = moduleObject.getString("id");
ZonedDateTime downloadedAt = ZonedDateTime.parse(moduleObject.getString("downloadedAt"));
```

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModuleMetadata.java
```diff
@@ -0,0 +1,30 @@
...
+    this.moduleId = moduleId;
+  }
+
+  public ZonedDateTime getDownloadedAt() {
+    return downloadedAt;
+  }
+
+  public void setDownloadedAt(ZonedDateTime downloadedAt) {
+    this.downloadedAt = downloadedAt;
+  }
```
nikke234, line 29: Nit: I wouldn't add setters unless they're actually needed

src/main/java/fi/aalto/cs/apluscourses/model/ModuleVirtualFileVisitor.java
```diff
@@ -0,0 +1,34 @@
...
+import fi.aalto.cs.apluscourses.intellij.services.PluginSettings;
+import java.time.ZonedDateTime;
+import org.jetbrains.annotations.NotNull;
+
+public class ModuleVirtualFileVisitor extends VirtualFileVisitor<Object> {
+
+  private volatile boolean hasChanges = false;
+  private final long downloadedAt;
+
+  public boolean isHasChanges() {
```
nikke234, line 14: funny name :smile:

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
```diff
@@ -100,27 +99,33 @@ Course getCourse(@Nullable URL courseUrl) {
...
-    Set<String> projectModules = getProjectModules();
+    Set<String> projectModuleNames = getProjectModuleNames();
 
     // Minor concurrency issue: we make the list of updatable modules outside of a read action, so
     // (however unlikely) a module may be removed from the project in the meantime. This shouldn't
     // be a big issue however, as it would just lead to a update notification for a module that has
     // been removed.
 
-    Map<String, String> localModuleIds;
+    Map<String, IntelliJModuleMetadata> localModuleIds;
```
nikke234, line 109: Nit: the name should probably be `localModuleMetadata`

src/test/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdaterTest.java
```diff
@@ -0,0 +1,143 @@
...
+  @Test
+  public void testGetProjectModuleNamesWithCorrectModulesReturnsValid() {
+    //  given
+    Project project = getProject();
+    createAndAddModule(project, "module1", MODULE_TYPE_ID);
+    //  modules are identified by name, so only one is actually created
+    createAndAddModule(project, "module1", MODULE_TYPE_ID);
+    createAndAddModule(project, "module2", MODULE_TYPE_ID);
+    Module[] modules = ModuleManager.getInstance(project).getModules();
+    assertEquals("Only one module named 'module1' is created.", 3, modules.length);
```
nikke234, line 43: Isn't the text incorrect here if the length is 3?

src/test/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdaterTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+    //  when
+    Set<String> projectModules = mainViewModelUpdater.getProjectModuleNames();
+
+    //  then
+    assertEmpty("There are 0 (zero) module names listed for a closed project.", projectModules);
+  }
+
+  @Test
+  public void testAddSameModuleTwiceFails() throws MalformedURLException {
```
nikke234, line 75: I would suggest renaming this test to `testGetModulesReturnsUnmodifiableList` and move it to course tests

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModuleMetadata.java
OlliKiljunen: But it will be needed soon, right?

src/main/java/fi/aalto/cs/apluscourses/model/ModuleVirtualFileVisitor.java
superseacat: autogenerated by IJ, I'll just change it to `hasChanges()` as it reads better this way, despite overlaps with the member name :yum:

src/test/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdaterTest.java
superseacat: yes, because we're using the `HeavyPlatformTestCase` then there is a default `project` with a single default `module` available (called similar to a test method, smth like `testGetProjectModuleNamesWithCorrectModulesReturnsValid`**`X`**)

PR #202, opened 2020-06-08
A+ Token UI

nikke234:
# Description of the PR

Adds the UI for inserting the token. The look of the UI is very basic as we're not sure yet if we're going to use tokens. The user is prompted for the token as needed, so the dialog isn't actually shown anywhere yet. 

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
@@ -1,0 +1,9 @@
+package fi.aalto.cs.apluscourses.model;
+
+import org.jetbrains.annotations.NotNull;
+
+public class APlusAuthentication {
+  public APlusAuthentication(@NotNull char[] token) {
+    // TODO: use some sort of GuardedString
+  }
+}
--- a/src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
@@ -5,7 +5,7 @@ import org.jetbrains.annotations.NotNull;
 
 public class APlusAuthentication {
   @NotNull
-  private char[] token;
+  private final char[] token;
 
   /**
    * Initializes an instance with the given token. Note, that the given token array is not cleared
--- a/src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
@@ -1,17 +1,26 @@
 package fi.aalto.cs.apluscourses.model;
 
+import java.util.Arrays;
 import org.jetbrains.annotations.NotNull;
 
 public class APlusAuthentication {
   @NotNull
-  private String token;
+  private char[] token;
 
+  /**
+   * Initializes an instance with the given token. Note, that the given token array is not cleared
+   * or overwritten.
+   */
   public APlusAuthentication(@NotNull char[] token) {
-    this.token = String.copyValueOf(token);
+    this.token = token.clone();
   }
 
   @NotNull
-  public String getToken() {
+  public char[] getToken() {
     return token;
+  }
+
+  public void clear() {
+    Arrays.fill(token, '\0');
   }
 }
--- a/src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
@@ -3,7 +3,15 @@ package fi.aalto.cs.apluscourses.model;
 import org.jetbrains.annotations.NotNull;
 
 public class APlusAuthentication {
+  @NotNull
+  private String token;
+
   public APlusAuthentication(@NotNull char[] token) {
-    // TODO: use some sort of GuardedString
+    this.token = String.copyValueOf(token);
+  }
+
+  @NotNull
+  public String getToken() {
+    return token;
   }
 }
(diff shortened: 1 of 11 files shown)


src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
```diff
@@ -0,0 +1,17 @@
+package fi.aalto.cs.apluscourses.model;
+
+import org.jetbrains.annotations.NotNull;
+
+public class APlusAuthentication {
+  @NotNull
+  private String token;
+
+  public APlusAuthentication(@NotNull char[] token) {
+    this.token = String.copyValueOf(token);
```
superseacat, line 10: what would you think of storing the `token` itself as `char[]` as well ([some claim](https://www.baeldung.com/java-storing-passwords) it to be more secure) :lock:

nikke234: I thought about it, but we probably want to keep the token in memory as long as the IDE is running, so that the user doesn't have to enter it repeatedly. If the token is in memory as long as the IDE is running, then I'm not sure at what point we would actually overwrite the `char[]` token.

However, there's not much harm in storing it as `char[]`, so we can play it safe and do it anyways.

nikke234:
> question shall we have an `Action` or something to ask for a `token`?

That would probably useful, so maybe an item in the A+ menu? And then API actions check if the token has been entered, and if not, prompt the user for it

superseacat:
> > question shall we have an `Action` or something to ask for a `token`?
> 
> That would probably useful, so maybe an item in the A+ menu? And then API actions check if the token has been entered, and if not, prompt the user for it

this sounds good, will you implement it as a part of this task?

nikke234:
> > > question shall we have an `Action` or something to ask for a `token`?
> > 
> > 
> > That would probably useful, so maybe an item in the A+ menu? And then API actions check if the token has been entered, and if not, prompt the user for it
> 
> this sounds good, will you implement it as a part of this task?

Yep I'll add the action :+1:

src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
```diff
@@ -0,0 +1,26 @@
...
+   * or overwritten.
+   */
+  public APlusAuthentication(@NotNull char[] token) {
+    this.token = token.clone();
+  }
+
+  @NotNull
+  public char[] getToken() {
+    return token;
+  }
```
OlliKiljunen, line 21: Remark: Let's see later what kind of interface this class should have. It could be possible that this class wouldn't actually have a getter but instead it would write the token straight to a network stream. But let's keep this as-is for now.

src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
```diff
@@ -0,0 +1,26 @@
...
+
+  public void clear() {
+    Arrays.fill(token, '\0');
+  }
```
OlliKiljunen, line 25: Have you designed who calls this and when?

src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
```diff
@@ -0,0 +1,26 @@
+package fi.aalto.cs.apluscourses.model;
+
+import java.util.Arrays;
+import org.jetbrains.annotations.NotNull;
+
+public class APlusAuthentication {
+  @NotNull
+  private char[] token;
```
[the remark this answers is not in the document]
nikke234: Oops, true

src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
nikke234: Agreed :+1:  that would also make it easy to switch to username & password later, since this would essentially just be a black box that modifies the HTTP headers as needed

src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
nikke234: Not sure yet. As mentioned in another comment to Nikolai, we'll probably end up having the token in memory whenever the IDE is running (and the token has been loaded), so it's not easy to decide where this would get called.

Anyways the utility of this is somewhat limited, since the GC may have moved the array at some point, potentially leaving the token in plaintext somewhere else in memory, which we can't overwrite. Also, if someone is actually reading the memory, then a breach has already occurred. If we really want to play it safe, we should probably use something like [Oracle's GuardedString](https://docs.oracle.com/html/E28160_01/org/identityconnectors/common/security/GuardedString.html), which encrypts its contents using a random key, the plaintext contents are accessed using callbacks, and the arrays are overwritten always after a callback is done.

OlliKiljunen: I looked at that `GuardedString` too but I guess it may be an overkill here. After all, we have no nuclear missile launch codes here. I guess we can achieve with carefully used `char[]` token a security level in which the effort to crack the system overly exceeds the gain an attacker could possibly achieve.

Bigger risk than someone maliciously getting access to the token in some GC-abandoned memory area, is, IMO, the possibility that some of us (or someone maintaining the software) accidentally end up writing code that prints the token to some log dump or somewhere else where it doesn't belong. So I'd prefer such design that minimizes the chance for that to happen.

PR #210, opened 2020-06-09
update Travis CI build pipeline with data for new endpoint

superseacat:
# Description of the PR

I kindly invite you to discuss the approach/framework chosen

+ works in pair with [Aalto-LeTech/course-templates](https://github.com/Aalto-LeTech/course-templates) to populate the newly created A+ `groups` endpoint with data
+ [rest-assured](http://rest-assured.io/) reads nicely, check it out

* [x] cleanup before merging

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,5 +24,8 @@ deploy:
     on:
       branch: test
 
+env:
+  - CI=true
+
 notifications:
   slack: $SLACK
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ addons:
     token: $SONARCLOUD_TOKEN
 
 script:
-  - ./start_local_a+_env.sh && ./gradlew check && chmod +x start_local_a+_env.sh && sonar-scanner
+  - chmod +x start_local_a+_env.sh && ./start_local_a+_env.sh && ./gradlew check && sonar-scanner
 
 deploy:
   - provider: script
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ addons:
     token: $SONARCLOUD_TOKEN
 
 script:
-  - ./gradlew check && chmod +x start_local_a+_env.sh && ./start_local_a+_env.sh && sonar-scanner
+  - ./start_local_a+_env.sh && ./gradlew check && chmod +x start_local_a+_env.sh && sonar-scanner
 
 deploy:
   - provider: script
--- a/src/test/java/fi/aalto/cs/apluscourses/integration/ApiTest.java
+++ b/src/test/java/fi/aalto/cs/apluscourses/integration/ApiTest.java
@@ -2,18 +2,17 @@ package fi.aalto.cs.apluscourses.integration;
 
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.CoreMatchers.hasItems;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
 
 import io.restassured.http.ContentType;
 import org.apache.http.HttpStatus;
+import org.junit.ClassRule;
 import org.junit.Test;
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -12,7 +12,7 @@ addons:
     token: $SONARCLOUD_TOKEN
 
 script:
-  - ./gradlew check && chmod +x start_local_a+_env.sh && ./start_local_a+_env.sh && sonar-scanner
+  - ./start_local_a+_env.sh && ./gradlew check && chmod +x start_local_a+_env.sh && sonar-scanner
```
OlliKiljunen, line 15: Shouldn't `chmod +x start_local_a+_env.sh` come before `./start_local_a+_env.sh` ? Or is it unnecessary?

src/test/java/fi/aalto/cs/apluscourses/integration/ApiTest.java
```diff
@@ -0,0 +1,54 @@
...
+        .preemptive()
+        .basic("root", "root")
+        .when()
+        .contentType(ContentType.JSON)
+        .get("http://localhost:8000/api/v2/courses/1/groups/2/")
+        .then()
+        .assertThat()
+        .statusCode(HttpStatus.SC_OK)
+        .body("members.id", hasItems(503, 504))
+        .body("members.username", hasItems("student5", "student6"));
```
OlliKiljunen, line 56: Wow, this is cool! 🤩

src/test/java/fi/aalto/cs/apluscourses/integration/ApiTest.java
```diff
@@ -0,0 +1,58 @@
...
+    //            "student_id": "135846",
+    //            "email": "student3@example.org",
+    //            "full_name": "Ben Bond",
+    //            "is_external": false
+    //        }
+    //    ],
+    //    "timestamp": "2020-06-09T10:41:30.250085+03:00"
+    //}
+
+    given()
```
superseacat, line 45: look, guys, no need to write `//  given-when-then` as a comment anymore :stuck_out_tongue_closed_eyes:

.travis.yml
superseacat: this is now right as it should be - first the instance of A+ is started, and then we run all the tests because our "integration" test is run with the same `gradle check` (`test`) actually

superseacat:
> So, **ApiTest** is run when `./gradlew test` is called (and currently fail if A+ server is not running locally)? It would be nice to be able to run unit tests without having A+ docker set up, so could it be possible that integration tests are run in a separate Gradle task? Just wondering...

there is now `CI=true` travis environment variable, that tells Travis to run the **ApiTest**s. this means, that when you do `gradle check` locally, then integration tests (currently only **ApiTest**, but use the `EnvironmentChecker` rule, if needed) are not triggered. 

if you actually do have an A+ (API) running locally and want to test against, then comment out the `@Rule` (or do something inside it) or create a environmental value (which is a superdoubtful move IMO :smile: )

.travis.yml
superseacat: yes, now I see... and surprisingly this still works :rofl:

nikke234: Shouldn't the `chmod +x start_loca_a+_env.sh` be the first one (or removed completely)?

nikke234: Oops, I commented before refreshing the page and seeing your reply

PR #212, opened 2020-06-09
Scala REPL Fixes

nikke234:
Unfortunately no one has been able to reproduce the original exception, but this fixes some issues in the REPL behavior and as a result the exception shouldn't be possible anymore.

<hr>

**Problem 1**
The REPL isn't available in a "fresh" course project.
**Solution**
Turning a project into a course project now automatically imports `O1Library` (and Scala SDK as a dependency), after which the IntelliJ Scala plugin recognizes that the project has Scala in it and enables the REPL. The automatically imported components are configured in the course configuration file.

<hr>

**Problem 2**
The REPL silently refuses to start if no module is selected.
**Solution**
The REPL action now selects modules as follows:
  + If a file of a module is open in the editor (and has "keyboard focus"), start the REPL for the module of that file.
  + Otherwise if a module (or file of a module) is selected in the project, start the REPL for that module
  + Otherwise start a project level REPL

<hr>

**Problem 3**
Starting the REPL for a module with no Scala SDK causes a "no Scala facet" error. This happens easily, since creating a default Java project leads to the following module structure:
structure
Here the module `untitled` actually doesn't have Scala SDK as a dependency and starting the REPL for it doesn't work.
**Solution**
When selecting the module to start the REPL for, also ensure that the module has a dependency on Scala SDK, otherwise fallback to the other module selection options (or project level REPL).

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
```diff
@@ -161,6 +165,25 @@ private Course tryGetCourse(@NotNull Project project, @NotNull URL courseUrl) {
...
+        = new ComponentInstallerImpl.FactoryImpl<>(new SimpleAsyncTaskManager());
+    ComponentInstaller installer = factory.getInstallerFor(course);
+
+    Runnable callback = null;
+    if (restartWhenFinished) {
+      callback = () -> ApplicationManager.getApplication().invokeLater(() -> {
+        if (dialogs.showRestartDialog(project)) {
+          ideRestarter.restart();
+        }
+      });
```
OlliKiljunen, line 181: Nit: This is a bit scary with so many lambdas 😅 . Would it be an overkill to have a class such as
```java
public class PostponedRunnable implements Runnable {
  private final Runnable task;
  private final Executor invoker;

  /* For testing purposes */
  PostponedRunnable(Runnable task, Executor invoker) {
    this.task = task;
    this.invoker = invoker;
  }

  public PostponedRunnable(Runnable task) {
    this(task, ApplicationManager.getApplication()::invokeLater);
  }

  @Override
  public void run() {
    invoker.execute(task);
  }
}
```
and then write `callback = new PostponedRunnable(() -> { ...`

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -189,6 +198,20 @@ public String getName() {
     return Collections.unmodifiableMap(resourceUrls);
   }
 
+  /**
+   * Returns a list of components that should be installed automatically for this course.
+   *
+   * @return A list of components that should be installed automatically for this course.
+   */
+  @NotNull
+  public List<Component> getAutoInstallComponents() {
```
superseacat, line 207: "test me" :yum: ?

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -58,12 +63,14 @@ public Course(@NotNull String name,
...
-                @NotNull Map<String, URL> resourceUrls) {
+                @NotNull Map<String, URL> resourceUrls,
+                @NotNull List<String> autoInstallComponentNames) {
     this.name = name;
     this.modules = modules;
     this.requiredPlugins = requiredPlugins;
     this.resourceUrls = resourceUrls;
     this.libraries = libraries;
+    this.autoInstallComponentNames = autoInstallComponentNames;
     components = Stream.concat(modules.stream(), libraries.stream())
```
superseacat, line 74: can this be extracted into a method and have some simple tests (for + and - cases e.g.)?

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -290,6 +313,28 @@ private static String getCourseName(@NotNull JSONObject jsonObject, @NotNull Str
     return resourceUrls;
   }
 
+  @NotNull
+  private static List<String> getCourseAutoInstallComponentNames(@NotNull JSONObject jsonObject,
```
superseacat, line 317: please, consider making tests for this method with complex branching logics :mortar_board:

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/ReplAction.scala
```diff
@@ -24,23 +26,62 @@ class ReplAction extends RunConsoleAction {
...
+        val configuration = setting.getConfiguration.asInstanceOf[ScalaConsoleRunConfiguration]
+        if (setConfigurationConditionally(project, module, configuration)) {
+          RunConsoleAction.runExisting(setting, runManagerEx, project)
+        }
+      case None =>
+        val configuration = setting.getConfiguration.asInstanceOf[ScalaConsoleRunConfiguration]
+        configuration.setName("Scala REPL")
+        RunConsoleAction.runExisting(setting, runManagerEx, project)
+    }
+  }
```
superseacat, line 73: this looks.... _different_ :rofl: 
nice job, man, looks much more like the proper Scala code I've seen but unable to write myself :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
superseacat: I agree with @OlliKiljunen here, please, think of how this can be simplified (I don't have any ideas/proposal how) :confounded:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
```diff
@@ -161,6 +165,25 @@ private Course tryGetCourse(@NotNull Project project, @NotNull URL courseUrl) {
     }
   }
 
+  private void startAutoInstalls(@NotNull Project project,
```
superseacat, line 168: writing (unit) tests has a hidden agenda - testability enforces simplicity. Therefore, I would really love to see test cases covering this method's logics as a whole or piece-by-piece (it's totally fine to make it `public` to be testable)

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/ReplAction.scala
```diff
@@ -50,67 +91,57 @@ class ReplAction extends RunConsoleAction {
...
+  private def hasScalaSdkLibrary(@NotNull module: Module): Boolean = {
+    var foundScalaSdk = false
+    ModuleRootManager.getInstance(module)
+      .orderEntries()
+      .forEachLibrary((library: Library) => {
+        foundScalaSdk = foundScalaSdk || Option(library.getName).exists(_.contains("scala-sdk-"))
+        !foundScalaSdk
+      })
+    foundScalaSdk
   }
```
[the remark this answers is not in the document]
nikke234: Good idea! `any` should probably also get the predicate (yours is more like a `nonEmpty`), so perhaps

```
def any(enumerator: OrderEnumerator,
        predicate: OrderEntry => Boolean): Boolean = {

  val rootPolicy = new RootPolicy[Boolean] {
    override visitOrderEntry(orderEntry: OrderEntry, 
                             value: Boolean): Boolean =
      value || predicate(orderEntry)
  }

  enumerator.process(rootPolicy, false)

}
```

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
nikke234: I should've seen this coming :smile: But you're absolutely right, I'll add tests :+1:

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/ReplAction.scala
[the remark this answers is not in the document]
OlliKiljunen: Right, but there's already that `enumerator.satisfying(predicate)` method so if we rename my method to `nonEmpty` then `any(enumerator, predicate)` is just the same as `nonEmpty(enumerator.satisfying(predicate))`.

An issue with `RootPolicy` is that there is, I think, no way to stop it but it always traverse all. Okay, hardly an issue in our case but in theory.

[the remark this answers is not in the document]
nikke234: I do like `RootPolicy` as it provides a more fuctional approach than `forEach`, but the short circuiting is nice to have so I'll go with `nonEmpty` implemented using `forEach`

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
nikke234: I'm getting flashbacks to `FunctionUtil.lessThanOrEqualTo` :smile: But I admit, this one gets confusing to read with the multiple lambdas. I'll most likely go with that `PostponedRunnable` proposal :+1:

PR #214, opened 2020-06-10
Update modules

OlliKiljunen:
# Description of the PR

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
```diff
@@ -39,10 +40,10 @@ public String getName() {
   }
 
   public static final String COURSE_CONFIGURATION_FILE_URL
-      = "https://grader.cs.hut.fi/static/O1_2020/projects/o1_course_config.json";
+      = "http://0.0.0.0:8000/o1_course_config.json";
 
-  //  15 minutes in milliseconds
-  public static final long MAIN_VIEW_MODEL_UPDATE_INTERVAL = 15L * 60L * 1000L;
+  //  7 seconds in milliseconds
+  public static final long MAIN_VIEW_MODEL_UPDATE_INTERVAL = 7L * 1000L;
```
nikke234, line 46: Local testing leftovers? :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
```diff
@@ -73,13 +77,15 @@ public CourseProjectAction(@NotNull CourseFactory courseFactory,
...
-                             @NotNull CourseProjectActionDialogs dialogs) {
+                             @NotNull CourseProjectActionDialogs dialogs,
+                             @NotNull InstallerDialogs.Factory installerDialogsFactory) {
     this.courseFactory = courseFactory;
     this.createCourseFile = createCourseFile;
     this.settingsImporter = settingsImporter;
     this.installerFactory = installerFactory;
     this.ideRestarter = ideRestarter;
     this.dialogs = dialogs;
+    this.installerDialogsFactory = installerDialogsFactory;
```
nikke234, line 88: This isn't set in the constructor with no parameters, which causes a NPE when using this action

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
```diff
@@ -73,13 +77,15 @@ public CourseProjectAction(@NotNull CourseFactory courseFactory,
                              @NotNull SettingsImporter settingsImporter,
                              @NotNull ComponentInstaller.Factory installerFactory,
                              @NotNull PostponedRunnable ideRestarter,
-                             @NotNull CourseProjectActionDialogs dialogs) {
+                             @NotNull CourseProjectActionDialogs dialogs,
+                             @NotNull InstallerDialogs.Factory installerDialogsFactory) {
```
nikke234, line 81: A possibility to consider: there's a minor inconsistency between these two, so perhaps both could be factories or both could be interfaces

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
```diff
@@ -28,10 +27,17 @@
   @NotNull
   private APlusProject aplusProject;
 
+  @NotNull
+  private final Project project;
+
```
nikke234, line 32: With this addition it might make sense to remove the `APlusProject` instance from this class, since none of the `APlusProject` "features" are actually used here anymore

src/main/java/fi/aalto/cs/apluscourses/model/Component.java
```diff
@@ -48,8 +49,7 @@ public String getName() {
    */
   public boolean isActive() {
     int state = stateMonitor.get();
-    int depState = dependencyStateMonitor.get();
-    return state == FETCHING || state == LOADING || depState == DEP_WAITING;
+    return state == FETCHING || state == LOADING || state == UNINSTALLING;
```
nikke234, line 52: When overlapping with a main view model update, the state shows "Double-click to download" while a module is updating. This probably isn't an important issue in practice, since the update interval is long enough that collisions are unlikely, but it's something to keep in mind.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
```diff
@@ -125,7 +128,7 @@ private void extractZip(File file) throws IOException {
    * This method should only be called after extractZip has been called.
    */
   @Nullable
-  private String getIdFileContents() {
+  protected String readVersionId() {
```
nikke234, line 131: `@Override`?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
```diff
@@ -77,8 +69,10 @@ protected int resolveStateInternal() {
 
   @Override
   public void load() throws ComponentLoadException {
+    ModuleManager moduleManager = project.getModuleManager();
```
nikke234, line 72: This is actually one of those places I mentioned where I'm uncertain if a read action is needed or not

src/main/java/fi/aalto/cs/apluscourses/model/ComponentInstallerImpl.java
```diff
@@ -15,10 +15,21 @@
 
   private final ComponentSource componentSource;
   private final TaskManager<T> taskManager;
+  private final Dialogs dialogs;
 
-  public ComponentInstallerImpl(ComponentSource componentSource, TaskManager<T> taskManager) {
+  /**
+   * Constructor.
```
nikke234, line 21: :rofl:

src/main/java/fi/aalto/cs/apluscourses/model/Component.java
nikke234: Why was the dependency state check removed btw?

OlliKiljunen: > Why was the dependency state check removed btw?

I was thinking that it just tells that some other component is active. Also, `depState` doesn't matter unless `state == LOADED` so that should be checked there. But it became quite complex. Considering, how this method is used we don't have to care about `depState` here.

> When overlapping with a main view model update, the state shows "Double-click to download" while a module is updating. This probably isn't an important issue in practice, since the update interval is long enough that collisions are unlikely, but it's something to keep in mind.

Is your concern only UI related or also about the functionality? Because this makes me think if we should make `stateMonitor`s to be shared between different model object instances. Idk.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
OlliKiljunen: You mean just accessing ModuleManager because the actual thing is done inside a write action? Surely, it's possible that getting ModuleManager instace w/o read lock causes errors. In this case, it could be easily fixed by moving call to `getModuleManager()` inside the write action.

What I think is more concerning case is `resolveStateInternal()` method above because it's called multiple times, and making it read action would be unfortunate.

nikke234: Yep, I'm referring specifically to the `ModuleMaager` access here.
> You must not access the model outside a read or write action. The corresponding objects are not guaranteed to survive between several consecutive read actions. As a rule of thumb, whenever starting a read action, check if PSI/VFS/project/module is still valid.

 I guess that means to be safe the module manager access should be moved inside the write action. It would be neater if the read action was done inside `APlusProject` so that clients don't have to worry about wrapping `APlusProject` method calls inside read actions, but then we couldn't conveniently put it inside the write action here. According to the quote we should also check here that the project is still valid, but I'm not 100% sure what that would mean in practice.

It would indeed be unfortunate if `resolveStateInternal` needs a read action. One option is to fine-tune the read actions inside `APlusProject`, since some things such as checking the existence of a directory don't need a read action. 

If I understand [this](https://www.jetbrains.org/intellij/sdk/docs/basics/architectural_overview/general_threading_rules.html) correctly, the real problem is long read actions, so perhaps multiple short consecutive read actions aren't too bad, since the UI thread can get write actions in between, but we should consider this carefully.

src/main/java/fi/aalto/cs/apluscourses/model/Component.java
nikke234: Yes, the concern is purely UI related. (This goes back to the old discussion of how we're mixing concurrency control and UI here :smile:.)

If I recall correctly, if the main view model gets updated while a module is in "Waiting for dependencies..." state, then the displayed text becomes "Double-click to download" until the main view model gets updated again (15 minutes later in practice), at which point it changes to "Installed". However, I would have to double-check this again, I may remember it incorrectly...

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
OlliKiljunen: I set those accesses that are inside `fi.aalto.cs.apluscourses.intellij.model` to use always read and write actions. I didn't notice any issues but let's see.

PR #226, opened 2020-06-29
Exercise Submission Dialog

nikke234:
# Description of the PR

Adds the exercise submission dialog that gets opened when an exercise from the exercise tree is double-clicked. The dialog contains information such as files to be submitted, group selection, and number of previous submissions.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
Should this have manual testing instructions?

- [ ] Yes
- [ ] Not relevant


superseacat:
I disagree with the

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
```diff
@@ -0,0 +1,186 @@
...
+    CourseViewModel courseViewModel = mainViewModel.courseViewModel.get();
+    ExercisesTreeViewModel exercisesTreeViewModel
+        = mainViewModel.exercisesViewModel.get();
+    APlusAuthenticationViewModel authenticationViewModel
+        = mainViewModel.authenticationViewModel.get();
+
+    e.getPresentation().setEnabled(e.getProject() != null && exercisesTreeViewModel != null
+        && authenticationViewModel != null && courseViewModel != null
+        && courseViewModel.getModel() != null);
+  }
```
OlliKiljunen, line 71: Nit: This `update()` method has quite a lot logic in it plus it accesses several synchronized variables. Perhaps a better solution would be moving this logic to the main view-model (say `isSubmitAvailable()` method). To (micro-?)optimize it, the value should be stored in a (volatile) boolean field which should be updated when one of the `courseViewModel`, `exercisesViewModel`, etc. changes (in the constructor `MainViewModel`, add `courseViewModel.addValueObserver(dummy -> updateIsSubmitAvailable())` and similar). Maybe even easier solution is to make `isSubmitAvailable` yet another observable (read-only) property in the main view-model. In this case, add `isSubmitAvailable.declareDependentOn(courseViewModel)` and similar to the constructor.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
```diff
@@ -0,0 +1,186 @@
...
+  private SubmissionHistory tryGetSubmissionHistory(@NotNull Exercise exercise,
+                                                    @NotNull APlusAuthentication authentication,
+                                                    @Nullable Project project) {
+    try {
+      return SubmissionHistory.getSubmissionHistory(exercise.getId(), authentication);
+    } catch (IOException e) {
+      notifyNetworkError(e, project);
+      return null;
+    }
+  }
```
OlliKiljunen, line 151: Do we want to have some caching here and in `tryGetUserGroups` (not inside the method but somewhere)?

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/VfsUtil.java
```diff
@@ -4,15 +4,66 @@
...
+   * Recursively looks for a file in the given directory with the given name.
+   *
+   * @param directory The directory from which the search is done.
+   * @param filename  The name of the file that is searched for.
+   * @return A {@link Path} to the found file, or {@code null} if a file isn't found.
+   */
+  @CalledWithReadLock
+  @Nullable
+  public static Path findFileInDirectory(@NotNull Path directory, @NotNull String filename) {
+    VirtualFile virtualFile = com.intellij.openapi.vfs.VfsUtil.findFile(directory, true);
```
OlliKiljunen, line 28: Nit: `import com.intellij.openapi.vfs.VfsUtil` and `VfsUtilCore` ?

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/VfsUtil.java
```diff
@@ -4,15 +4,66 @@
...
+      }
+
+      if (file.getName().equals(filename)) {
+        path = Paths.get(file.getPath());
+        return false;
+      }
+
+      return true;
+    }
+  }
```
OlliKiljunen, line 65: I like a lot! :+1:

src/main/java/fi/aalto/cs/apluscourses/model/Group.java
```diff
@@ -0,0 +1,76 @@
...
+   */
+  @NotNull
+  public static Group fromJsonObject(@NotNull JSONObject jsonObject) {
+    long id = jsonObject.getLong("id");
+    JSONArray members = jsonObject.getJSONArray("members");
+    List<String> memberNames = new ArrayList<>(members.length());
+    for (int i = 0; i < members.length(); ++i) {
+      memberNames.add(members.getJSONObject(i).getString("full_name"));
+    }
+    return new Group(id, memberNames);
```
OlliKiljunen, line 40: Remark: There are quite a lot of these JSON parsing functions with very similar logics in our model classes. At some point, it would be good to do some refactoring and separate lower level parsing code from the domain object building.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
```diff
@@ -0,0 +1,186 @@
...
+
+  @Nullable
+  private SubmittableExercise tryGetSubmittableExercise(@NotNull Exercise exercise,
+                                                        @NotNull APlusAuthentication authentication,
+                                                        @Nullable Project project) {
+    try {
+      SubmittableExercise submittableExercise = SubmittableExercise.fromExerciseId(
+          exercise.getId(), authentication);
```
OlliKiljunen, line 129: Cannot we somehow reuse `exercise` object we already have?

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/VfsUtil.java
nikke234: Can't do it for `VfsUtil` since our class is named `VfsUtil` as well :smile: but yes for `VfsUtilCore`

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
nikke234: Probably yes, but I would start by caching the exercise requests (especially the request for all exercises), as those are much larger

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
nikke234: Do you mean specifically in the action here, or do you mean the class `SubmittableExercise` in general?

`SubmittableExercise` includes the names of the files to be submitted which aren't available in `Exercise`.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
nikke234: Good point! :+1: I'll think about which of the alternatives I'll go with

src/main/java/fi/aalto/cs/apluscourses/model/Group.java
nikke234: Good point! :+1: Although this parsing code is still on quite a high level of abstraction so it's not too bad for now

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
nikke234: And of course caching the number of previous submissions runs the risk of telling the student that they are making submission 5/10 when they are actually making submission 10/10 :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
```diff
@@ -0,0 +1,186 @@
...
+    APlusAuthenticationViewModel authenticationViewModel
+        = mainViewModel.authenticationViewModel.get();
+
+    e.getPresentation().setEnabled(e.getProject() != null && exercisesTreeViewModel != null
+        && authenticationViewModel != null && courseViewModel != null
+        && courseViewModel.getModel() != null);
+  }
+
+  @Override
+  public void actionPerformed(@NotNull AnActionEvent e) {
```
superseacat, line 74: there are multiple potential `null`s inside the `actionPerformed(...)` (like `authentication`, `courseViewModel.getModel() != null)` is always `true`, etc), could you please re-check IDEA's notifications for those

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
```diff
@@ -0,0 +1,186 @@
...
+  @Nullable
+  private List<Path> tryGetFilePaths(@NotNull List<String> filenames,
+                                     @NotNull Module module,
+                                     @Nullable Project project) {
+    Path modulePath = Paths.get(ModuleUtilCore.getModuleDirPath(module));
+    List<Path> filePaths = new ArrayList<>(filenames.size());
+    for (String filename : filenames) {
+      Path filePath = VfsUtil.findFileInDirectory(modulePath, filename);
+      if (filePath == null) {
+        notifier.notify(new MissingFileNotification(module.getName(), filename), project);
```
superseacat, line 175: so there is a notification for each missing file, right?
just thinking, that if there's some fault in loading the modules, will it result in dozens of notifications or... :question:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
```diff
@@ -0,0 +1,186 @@
...
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import org.jetbrains.annotations.CalledWithReadLock;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class SubmitExerciseAction extends AnAction {
```
superseacat, line 39: I believe encapsulation can be sacrificed in exchange for testability, so I suggest to make `public` all the `trySomething...` methods in this class and draw some tests for them as well (this one `tryGetFilePaths` especially) :tophat:

superseacat, line 39: can it be also `DumbAware`?

src/test/java/fi/aalto/cs/apluscourses/intellij/notifications/NetworkErrorNotificationTest.java
```diff
@@ -0,0 +1,23 @@
...
+import static org.hamcrest.Matchers.containsString;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class NetworkErrorNotificationTest {
+
+  @Test
+  public void testNetworkErrorNotification() {
+    Exception exception = new Exception("hello there");
```
superseacat, line 12: seems, like this is testing two things: title concatenation (in a way) and also the method that extracts the message from the `Exception`. if this is the intention, then a small improvement could be in externalizing the `"hello there"` into a `String`, passing it to the newly created exception, and then checking that same `String` availability from the assert. this way, if the behaviour of `exception.getMessage()` changes (half of the Java code in the world would break), but the test case will indicate as well :stuck_out_tongue_winking_eye:

src/main/java/fi/aalto/cs/apluscourses/intellij/notifications/NotSubmittableNotification.java
```diff
@@ -0,0 +1,20 @@
...
+public class NotSubmittableNotification extends Notification {
+
+  /**
+   * Construct a notification that explains that the exercise cannot be submitted from the plugin.
+   */
+  public NotSubmittableNotification() {
+    super(
+        "A+",
+        "Cannot submit exercise",
+        "This exercise can only be submitted from the A+ web interface.",
```
superseacat, line 15: a UX matter (that might be too tedious to implement), but how hard would it be to add a hyperlink to that particular "web interface" submission page into the notification (planned to make a similar proposal for the "set token" dialog as well)?

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/VfsUtil.java
```diff
@@ -4,15 +4,66 @@
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.openapi.vfs.VirtualFileVisitor;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import org.jetbrains.annotations.CalledWithReadLock;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 public class VfsUtil {
```
superseacat, line 12: missed this previously, but is there some special reason why we're shadowing the name of the JB's API classes?

src/main/java/fi/aalto/cs/apluscourses/model/ExerciseGroup.java
```diff
@@ -76,8 +76,8 @@ public static ExerciseGroup fromJsonObject(@NotNull JSONObject jsonObject) {
   @NotNull
   public static List<ExerciseGroup> getCourseExerciseGroups(
       @NotNull Course course, @NotNull APlusAuthentication authentication) throws IOException {
-    URL url = new URL(String.format("%1$s/courses/%2$s/exercises/",
-        PluginSettings.A_PLUS_API_BASE_URL, course.getId()));
+    URL url = new URL(PluginSettings.A_PLUS_API_BASE_URL + "/courses/" + course.getId()
```
superseacat, line 79: :+1:

src/main/java/fi/aalto/cs/apluscourses/model/Group.java
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Get all of the groups from the A+ API for the user corresponding to the given authentication.
+   * A group with id 0 and a single member name "Submit alone" is added to the beginning of the
+   * list.
+   *
+   * @return A list of {@link Group}s that the user is a member of in the given course.
+   * @throws IOException If an error occurs (e.g. network error).
+   */
+  @NotNull
+  public static List<Group> getGroups(@NotNull Course course,
```
superseacat, line 52: I'd propose to split this method into (at least) two: one for fetching the data, and another one from processing that data into a collection of groups. in this case it would be sensible to: 
+ make an integration test (like [shown here](https://github.com/Aalto-LeTech/intellij-plugin/blob/e35f1c5d61a69275b68deac6d2d71abcbb8f2fcf/src/test/java/fi/aalto/cs/apluscourses/integration/ApiTest.java#L18-L31))
+ test the rest of the `getGroups` methods logic - missing for now

superseacat: I would insist on **writing** those simple **integration tests**, that verify A+ API's consistency on our side, because, remember, there are no test on A+ side :disappointed:

src/main/java/fi/aalto/cs/apluscourses/model/Group.java
```diff
@@ -0,0 +1,76 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.json.JSONTokener;
+
+public class Group {
+
+  private long id;
+
+  private List<String> memberNames;
```
superseacat, line 20: xm... do you know if the names (of the `full_name` uniqueness is guaranteed on the A+ side)?

to ensure we're safe I suggest to include the `id` of those students that form a pair. eventually, it would make sense to use a bit more complex object to represent a student (maybe a class `Student` even), but atm I'd at least encourage to include the identifiers. made a `group` between our test users on the dev O1 in: 
<details>
  <summary><a href="https://plus.cs.aalto.fi/api/v2/courses/163/mygroups/">/api/v2/courses/163/mygroups/</a></summary>

```json
{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 1284,
            "url": "https://plus.cs.aalto.fi/api/v2/courses/163/mygroups/1284/",
            "members": [
                {
                    "id": 19445,
                    "url": "https://plus.cs.aalto.fi/api/v2/users/19457/",
                    "full_name": "Kuusikymmentäneljä TESTI-Opiskelija"
                },
                {
                    "id": 19446,
                    "url": "https://plus.cs.aalto.fi/api/v2/users/19458/",
                    "full_name": "Kuusikymmentäviisi TESTI-Opiskelija"
                }
            ],
            "timestamp": "2020-07-02T09:51:12.035572+03:00"
        }
    ]
}
```
</details>

src/main/java/fi/aalto/cs/apluscourses/model/SubmittableExercise.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Makes a request to the A+ API to get the details of the given exercise.
+   *
+   * @throws IOException If an IO error occurs (e.g. network error).
+   */
+  @NotNull
+  public static SubmittableExercise fromExerciseId(long exerciseId,
+                                                   @NotNull APlusAuthentication authentication)
+      throws IOException {
+    URL url = new URL(PluginSettings.A_PLUS_API_BASE_URL + "/exercises/" + exerciseId + "/");
```
superseacat, line 76: could this call to the A+ API be verified with an integration test (like described [here](https://github.com/Aalto-LeTech/intellij-plugin/pull/226/files#r448789130))

src/test/java/fi/aalto/cs/apluscourses/ui/exercise/SubmissionDialogTest.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.Arrays;
+import java.util.List;
+import javax.swing.Action;
+import javax.swing.ComboBoxModel;
+import javax.swing.JComboBox;
+import org.jetbrains.annotations.NotNull;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class SubmissionDialogTest extends LightIdeaTestCase {
```
superseacat, line 20: :+1: "sky's the limit"

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
nikke234: I'll double-check :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
nikke234: The method returns after showing the first notification so only the first missing file gets a notification

src/main/java/fi/aalto/cs/apluscourses/intellij/notifications/NotSubmittableNotification.java
nikke234: Both are very good ideas! It shouldn't be too difficult, I'll try adding them

src/main/java/fi/aalto/cs/apluscourses/model/Group.java
nikke234: Good point! Willl do

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
superseacat, line 74: I definitely want to see test(s) for this method! :angel:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
superseacat: no excuses now :rofl: 

- https://github.com/Aalto-LeTech/intellij-plugin/pull/233/files#r451494895
- https://dzone.com/articles/mock-private-method

nikke234: Unfortunately I don't think so, since it looks for the files that need to be submitted

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/VfsUtil.java
nikke234: probably just too lazy to come up with an original name :laughing:

Issue #91, opened 2020-03-05
refactoring the code after Sprint #3


superseacat:
identify shortcoming and issues we have and **want** to fix. come out with the desired architecture.

the idea is to check that the code is:
* readable
* understandable
* uniform
* conforms to the previously agreed style and architecture (well-structured)

superseacat:
perform the IJ's static code analysis tool: https://www.jetbrains.com/help/idea/running-inspections.html#run-inspections-manually and then fix where feasible

superseacat:
https://github.com/Aalto-LeTech/intellij-plugin/pull/193/files/e777657f3f3430cdcbfa7f48c43fee6dd7280dcf..46aa6b7610b988cdca15e45b1c888c883393456b#r434051373

OlliKiljunen:
Related to #366

superseacat:
this legendary ticket must go, more detailed ones will be created!

PR #233, opened 2020-07-07
Submission results

superseacat:
# Description of the PR

+ tests
+ optimizations
+ questions

are coming...

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


src/main/java/fi/aalto/cs/apluscourses/model/SubmissionResult.java
```diff
@@ -0,0 +1,70 @@
...
+   * A result of a single submission.
+   *
+   * @param exerciseId an id of the exercise.
+   * @param submissionIds an {@link List} of submission ids.
+   * @param submissionsCount an amount of submissions made.
+   * @param pointsToPass a required amount of points to pass the exercise.
+   * @param maxPoints a possible maximum amount of points for the exercise.
+   * @param totalPoints a total amount of points achieved for the exercise.
+   */
+  public SubmissionResult(int exerciseId, List<Integer> submissionIds, int submissionsCount,
```
superseacat, line 31: in general, there is a separate API endpoint to fetch the data for a single submission: https://plus.cs.aalto.fi/api/v2/submissions/2921874/ , but at this point decided not to implement a call for it

src/main/java/fi/aalto/cs/apluscourses/model/SubmissionResult.java
```diff
@@ -0,0 +1,70 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+public class SubmissionResult {
+
+  private int exerciseId;
+  private List<Integer> submissionIds;
+  private int submissionsCount;
+  //  private int maxSubmissionsCount;
```
superseacat, line 16: waiting for [this](https://github.com/apluslms/a-plus/issues/605) to complete

src/main/java/fi/aalto/cs/apluscourses/utils/async/ScheduledTaskExecutor.java
```diff
@@ -0,0 +1,23 @@
...
+   * @param runnable a {@link Runnable} task to execute.
+   * @param startAfter an interval to start the task after.
+   * @param repeatEvery an interval to repeat the task every.
+   * @param timeUnit a time unit to apply to previous 2 (two) parameters.
+   */
+  public ScheduledTaskExecutor(Runnable runnable, long startAfter, long repeatEvery,
+      TimeUnit timeUnit) {
+    AppExecutorUtil.getAppScheduledExecutorService()
+        .scheduleWithFixedDelay(runnable, startAfter, repeatEvery, timeUnit);
+    //todo: incremental interval how-to?
```
superseacat, line 21: the schedules overall work in the similar manner, so in order to have an increasing execution interval, one will have to go a bit deeper to the `Thread` level, and atm not sure if I want to :stuck_out_tongue:

src/test/java/fi/aalto/cs/apluscourses/model/SubmissionsDashboardTest.java
```diff
@@ -0,0 +1,15 @@
+package fi.aalto.cs.apluscourses.model;
+
+import java.io.IOException;
+import org.junit.Test;
+
+public class SubmissionsDashboardTest {
+
+  @Test
+  public void testGetSubmissionsResults() throws IOException {
+    String token = "92384573df6ba9e45ac6297cc2b23a38592c90f7";
```
superseacat, line 10: oops! :fire:

src/main/java/fi/aalto/cs/apluscourses/model/SubmissionsDashboard.java
```diff
@@ -0,0 +1,67 @@
...
+  public static SubmissionsDashboard getSubmissionsDashboard(long courseId,
+      @NotNull APlusAuthentication authentication) throws IOException {
+    URL url = getSubmissionsDashboardAPIURL(courseId);
+    InputStream inputStream = CoursesClient.fetch(url, authentication::addToRequest);
+    JSONObject response = new JSONObject(new JSONTokener(inputStream));
+    return fromJsonObject(response);
+  }
+
+  // todo: to propose combining all the calls like this together to test
+  public static URL getSubmissionsDashboardAPIURL(long courseId) throws MalformedURLException {
```
superseacat, line 64: so the proposal is to somehow collect the API calls made against the A+ (this is only one naive idea how-to), please, comment @OlliKiljunen @nikke234

src/main/java/fi/aalto/cs/apluscourses/intellij/activities/InitializationActivity.java
```diff
@@ -61,6 +64,13 @@ public void runActivity(@NotNull Project project) {
     PluginSettings.getInstance().createUpdatingMainViewModel(project);
     ActionUtil.launch(RequiredPluginsCheckerAction.ACTION_ID,
         new ExtendedDataContext().withProject(project));
+
+    // todo: this validation should be somewhere in the Project (A+) class
```
superseacat, line 68: still thinking where to drop such a check... should a project (A+ project e.g.) know if the scope has been turned into an "A+ project" already... ?

src/main/java/fi/aalto/cs/apluscourses/model/SubmissionsDashboard.java
OlliKiljunen: I agree that's a very good idea. Perhaps I'd prefer to have something like `ApiClient` class (different objects with different URLs and authentication info could be used in test/production) in data-access layer, `SubmissionResultList` (or possibly just `List<SubmissionResult>`) in model layer. The name "dashboard" again, sounds being about how he submissions are _presented_ to the user, so perhaps that name would make more sense in the presentation layer.

But overall, this approach is very advisable! :+1:

src/test/java/fi/aalto/cs/apluscourses/integration/ApiTest.java
```diff
@@ -14,6 +16,24 @@
   @ClassRule
   public static final EnvironmentChecker checker = new EnvironmentChecker("CI");
 
+  @Test
+  public void getSubmissionsResultsReturns() throws MalformedURLException {
```
superseacat, line 20: there is no data unless https://github.com/Aalto-LeTech/intellij-plugin/issues/234 is ready

build.gradle
```diff
@@ -26,6 +26,8 @@ dependencies {
     testCompile group: 'junit', name: 'junit', version: '4.12'
     testCompile group: 'org.hamcrest', name: 'hamcrest', version: '2.2'
     testCompile group: 'org.mockito', name: 'mockito-core', version: '3.2.4'
+    testCompile group: 'org.powermock', name: 'powermock-module-junit4', version: '2.0.2'
+    testCompile group: 'org.powermock', name: 'powermock-api-mockito2', version: '2.0.2'
```
superseacat, line 30: :wave: to the **`PowerMockito`** :wave:

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectActionTest.java
```diff
@@ -163,6 +160,7 @@ public void testCreateCourseProject() {
 
   @Test
   public void testNotifiesUserOfCourseInitializationError() {
+    createMockObjects();
```
superseacat, line 163: this is sure ugly, but it seems, that methods annotated with `@Before` are executed before the class is somehow properly instantiated (and as I've now added the `BasePlatformTestCase` which is needed to get the embedded action executed properly (to have the context)), so the project, that is a result of `getProject()` is `null`.

I'd be happy to receive advice/help on solving this, thus secondary task :sob:

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -22,4 +22,14 @@
           courseViewModel,
           authenticationViewModel,
           ExercisesTreeViewModel::fromCourseAndAuthentication);
+
+  @NotNull
+  public ObservableProperty<APlusAuthenticationViewModel> getAuthenticationViewModel() {
```
superseacat, line 27: getting issues when trying to mock the object references (with `.` operator), so that's why getters :confused:

src/test/java/fi/aalto/cs/apluscourses/integration/ApiTest.java
```diff
@@ -14,19 +17,49 @@
...
+        .auth()
+        .preemptive()
+        .basic("zoralst1", "zoralst1")
+        .when()
+        .contentType(ContentType.JSON)
+        .get(url)
+        .then()
+        .assertThat()
+        .statusCode(HttpStatus.SC_OK)
+        .body("modules[0].id", equalTo(10))
```
superseacat, line 37: FFS, so this is completely different from the `local` id (keep in mind and also you might need [this](https://docs.travis-ci.com/user/running-build-in-debug-mode/#restarting-a-job-in-debug-mode-via-api))

build.gradle
nikke234: Oh no...

src/main/java/fi/aalto/cs/apluscourses/intellij/activities/InitializationActivity.java
nikke234: Isn't the current solution quite good, since the existence of a course file is a good way to check if it's a course project or not?

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/CourseFileManager.java
```diff
@@ -145,6 +145,10 @@ private File getCourseFile(@NotNull Project project) {
         .toFile();
   }
 
+  public boolean courseFileExists(@NotNull Project project){
+    return getCourseFile(project).exists();
+  }
```
nikke234: Perhaps make this `synchronized`? Not 100% sure, but better safe than sorry :thinking:

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
```diff
@@ -164,16 +181,40 @@ public void initializeLocalSettings() {
...
+   * Method (getter) to check the property, representing if the project has been "Turned into A+
+   * project".
+   *
+   * @param project a {@link Project} to set the setting to.
+   * @return a boolean value that represent the state of the project.
+   */
+  public boolean isAPlusProjectSetting(@NotNull Project project) {
+    return Boolean.parseBoolean(
+        PropertiesComponent.getInstance(project).getValue(A_PLUS_IS_A_PLUS_PROJECT.getName()));
+  }
```
nikke234, line 207: I understand the idea here, but could we just use the `CourseFileManager.courseFileExists` instead?

nikke234, line 207: I accidentally commented on outdated code earlier, but I quite liked the `CourseFileManager.courseFileExists` method, is there a reason to make it a local setting instead?

src/main/java/fi/aalto/cs/apluscourses/model/SubmissionResult.java
nikke234: I think it's now coming to the list API instead of the points API, I can fix this in my branch

src/main/java/fi/aalto/cs/apluscourses/model/SubmissionResult.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.util.List;
+import java.util.stream.Collectors;
+import org.jetbrains.annotations.NotNull;
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+public class SubmissionResult {
+
+  private int exerciseId;
```
nikke234, line 13: Nit: `long` has been used elsewhere for IDs, I'm not sure if it's actually needed but I did it to play it safe

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectActionTest.java
OlliKiljunen: My guess is that `BasePlatformTestCase` overrides the basic functionality of JUnit in a way that `@Before` methods are not called.

If you're willing to get rid of `BasePlatformTestCase` (something that I'd gladly welcome 😉 but opinions vary, I know) I'd give the action an `ActionLauncher` object (define that interface somewhere) as a constructor parameter, so I could mock `ActionUtil.launc(...)` method but I think the approach you've taken here is ok.

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/GetSubmissionsDashboardActionTest.java
```diff
@@ -0,0 +1,105 @@
...
+    doReturn(mockSubmissionsDashboard).when(spyGetSubmissionsDashboardAction)
+        .tryGetSubmissionsDashboard(1L, mockAuthentication, project);
+
+    //  when
+    spyGetSubmissionsDashboardAction.actionPerformed(anActionEvent);
+
+    //  then
+    verify(spyGetSubmissionsDashboardAction, times(0))
+        .tryGetSubmissionsDashboard(1L, mockAuthentication, project);
+    verify(mockCourse, times(0))
```
OlliKiljunen, line 102: Nit: Just noting that there is `never()` alias for `times(0)`. Also, `times(1)` can be omitted.

src/test/java/fi/aalto/cs/apluscourses/model/SubmissionsDashboardTest.java
```diff
@@ -0,0 +1,59 @@
...
+        + "\"submission_count\":0,\"points_to_pass\":1,\"id\":26160,\"points\":0,"
+        + "\"best_submission\":null,"
+        + "\"url\":\"https://plus.cs.aalto.fi/api/v2/exercises/26160/?format=json\","
+        + "\"max_points\":2},{\"passed\":true,\"official\":false,\"difficulty\":\"A\","
+        + "\"name\":\"|fi:TehtÃ¤vÃ¤1(FlappyBug5)|en:Assignment1(FlappyBug5)|\",\"submissions\":[],"
+        + "\"submission_count\":0,\"points_to_pass\":0,\"id\":26162,\"points\":0,"
+        + "\"best_submission\":null,"
+        + "\"url\":\"https://plus.cs.aalto.fi/api/v2/exercises/26162/?format=json\","
+        + "\"max_points\":10}]}]}";
+  }
```
OlliKiljunen, line 58: First, I really appreciate the effort of writing this. Second, could this be read from a (resource) file so it would be more readable?

superseacat: it's not hand-written, but the real condensed JSON response from the corresponding (pre-configured) A+ API. it used to be a valid JSON file-object actually, but reading it from the resource is an operation itself, and here I did not want to test that, but only the method logic.

currently, we use the following approaches to mocking JSONs:
1. creating a `JSON Object` and populating it by hand (like `jsonObject.put(...)`
2. creating a `JSON Object` based on the existing `String` representing the data

having an object as a "non-condensed" `String` will eventually take a tremendous amount of space, and I eventually do not want to read the file from the resource, what could be the alternatives then?

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectActionTest.java
superseacat: I think I've tried multiple approaches, like `@RunWith(JUnit4Runner.class)`, but no luck, so I'm not sure atm if `BasePlatformTestCase` is the reason, thus, because it uses JUnit3, it's the first candidate for the malfunction :disappointed:

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
superseacat: to be fair, I can't remember now... maybe I considered it to be a bit safer than relying on the file existence (can be deleted). but you are right, the brought `CourseFileManager.getInstance().getCourseFile().exists()` could work as well :lobster: 

or maybe [this StackOverflow issue description](https://stackoverflow.com/questions/31350336/within-intellij-ide-the-java-file-exists-always-returns-false) of the problem could have been a reason, I think I had something like that while creating tests

superseacat: I would not change it now...

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/GetSubmissionsDashboardAction.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * A wrapper method for a method that pulls the A+ API for course exercises' results.
+   *
+   * @param courseId       an id of the course results data to pull.
+   * @param authentication an {@link APlusAuthentication} object to use with API.
+   * @param project        a current {@link Project} to notify to.
+   * @return a fully-inflated {@link SubmissionsDashboard} or null if the API call was unsuccessful.
+   */
+  @Nullable
+  public SubmissionsDashboard tryGetSubmissionsDashboard(long courseId,
```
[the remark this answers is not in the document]
superseacat: what is the benefit of having it as a `String`?
I think I've just looked at what was used for exercise and that's why a `long`, no hard feelings here, should I change it right away or https://github.com/Aalto-LeTech/intellij-plugin/issues/91 :stuck_out_tongue_winking_eye:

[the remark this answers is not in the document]
OlliKiljunen: Someone could argue that IDs being numeric is an implementation detail of A+ system that is leaking to our code... but that would be already very far-fetched over-protectiveness. I vote for `long`s everywhere.

(And if A+ someday start using non-numeric IDs or we want to support some other LMS that doesn't have numeric IDs, we can change that. That probably won't happen until 2030s, though.)

src/main/java/fi/aalto/cs/apluscourses/intellij/activities/InitializationActivity.java
```diff
@@ -61,6 +64,12 @@ public void runActivity(@NotNull Project project) {
     PluginSettings.getInstance().createUpdatingMainViewModel(project);
     ActionUtil.launch(RequiredPluginsCheckerAction.ACTION_ID,
         new ExtendedDataContext().withProject(project));
+
+    if (PluginSettings.getInstance().isAPlusProjectSetting(project)) {
+      new ScheduledTaskExecutor(() -> ActionUtil.launch(GetSubmissionsDashboardAction.ACTION_ID,
+          new ExtendedDataContext().withProject(project)),
+          0, PluginSettings.REASONABLE_DELAY_FOR_SUBMISSION_RESULTS_UPDATE, TimeUnit.SECONDS);
+    }
```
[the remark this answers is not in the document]
superseacat: you're absolutely right :star:, a bug (fixed)!

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/GetSubmissionsDashboardAction.java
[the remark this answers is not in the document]
superseacat: I'd say using `long` for ID is quite a common practice, so I'd say it's not a leakage but more of a coincidence of best-practices of Python and Java :laughing:

src/test/java/fi/aalto/cs/apluscourses/model/SubmissionsDashboardTest.java
OlliKiljunen: Pretty JSON is roughly double the size of minified JSON, ~8KB as Java string (UTF-16). So I don't see memory as a problem here. But you're right that reading a resource file is a too complicated task to be done in a simple unit test. Some kind of helper function like `String json = testResources.getString("large.json");` could be implemented for that purpose but, honestly, this "in-code minified JSON" ain't that bad either.

src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
```diff
@@ -1,33 +1,68 @@
...
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 public class APlusAuthentication implements Authentication {
+
+  public static final String AUTHORIZATION_HEADER = "Authorization";
+  public static final String A_COURSES_PLUGIN = "A+ Courses Plugin";
+  public static final String A_API = "A+ API";
+
+  private final Object authenticationLock = new Object();
```
OlliKiljunen, line 18: I'd remove all `synchronized` blocks from this class because as far as I understand, they are no longer needed.

src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
```diff
@@ -38,8 +73,18 @@ public void clear() {
    * @return True, if strings are equal, otherwise false.
    */
   public synchronized boolean tokenEquals(@Nullable String string) {
```
OlliKiljunen, line 75: Seems like this method had double-`synchronized` already in my implementation... 😬 My bad.

Fortunately, neither of them is no longer required.

src/main/java/fi/aalto/cs/apluscourses/model/Authentication.java
```diff
@@ -3,8 +3,5 @@
 import fi.aalto.cs.apluscourses.utils.CoursesClient;
 
 public interface Authentication extends CoursesClient.HttpAuthentication {
```
OlliKiljunen, line 5: Yeah this is empty but that's ok at least for now, IMO.

On the other hand, I'd like to have `Authentication` objects in `model` as they really are part of our domain. And on the other hand, I wouldn't like `CoursesClient` to be dependent on `model` package, so this way it all works out quite okay-ishly.

src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
superseacat: relying on the safety of IJ's implementation where it comes to credential retrieval :rocket:

src/main/java/fi/aalto/cs/apluscourses/model/Authentication.java
superseacat: I had approximately the same thoughts, thanks for a cross-check-up! :bow:

PR #237, opened 2020-07-09
Actual submission

OlliKiljunen:
# Description of the PR

+

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


superseacat:
atm the target branch is `test` is this intended? @OlliKiljunen

OlliKiljunen:
@superseacat commented:
> atm the target branch is `test` is this intended? @OlliKiljunen

I changed it 👍

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
```diff
@@ -90,33 +91,61 @@ public void actionPerformed(@NotNull AnActionEvent e) {
...
+      if (!new SubmissionDialog(viewModel, project).showAndGet()) {
+        return;
+      }
+
+      try {
+        viewModel.buildSubmission().submit();
+        return;
+      } catch (IOException ex) {
+        ioException = ex;
+      }
```
superseacat, line 148: I almost like understand this, but please, consider simplifying it! :nerd_face: 

and a question, what would happen, if (L144) `viewModel.buildSubmission().submit();` keeps throwing an `IOException` :grey_question:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
```diff
@@ -162,24 +191,6 @@ private SubmissionHistory tryGetSubmissionHistory(@NotNull Exercise exercise,
     }
   }
 
-  @CalledWithReadLock
-  @Nullable
-  private List<Path> tryGetFilePaths(@NotNull List<String> filenames,
```
superseacat, line 167: please check L76 and L86 - one call to `e.getProject()` can be "optimized" :wink:

src/test/java/fi/aalto/cs/apluscourses/model/SubmittableExerciseTest.java
```diff
@@ -1,8 +1,6 @@
 package fi.aalto.cs.apluscourses.model;
 
 import java.util.Arrays;
-import java.util.Collections;
-
 import org.json.JSONArray;
 import org.json.JSONObject;
 import org.junit.Assert;
```
superseacat, line 6: would be nice to have an integration test (either here or in APITest)

src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
```diff
@@ -147,6 +154,50 @@ public static void fetchAndConsume(@NotNull URL url,
...
+  /**
+   * Sends a POST request to the given URL.
+   *
+   * @param url            A URL.
+   * @param authentication The method of authentication.
+   * @param data           Map of request data.  Values can be strings, numbers or files.
+   *
+   * @throws IOException In case of I/O related errors or non-successful response.
+   */
+  public static void post(@NotNull URL url,
```
superseacat, line 166: "I strongly disagree with the" 
@OlliKiljunen @nikke234 

so my understanding of this class is: responsible for all the communication outside the plugin (course config & A+), and yet, the bunch of methods inside is not tested... my bad, I've overlooked this, but atm, my proposal is: either all the stuff inside is tested OR we research some ready-made third-party solutions that can at least partially cover the functionality. but after all, this **must** be tested (good question is by whom and when :smile: )

src/main/java/fi/aalto/cs/apluscourses/utils/LongIdObject.java
```diff
@@ -0,0 +1,28 @@
+package fi.aalto.cs.apluscourses.utils;
+
+/**
+ * A class whose objects' equivalence is determined based on a long-valued ID field.
+ */
+public class LongIdObject {
```
superseacat, line 6: could you please elaborate a bit on the purpose of this class, but first check if [this](https://docs.oracle.com/javase/8/docs/api/java/lang/Long.html) could have a use here?

OlliKiljunen: We have some model classes (`Exercise`, `Group`, else?) whose `equals(Object)` method should(?) be like (an example for `Group`)
```java
@Override
public boolean equals(Object obj) {
  return obj instanceof Group && ((Group) obj).getId() == getId();
}
```
(I realized I should override `equals(Object)` when I was writing tests and wanted to assert if a `Group` object returned by some method was equal to one I had used in the test. If `equal` is not overridden, reference equality is applied, but I was unsure if the object instance returned by that method was the same or not, in that case, it was unimportant implementation detail whether the same instance is used or not as long as they represent the same group.) 

I wrote this class to avoid writing similar `equals(Object)` and `hashcode()` methods to multiple classes. So the purpose of this class is/was to reduce code duplication.

However, now when I think it, there is a caveat in this solution because `a.equals(b)` (where `a` and `b` are e.g. `Group`s) only returns true if they have the exact same runtime type (because of `obj.getClass() == getClass()` clause in `equals(Object obj)`). On the other hand, that is needed because we don't want a group to equal to an exercise having the same `id` but, on the other hand, consider for example:
```java
Group group = new Group(7, Collections.emptyList());
Group sameGroup = Mockito.spy(group);
Assert.assertEquals(group, sameGroup);
```
which fails because the two objects have different runtime types. So, the conventional way to implement `equals(Object)` and `hashcode()` in `Group` class (as well as other such classes like `Exercise`, separately) is perhaps the way to go.

I'll remove this class.

src/main/java/fi/aalto/cs/apluscourses/model/Authentication.java
```diff
@@ -0,0 +1,15 @@
...
+public interface Authentication extends CoursesClient.HttpAuthentication {
+
+  void clear();
+
+  int maxTokenLength();
+
+  void setToken(@NotNull char[] newToken);
+
+  boolean isSet();
+}
```
nikke234, line 15: Is there a benefit in having this interface instead of `APlusAuthentication` directly extending `CoursesClient.HttpAuthentication`?

src/main/java/fi/aalto/cs/apluscourses/model/Submission.java
```diff
@@ -0,0 +1,55 @@
...
+   *
+   * @param exercise       Exercise.
+   * @param submissionInfo Information for the submission.
+   * @param files          Map from keys to file paths.
+   * @param group          Group in which the submission is made.
+   */
+  public Submission(@NotNull Exercise exercise,
+                    @NotNull SubmissionInfo submissionInfo,
+                    @NotNull Map<String, Path> files,
+                    @NotNull Group group) {
```
nikke234, line 29: Nit: the comment here could mention what the keys in the map are

src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
```diff
@@ -1,27 +1,57 @@
...
 
-  /**
-   * Initializes an instance with the given token. Note, that the given token array is not cleared
-   * or overwritten.
-   */
-  public APlusAuthentication(@NotNull char[] token) {
-    this.token = token.clone();
+  public APlusAuthentication(int maxLength) {
+    this.token = new char[maxLength];
+    this.maxLength = maxLength;
```
nikke234, line 15: I'm not sure if this single-array-instance limitation actually brings any benefit in this case (while imposing an arbitrary length limit)? Then again if the tokens are always < 40 it's no big deal either way

src/main/java/fi/aalto/cs/apluscourses/model/Authentication.java
OlliKiljunen: Not really, I guess (except what is the usual benefit of having an interface). Actually, we need an interface that has `addToRequest()` and `isSet()` methods but which can be either token-authentication or password-authentication. So I'll change this to be that interface and let `APlusAuthentication` stay as a class for token authentication.

src/main/java/fi/aalto/cs/apluscourses/model/Submission.java
OlliKiljunen: Right. Now it's just stupid. 😅

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
```diff
@@ -41,146 +49,135 @@
...
+    exerciseDataSource.submit(submission.buildSubmission());
   }
 
   private void notifyNetworkError(@NotNull IOException exception, @Nullable Project project) {
     notifier.notify(new NetworkErrorNotification(exception), project);
   }
+
+  public interface ModuleSource {
+    Module[] getModules(Project project);
+  }
```
nikke234, line 182: Is this interface for testing purposes? If `SubmitExerciseActionTest` extends `LightIdeaTestCase`, then the standard way of getting modules should work and this shouldn't be needed

src/main/java/fi/aalto/cs/apluscourses/utils/observable/ObservableProperty.java
```diff
@@ -38,8 +49,12 @@
     callback.valueChanged(observer, get());
   }
 
+  public <O> void addValueObserver(@NotNull O observer, @NotNull SimpleCallback<O> callback) {
+    addValueObserver(observer, new CallbackWrapper<>(callback));
+  }
+
   public void declareDependentOn(@NotNull ObservableProperty<?> property) {
-    property.addValueObserver(this, (self, dummy) -> self.onValueChanged(self.get()));
+    property.addValueObserver(this, ObservableProperty<T>::onValueChangedInternal);
```
nikke234, line 57: I liked the lambda more :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
OlliKiljunen: On the other hand, we don't need `LightIdeaTestCase` if we have this. But if you dislike it, we can get rid of it, I think. Yes, it's for testing purposes.

src/main/java/fi/aalto/cs/apluscourses/utils/observable/ObservableProperty.java
```diff
@@ -67,4 +90,29 @@ default void valueChangedUntyped(@NotNull Object observer, @Nullable T value) {
...
+    }
+
+    @Override
+    public void valueChanged(@NotNull O observer, @Nullable T value) {
+      callback.valueChanged(observer);
+    }
+  }
+
+  @FunctionalInterface
+  public static interface Validator<T> {
```
nikke234, line 114: "Modifier 'static' is redundant for inner interfaces"

src/main/java/fi/aalto/cs/apluscourses/utils/FactorySelector.java
```diff
@@ -0,0 +1,36 @@
...
+    return factory == null
+        ? createInternal(klass.getSuperclass(), obj, arg)
+        : factory.create(obj, arg);
+  }
+
+  public interface Factory<T, A, R> {
+    @NotNull
+    R create(@NotNull T model, @Nullable A arg);
+  }
+}
```
nikke234, line 36: This class adds a lot of complexity compared to the benefits it brings (as far as I can see). Personally, I would really prefer a simpler approach even if it means some code duplication etc. :smile:

src/main/java/fi/aalto/cs/apluscourses/utils/observable/ObservableProperty.java
OlliKiljunen: Ah yes, this was because I though I'd need `onValueChangedInternal` for something else too but ended up removing that use of it so it only was referenced here. I agree, it is unnecessary for this single purpose, so I'll reintroduce that lambda.

src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
nikke234: The ready-made third-party solution here is Apache's HTTP client :smile: I agree that tests would be nice, although I don't personally think it's crucial they come in this PR

src/main/java/fi/aalto/cs/apluscourses/utils/observable/ObservableProperty.java
OlliKiljunen: Although that lambda is a bit tricky because one have to be careful not to write e.g.
```java
(self, dummy) -> self.onValueChanged(this.get())
```
but we are careful, right?

src/main/java/fi/aalto/cs/apluscourses/utils/observable/ObservableProperty.java
OlliKiljunen: Apparently, yes. 👍

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
nikke234: Hmm I guess my brain was still on my branch where `SubmitExerciseAction` had like 8 constructor params with all the factories. Now the situation isn't that bad anymore, so actually having this interface is not a problem for me

src/main/java/fi/aalto/cs/apluscourses/model/Authentication.java
nikke234: Sounds good, then I guess `HttpAuthentication` can be removed if this will have `addToRequest`

src/main/java/fi/aalto/cs/apluscourses/utils/observable/ObservableProperty.java
nikke234: Good point :smile: I'm sure once it's there, no one will change it without being careful.... hopefully..

src/main/java/fi/aalto/cs/apluscourses/ui/base/OurDialogWrapper.java
```diff
@@ -0,0 +1,45 @@
...
+      for (ValidationItem validationItem : validationItems) {
+        ValidationError validationError = validationItem.validate();
+        if (validationError != null) {
+          return new ValidationInfo(validationError.getDescription(),
+              validationItem.getComponent());
+        }
+      }
+      return null;
+    }
+  }
```
nikke234, line 44: Does this add any functionality to the `DialogWrapper` validation functionality, or is it an alternative interface to the same thing?

src/main/java/fi/aalto/cs/apluscourses/utils/FactorySelector.java
OlliKiljunen: Hmm... I was just thinking how easily it would solve an issue I was facing (well, not really facing but more like thinking of) with `Authentication`.

So, for integration tests (and in the future in production, too), we need to use different authentication method. The most conventional way to implement that is having two subclasses of `Authentication`: `TokenAuthentication` and `PasswordAuthentication`. When we'll support password authentication also in the UI, we'll have `PasswordAuthenticationViewModel` (`TokenAuthenticationViewModel` we already have, though it's differently named). We would also have separate views (dialogs) for these.

In this case, we could just set up
```java
Dialogs.register(TokenAuthenticationViewModel.class, TokenAuthenticationDialog::new);
Dialogs.register(PasswordAuthenticationViewModel.class, PasswordAuthenticationDialog::new);
```
and having an `AuthenticationViewModel` (not knowing is it token or password), we could call
```java
Dialogs.create(authenticationViewModel).showAndGet()
```
to show the correct kind of dialog to the user and consider authentication configured.

So I like this class but I think we can drop it if you want to.

src/main/java/fi/aalto/cs/apluscourses/ui/base/OurDialogWrapper.java
OlliKiljunen: It is an alternative interface to better support validation on presentation level where it should be done (separating logic and UI).

src/main/java/fi/aalto/cs/apluscourses/utils/FactorySelector.java
nikke234: That makes sense, but when we are showing the authentication dialog, we don't (I think) really care what the current authentication method is, instead we care how the use wants to authenticate _now_. Maybe the different authentication methods have different menu items or there is some other way for the user to choose, but when the authentication dialog is shown, we should already know which one we are showing.

Then again showing the same type of dialog as the existing one could be useful if the authentication is incorrect and we want to prompt the user to re-enter it so it could be useful now that I think about it

nikke234: Although even if we prompt the user to re-authenticate, it may make sense to let them select between token/user+pw

OlliKiljunen: You're correct. Well, the authentication was more like an example for a case where this class could be useful. Selector mechanisms that choose a view based on view model type are, however, typical for UI frameworks following the presentation model (MVVM), so this is somewhat of a common pattern.

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
```diff
@@ -0,0 +1,105 @@
...
+  protected class TreeMouseListener extends MouseAdapter {
+
+    @Override
+    public void mouseClicked(@NotNull MouseEvent e) {
+      if (e.getClickCount() == 2 && selectedItem != null) {
+        ActionEvent actionEvent = new ActionEvent(this, ActionEvent.ACTION_PERFORMED, null);
+        nodeAppliedListeners.forEach(listener -> listener.actionPerformed(actionEvent));
+      }
+    }
+  }
```
nikke234, line 104: I forgot to add this myself, but an enter key listener (toggle expansion if non-leaf none, call node listeners if leaf node) would be a nice-to-have addition

src/main/java/fi/aalto/cs/apluscourses/presentation/base/TreeViewModel.java
```diff
@@ -1,7 +1,9 @@
 package fi.aalto.cs.apluscourses.presentation.base;
 
-import javax.swing.tree.TreeModel;
+import java.util.List;
+import org.jetbrains.annotations.Nullable;
 
 public interface TreeViewModel {
-  TreeModel toTreeModel();
+  @Nullable
+  List<? extends TreeViewModel> getSubtrees();
```
nikke234, line 8: This is a very nice solution and also takes the `DefaultMutableTreeNode` stuff out of the view models :+1: To nit: perhaps `getChildren` is a more descriptive name?

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExercisesTreeViewModel.java
```diff
@@ -1,100 +1,49 @@
...
-      if (selected.isPresent()) {
-        return selected.get();
-      }
-    }
-    return null;
+    return getGroupViewModels().stream()
+        .flatMap(group -> group.getExerciseViewModels().stream())
+        .filter(ExerciseViewModel::isSelected)
+        .findFirst()
+        .orElse(null);
```
nikke234, line 37: Nice! :sunglasses:

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/SubmissionViewModel.java
```diff
@@ -1,54 +1,55 @@
...
 
-  private Project project;
+  public final ObservableProperty<Group> selectedGroup =
+      new ObservableReadWriteProperty<>(null, SubmissionViewModel::validateGroupSelection);
 
-  private Module selectedModule;
+  @Nullable
+  private static ValidationError validateGroupSelection(@Nullable Group group) {
+    return group == null ? new GroupNotSelectedError() : null;
+  }
```
nikke234, line 38: lambdaphobia? :laughing:

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/SubmissionViewModel.java
```diff
@@ -62,53 +63,53 @@ public String getPresentableExerciseName() {
...
+  public Submission buildSubmission() {
+    Group group = Objects.requireNonNull(selectedGroup.get());
+    return new Submission(exercise, submissionInfo, files, group);
   }
 
-  public int getMaxNumberOfSubmissions() {
-    return exercise.getSubmissionsLimit();
+  public ValidationError validateSubmissionCount() {
+    return getCurrentSubmissionNumber() > getMaxNumberOfSubmissions()
+        ? new MaxNumberOfSubmissionsExceededError() : null;
```
nikke234, line 85: Does this prevent submissions or is it just a warning? If it prevents submission then we need to be sure that our max submissions takes any deviations into account

OlliKiljunen: I change it to a warning. I was already going to but thanks for reminding! :+1:

src/test/java/fi/aalto/cs/apluscourses/model/GroupTest.java
```diff
@@ -7,6 +7,7 @@
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.junit.Test;
+import org.mockito.Mockito;
```
nikke234, line 10: unused import :wink:

src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
nikke234: If the concern is that the old array remains somewhere, then it could always be cleared when a new array is allocated

src/test/java/fi/aalto/cs/apluscourses/ui/CheckBoxTest.java
```diff
@@ -3,6 +3,7 @@
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 
+import fi.aalto.cs.apluscourses.ui.base.CheckBox;
```
nikke234, line 6: I think the test should be in the `base` package instead of importing `CheckBox` here

src/main/java/fi/aalto/cs/apluscourses/utils/FactorySelector.java
nikke234: Hmm, perhaps this fits into the "I don't know how it's implemented but I can still use it" -category that you mentioned on Friday, but for that to actually be true, it could at least use some javadocs.

As far as common MVVM patterns goes, I have no idea and defer to your knowledge there :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/services/Dialogs.java
```diff
@@ -0,0 +1,22 @@
+package fi.aalto.cs.apluscourses.intellij.services;
+
```
nikke234, line 2: We should perhaps reconsider the naming of this package, since neither `Dialogs` nor `PluginSettings` are services in the IntelliJ SDK sense

src/main/java/fi/aalto/cs/apluscourses/ui/APlusAuthenticationView.java
```diff
@@ -53,10 +55,16 @@ protected JComponent createCenterPanel() {
...
-      return null;
+    int length = inputField.getPassword().length;
+    int maxLength = authenticationViewModel.getMaxLength();
+    if (length == 0) {
+      return new ValidationInfo("Token must not be empty", inputField);
     }
-    return new ValidationInfo("Token must not be empty", inputField);
+    if (length > maxLength) {
+      return new ValidationInfo(String.format("Token must not be longer than %d characters",
+          maxLength), inputField);
```
nikke234, line 65: Nit: `else if`?

src/test/java/fi/aalto/cs/apluscourses/ui/CheckBoxTest.java
OlliKiljunen: You're correct, of course.

src/main/java/fi/aalto/cs/apluscourses/ui/APlusAuthenticationView.java
OlliKiljunen: I like it more without `else`. Or if we had `else if`, I'd put the final `return null` statement also in an `else` branch. Or what about
```java
return length == 0 ? new ValidationInfo(...)
  : length > maxLength ? new ValidationInfo(...)
  : null;
```

nikke234: Well, compared to the ternaries, rather keep it as is :smile:

OlliKiljunen: I think mutual exclusiveness of conditions doesn't matter here because each `if` branch ends with `return` statement. I'm pretty sure, in this case, all three options (without `else`, with `else`, and `? :`) compile to the exactly same bytecode, so the question is only about clarity of code.

nikke234: I was also referring to the clarify of the code, not actual bytecode, but I also completely forgot that the first if-branch has a `return` in it, which renders my point moot, my bad

src/main/java/fi/aalto/cs/apluscourses/utils/ArrayUtil.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Maps array to array.
+   *
+   * @param source    The source array.
+   * @param func      Mapping function that can throw an exception.
+   * @param generator Use {@code R[]::new}.
+   * @param <T>       Type of source elements.
+   * @param <R>       Type of result elements.
+   * @param <E>       Type of exception.
```
superseacat, line 20: the whole thing **is** about this `exception`, right?

OlliKiljunen: Not sure what you mean, but for functions that don't throw anything, we could also use
```java
Arrays.stream(source).map(func).toArray(generator)
```

We have quite a few of these kind of array mappings where array is populated in a `for` loop. Considering how `for` loops are a common source of bugs (especially off-by-1 errors) I think we should slowly start abstracting them from the business logic code.

src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
OlliKiljunen: Now when I thought about it, I see you were right. I reverted to use "create new authentication object for new token" implementation we had. It's certainly better.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
superseacat: I'm glad this is being discussed. @OlliKiljunen @nikke234 

The IJ Platform has a certain approach to testing (and here I mean the `Light/HeavyIdeaTestCase`s), so overall, I see little to none reason to introduce any additional code to ~~reinvent the wheel~~ overcome the challenges we face. 

In simple words, and this is my deeper position on this: any code that we write ourselves is a maintenance burden in the future, whereas any **3-rd party code** you reuse is someone else's problem (in terms of testing/maintenance etc.), so **I prefer** latter.  I'm aware there are additional factors in play here, like "control" and "flexibility", however, with the current resourcing, we are clearly not capable of producing the high level of code test coverage.

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseActionTest.java
```diff
@@ -0,0 +1,279 @@
...
+    moduleName = "MyModule";
+    modulePath = Paths.get(moduleName);
+    moduleFilePath = modulePath.resolve("MyModule.iml").toString();
+    module = mock(Module.class);
+    doReturn(moduleFilePath).when(module).getModuleFilePath();
+    doReturn(moduleName).when(module).getName();
+
+    project = mock(Project.class);
+
+    filePath = modulePath.resolve(fileName);
```
superseacat, line 134: this is like the example of what I mean here [in the earlier comment](https://github.com/Aalto-LeTech/intellij-plugin/pull/237/files#r457858232) (can be mocked)

also, taking into account the recent features, it might actually make sense to create own TestHelper classes, that in their turn extend IJ's `Light/HeavyPlatformTests` and provide the functionality to work with our own logical entities like `Submission`, `Exercise`, `Module`, etc.

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseActionTest.java
```diff
@@ -0,0 +1,279 @@
...
+    verifyNoInteractions(submissionDialogHelper);
+    verify(exerciseDataSource, never()).submit(any());
+
+    verify(notifier).notify(any(NotSubmittableNotification.class), eq(project));
+
+    verifyNoMoreInteractions(notifier);
+  }
+
+  @Test
+  public void testNotifiesOfNetworkError1() throws IOException {
```
superseacat, line 228: can we please try avoid using the numbers in the method names and writing out the difference between the test scenarios with words :wink:  (I know it's quite challenging sometimes)

src/main/java/fi/aalto/cs/apluscourses/model/APlusAuthentication.java
superseacat: this is to change for being stored in the IJ's `PasswordSafe` in a way, that there is not gonna be a token `char []` at all

src/main/java/fi/aalto/cs/apluscourses/model/Exercise.java
```diff
@@ -29,12 +29,22 @@ public static Exercise fromJsonObject(@NotNull JSONObject jsonObject) {
...
-  public String getName() {
-    return name;
+  @Override
+  public int hashCode() {
+    return Long.hashCode(id);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    return obj instanceof Exercise && ((Exercise) obj).getId() == getId();
```
superseacat, line 48: the latter one is `this.getId()`, right?

src/main/java/fi/aalto/cs/apluscourses/model/ExerciseDataSource.java
```diff
@@ -0,0 +1,25 @@
+package fi.aalto.cs.apluscourses.model;
+
+import java.io.IOException;
+import java.util.List;
+import org.jetbrains.annotations.NotNull;
+
+public interface ExerciseDataSource {
```
superseacat, line 7: so is it so, that the `ExerciseDataSource` is an **"Exercise"** and the `Exercise` is a **"ExerciseImpl"**?

src/main/java/fi/aalto/cs/apluscourses/model/Main.java
```diff
@@ -0,0 +1,39 @@
+package fi.aalto.cs.apluscourses.model;
+
+import java.util.List;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class Main {
```
superseacat, line 7: could there be maybe a JavaDoc or something that briefly explains the idea of the `Main` class here (not to mention the ambiguity of the Java `Main` class in general) :question: 

another question is related to the logical location of the class, as all the other classes in that package are rather very concrete and represent an A+ entity in a way :question:

src/main/java/fi/aalto/cs/apluscourses/model/Submission.java
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Constructs a new object instance.
+   *
+   * @param exercise       Exercise.
+   * @param submissionInfo Information for the submission.
+   * @param files          Map from keys to file paths.
+   * @param group          Group in which the submission is made.
+   */
+  public Submission(@NotNull Exercise exercise,
```
superseacat, line 26: is it the same `Submission` entity, that supposed to be used to host the results of the submission later on? If it is, then it could probably be extended with the `id` and as seen earlier `hashCode` and `equals`?

src/test/java/fi/aalto/cs/apluscourses/integration/ApiTest.java
```diff
@@ -15,18 +16,66 @@
...
   }
-}
\ No newline at end of file
+
+  @Test
+  public void testGetExercises() {
+    given()
+        .auth()
+        .preemptive()
+        .basic("root", "root")
```
superseacat, line 39: I suggest authenticating as a user with 'student' permissions

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseActionTest.java
```diff
@@ -0,0 +1,279 @@
...
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+
+public class SubmitExerciseActionTest {
```
superseacat, line 57: sir, I admire you for this!

src/main/java/fi/aalto/cs/apluscourses/utils/ArrayUtil.java
```diff
@@ -0,0 +1,39 @@
+package fi.aalto.cs.apluscourses.utils;
+
+import java.util.function.IntFunction;
+import org.jetbrains.annotations.NotNull;
+
+public class ArrayUtil {
```
superseacat, line 6: emmm... is this tested?

src/main/java/fi/aalto/cs/apluscourses/utils/FactorySelector.java
superseacat: and here are my two cents...

> Hmm... I was just thinking how easily it would solve an issue I was facing (well, not really facing but more like thinking of) with `Authentication`.

I strongly suggest we solve issues **a)** that exist, and **b)** when they appear - the whole point of agile is to deliver "business value" quickly. as an approach, it is totally allowing duplications/repetitions etc. as long as there is a benefit for the "value". it does not really cover the architectural side of things, but from the business POV, there is no benefit of a good architecture if you've lost the competition to the crappy-but-quickly-developed-service.



> This class adds a lot of complexity compared to the benefits it brings (as far as I can see). Personally, I would really prefer a simpler approach even if it means some code duplication etc.

here and in some other places, there are quite complex to perceive generic structures already introduced. They are nice and elegant, but again, **untested**. And eventually, there is no resource for testing them atm (and probably never in the future), not to mention future maintenance burden and being locked mostly to one team member availability. Repeating myself, I'd rather **a)** go with a stupid `for-loop` as a trivial (but no tests needed solution) or some kind of a **b)** library/third-party solution, that has been used 1kk times elsewhere, stable and "not-my-problem".

another point is, that things, for them to be used, should be simple and understandable. in other words, there's little benefit if 2/3 of the code uses the "good-old" `for-loop` approach and one does it with a fancy one...

> So, for integration tests (and in the future in production, too), we need to use different authentication method.

as far as I remember the **RestAssured** might have also some way to auth with token too, but here the "purpose" of testing is a bit exaggerated I think. as long as we can make sure, that for a certain "user" we can get the data we need from A+ -> we're good. the _raison d'etre_ for the integration tests is a bit unconventional...

src/main/java/fi/aalto/cs/apluscourses/utils/FactorySelector.java
```diff
@@ -0,0 +1,36 @@
+package fi.aalto.cs.apluscourses.utils;
+
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class FactorySelector<A, R> {
```
superseacat, line 8: is this tested?

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
OlliKiljunen: I'm also glad this is being discussed.
 
My position is that reliances on 3rd-party code are also a source of maintenance burden, and a really major one in cases where the 3rd-party reliance cannot be encapsulated and only used via meaningful interface (examples of nicely encapsulated 3rd party reliances are how we use https and zip libraries).

IntelliJ's approach to testing is that no real _unit_ tests are used, that is, nothing is tested in isolation but in every test there's the whole platform behind. IMO, that strips unit testing of its greatest advantage, which is, that every _unit_ of software can be implemented, understood and verified in itself, without having to implement, understand and verify the whole software. That's why I dislike `IdeaTestCase`s and I'd restrict their usage in higher-than-unit level testing.

src/main/java/fi/aalto/cs/apluscourses/model/Exercise.java
OlliKiljunen: Yes.

src/main/java/fi/aalto/cs/apluscourses/model/ExerciseDataSource.java
OlliKiljunen: No. `ExerciseDataSource` is an object that is used when communicating with the API. Basically, it is encapsulation of the API whose methods are called to get model objects.

Why do we need it? It is beneficial that the API code is in one place. Abstracting it behind an interface allows us to implement model classes (and everything else) separately from the API (currently not fully done because JSON parsing is still in the model classes but that should be changed). Now, if for example a new (non-backwards-compatible) version of API is published we can just write a new implementation of `ExerciseDataSource` to comply that. The choice between API version could then be dynamically by just using an object of desired class.

src/main/java/fi/aalto/cs/apluscourses/model/Main.java
OlliKiljunen: I thought it would be funny to have model correspondent for `MainViewModel` but now I see it was unnecessary. I'll remove this class.

src/main/java/fi/aalto/cs/apluscourses/model/Submission.java
OlliKiljunen: No, it shouldn't be the same. We discussed that and I think we decided that the other would be called `SubmissionResult`.

src/main/java/fi/aalto/cs/apluscourses/utils/ArrayUtil.java
OlliKiljunen: Soon.

src/main/java/fi/aalto/cs/apluscourses/utils/FactorySelector.java
OlliKiljunen: Soon.

src/main/java/fi/aalto/cs/apluscourses/utils/FactorySelector.java
nikke234: Pourquoi utilisez-vous le français?

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseActionTest.java
```diff
@@ -0,0 +1,279 @@
...
+
+    fileFinder = mock(FileFinder.class);
+    doReturn(filePath).when(fileFinder).tryFindFile(modulePath, fileName);
+    doCallRealMethod().when(fileFinder).findFile(any(), any());
+    doCallRealMethod().when(fileFinder).findFiles(any(), any());
+
+    moduleSource = mock(SubmitExerciseAction.ModuleSource.class);
+    doReturn(new Module[] { module }).when(moduleSource).getModules(project);
+
+    dialogs = new Dialogs();
```
nikke234, line 147: It looks like the assertions about which dialogs are shown have been removed, is there a reason for that?

src/main/java/fi/aalto/cs/apluscourses/model/Submission.java
superseacat: at this stage `SubmissionResult` represents what is closer logically to `ExerciseSubmissionResult`, not the `Submission` itself.

src/main/java/fi/aalto/cs/apluscourses/utils/FactorySelector.java
OlliKiljunen: Excellent points. A few notes:
> I'd rather a) go with a stupid for-loop as a trivial (but no tests needed solution)

I don't see how inlining `for` loops would decrease the need for testing, quite the contrary. (I assume we are talking about loops like `for (int i = 0; i < length; i++) { ... }`. ) I actually think that even small amount of additional logic included in the same method with a `for` loop increases its complexity quickly to a level where the method is no longer (thoroughly) testable. Knowing how `for` loops are often sources of bugs (off-by-one errors, unintentional closures, illegal modifications during iteration, ...) I'd avoid them, if possible, in business logic code.

>b) library/third-party solution, that has been used 1kk times elsewhere, stable and "not-my-problem"

Each library dependency is a real maintenance burden, IMO. This of course doesn't mean libraries shouldn't be used, but the design decision on which libraries are used, by which parts of the code and how, must be made with a careful consideration. Using libraries incautiously makes them easily spread from the parts of the system where they're actually needed to all around the codebase. For better maintainability, I'd like to keep library-reliant code behind [bridges](https://en.wikipedia.org/wiki/Bridge_pattern) or [facades](https://en.wikipedia.org/wiki/Facade_pattern) or such whenever possible.

superseacat:
60%+ of coverage, nice!

thus...

> please, try holding yourself from doing refactoring & feature development in the same PR (or even branch), separate (no matter how tempting and convenient it is not to, remember, you are writing code for others more than for yourself)

OlliKiljunen:
> 
> 
> 60%+ of coverage, nice!
> 
> thus...
> 
> > please, try holding yourself from doing refactoring & feature development in the same PR (or even branch), separate (no matter how tempting and convenient it is not to, remember, you are writing code for others more than for yourself)

But... the reason I refactor is to make classes unit-testable, so... how I should...? Wait, PowerMockito, isn't it? Damn! 😅

superseacat:
> Wait, PowerMockito, isn't it?

it's a bigger question...
on one hand, testability assumes certain code structure, on the other, changing your code just to make it testable :confounded: 

speaking on the matter, I don't believe PowerMockito is a silver bullet, thus I see little reasons why not to try it, do you?

OlliKiljunen:
> it's a bigger question...
> on one hand, testability assumes certain code structure, on the other, changing your code just to make it testable 😖

One of the biggest benefits of unit-testing is, IMO, that it also kinda verifies the design: that code is structured in small enough pieces that are not tightly coupled to each other. It is clear that this kind of clever design doesn't arise from nothing but requires careful engineering. So it is very natural that first implementations of a class or a method are not unit-testable but by improving the design (that is, by refactoring) they are made unit-testable (by loosening the coupling between classes, replacing concrete types with abstract interfaces, applying design patterns, etc).

In my experience, usually, once classes/methods are refactored to be unit-testable, the design is already so clear that the correctness of code (in the layman sense of the term) is almost self-evident and writing actual tests is more like a formality. So I don't see "changing code just to make it testable" as avoidant at all but an essential and prevalent task of SW development.

> speaking on the matter, I don't believe PowerMockito is a silver bullet, thus I see little reasons why not to try it, do you?

The problem with "too powerful" mocking frameworks is that they take away that design-verifying aspect of the unit-tests as they make even tightly-coupled classes "testable". With them, are we actually any longer testing _units_ (which ideally can be thought as a blackbox taking something in and giving something out) but, instead, concrete subsystems of our application (well, plugin)? In the latter case, I'm afraid that testing can assure only if a piece of code is well-functioning _in its current place in the system_ but tells very little if that piece of code can be applied or re-used elsewhere, which is an important point of unit-testing.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
```diff
@@ -49,40 +50,27 @@
...
   public SubmitExerciseAction() {
     this(
         PluginSettings.getInstance(),
-        SubmissionInfo::forExercise,
-        SubmissionHistory::forExercise,
-        Group::getGroups,
         VfsUtil::findFileInDirectory,
-        viewModel -> new ModuleSelectionDialog(viewModel),
-        viewModel -> new SubmissionDialog(viewModel),
+        DefaultModuleSource.INSTANCE,
```
nikke234, line 72: Perfect opportunity for a lambda instead the additional inner class :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/APlusAuthenticationAction.java
```diff
@@ -2,21 +2,45 @@
...
     }
   }
+
+  public MainViewModelProvider getMainViewModelProvider() {
+    return mainViewModelProvider;
+  }
+
+  public Dialogs getDialogs() {
+    return dialogs;
+  }
```
nikke234, line 45: This is a very minor (and @superseacat would most likely strongly disagree with me here :laughing:), but I wouldn't add getters just to test the default constructor, instead I would just test how the `actionPerformed` of a default-constructed `APlusAuthenticationAction` behaves

nikke234: Actually never mind, since the default `actionPerformed` probably can't be tested since it uses all kinds of platform stuff

PR #241, opened 2020-07-22
REPL Welcome and Imports

superseacat:
# Description of the PR

+ REPL Welcome message
+ REPL runs initial commands (from file)
+ allows starting REPL on folders and everywhere else (via right-click popup context menu)

probably some more tests are coming

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [x] Yes
- [ ] Not relevant


src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
```diff
@@ -239,4 +244,19 @@ public void startRegularSubmissionResultsPolling(@NotNull Project project) {
...
     }
   }
+
+  /**
+   * Method that adds a new file (pattern) to the list of files not being shown in the Project UI.
+   *
+   * @param ignoredFileName a {@link String} name of the file to be ignored.
+   * @param project a {@link Project} to ignore the file from.
+   */
+  public static void ignoreFileInProjectView(@NotNull String ignoredFileName,
```
superseacat, line 254: unsure if it's the right place, but where it could then be? :confused:

src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
```diff
@@ -34,4 +36,66 @@ object ModuleUtils {
...
+    moduleFilePath.substring(0, lastIndexOf + 1) // scalastyle:ignore
+  }
+
+  def initialReplCommandsFileExist(@NotNull replCommandsFileName: String,
+                                   @NotNull moduleFilePath: String): Boolean = {
+    val moduleDir = getModuleRoot(moduleFilePath)
+    Files.exists(Path.of(moduleDir + replCommandsFileName))
+  }
+
+  @NotNull
```
superseacat, line 88: @nikke234 I tried to call this guy from Java, but failed... is there some easy way?

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -56,16 +59,21 @@
...
-   *                        ids of the plugins and the values are the names of the plugins.
-   * @param resourceUrls    A map containing URLs to resources related to the course. The keys are
-   *                        the names of the resources and the values are the URLs.
+   * @param name                The name of the course.
+   * @param modules             The list of modules in the course.
+   * @param requiredPlugins     A map containing the required plugins for this course. The keys are
+   *                            the ids of the plugins and the values are the names of the plugins.
+   * @param resourceUrls        A map containing URLs to resources related to the course. The keys
+   *                            are the names of the resources and the values are the URLs.
+   * @param replInitialCommands A {@link HashMap}, with module name {@link String} as a key and a
```
OlliKiljunen, line 74: Does it have to be a `HashMap` or would any implementation of `Map` do?

src/main/scala/fi/aalto/cs/apluscourses/intellij/Repl.scala
```diff
@@ -0,0 +1,20 @@
...
+import com.intellij.openapi.module.Module
+import fi.aalto.cs.apluscourses.intellij.utils.ModuleUtils.{getReplInitialCommandsForModule, getUpdatedText}
+import org.jetbrains.plugins.scala.console.ScalaLanguageConsole
+
+class Repl(module: Module) extends ScalaLanguageConsole(module: Module) {
+
+  override def print(text: String, contentType: ConsoleViewContentType): Unit = {
+    var updatedText = text
+
+    if (text.contains("Type in expressions")) {
```
OlliKiljunen, line 13: This is quite a risky way to check that.

Would it be possible to at least make sure that the initial commands are not used to override actual things more than once per REPL?

src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
```diff
@@ -34,4 +36,66 @@ object ModuleUtils {
     nonEmpty
   }
 
+  def naiveValidate(@NotNull command: String): Boolean =
+    command.matches("import\\so1\\.[a-z]*(\\_|\\._)$")
```
OlliKiljunen, line 40: Should we have some special `//TODO` like syntax for O1 specific code?

src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
```diff
@@ -34,4 +36,66 @@ object ModuleUtils {
...
+
+  def clearCommands(@NotNull imports: Array[String]): Array[String] =
+    imports
+      .clone
+      .map(_.replace("import ", ""))
+      .map(_.replace("._", ""))
```
OlliKiljunen, line 46: Why?

src/main/scala/fi/aalto/cs/apluscourses/intellij/Repl.scala
superseacat: thanks, good point! I'll extend the check to verify the content is of a "WelcomeContentType" to make it sure :bowing_man:

src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
superseacat: well, the actual commands are stored as full commands (and being written to the file as well), however, in the REPL, the message looks like this:
```bash
Welcome to Scala 2.13.1 (OpenJDK 64-Bit Server VM, Java 11.0.7).
Loaded A+ Courses module [Train]. Auto-imported packages [o1, o1.train] for your convenience.
Write a line (or more) of Scala and press [Ctrl+Enter] to run it. Use [Up] and [Down] to scroll through 
your earlier inputs. Changes to the module are not loaded automatically. If you edit the files, restart 
the REPL with [Ctrl+F5] or the icon on the left. 
Type in expressions for evaluation. Or try :help.
```
(contains only the names of the imported packages). the only purpose is to have this text nice-lookin'. safe for now, as the regex you've noticed earlier will allow only imports, regardless of what someone can put into the course config. thus, not sure if it's 100% bulletproof (but at least foolproof it is :wink: )

src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
superseacat: I'm not sure if I fully got you, what did you mean here?

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
superseacat: any `Map` will do, I guess)

src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
OlliKiljunen: Oh, yes sure, that makes sense.

src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
OlliKiljunen: It would be good to somehow track what parts of our codebase have O1 specific code that must get ridden of when the plugin is made available to other courses too. My suggestion was that we marked those line for example with
```java
// O1CODE
command.matches("import\\so1\\.[a-z]*(\\_|\\._)$")
```
so `Ctrl+Shift+F O1CODE` would easily list all of them.

superseacat: a bright idea indeed! :bulb:

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
```diff
@@ -239,4 +244,19 @@ public void startRegularSubmissionResultsPolling(@NotNull Project project) {
...
+                                      @NotNull Project project) {
+    FileTypeManager fileTypeManager = FileTypeManager.getInstance();
+    String ignoredFilesList = fileTypeManager.getIgnoredFilesList();
+    Runnable runnable = () -> fileTypeManager
+        .setIgnoredFilesList(ignoredFilesList + ignoredFileName + ";");
+    WriteCommandAction.runWriteCommandAction(project, runnable);
+  }
```
nikke234, line 261: This causes no errors even if the file doesn't exist, right?

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
nikke234: I guess `PluginSettings` is fine, we can always refactor after Sprint 3. If it's in `PluginSettings`, then perhaps make it non-static and call it as `PluginSettings.getInstance().ignore...`

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
```diff
@@ -159,4 +160,19 @@ public String getVersionId() {
...
+  protected static String[] getReplInitialCommandsForModule(
+      com.intellij.openapi.module.Module module) {
+    return PluginSettings
+        .getInstance()
+        .getMainViewModel(module.getProject())
+        .courseViewModel
+        .get()
+        .getModel()
+        .getCourseReplInitialCommands()
+        .getOrDefault(module.getName(), null);
```
nikke234, line 176: I think `Map#get` returns null by default if the key isn't found

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -445,6 +470,45 @@ public void unregister() {
     // Subclasses may do things.
   }
 
+  @NotNull
+  public Map<String, String[]> getCourseReplInitialCommands() {
+    return replInitialCommands;
+  }
```
nikke234, line 476: Nit: I would rename this to `getReplInitialCommands`. Currently all of the private JSON helper methods include `Course` in the name and the public methods don't

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -167,9 +192,9 @@ public String getName() {
   }
 
   /**
-   * Returns the list of all modules in this course. If the course object is created with
-   * {@link Course#fromConfigurationData}, then the modules are returned in the order in which they
-   * are listed in the course configuration data.
+   * Returns the list of all modules in this course. If the course object is created with {@link
+   * Course#fromConfigurationData}, then the modules are returned in the order in which they are
+   * listed in the course configuration databp.
```
nikke234, line 197: Nit: typo in the last word? :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
```diff
@@ -176,4 +182,25 @@ protected boolean hasLocalChanges(@NotNull ZonedDateTime downloadedAt) {
...
+      File file = new File(getFullPath() + "/"
+          + PluginSettings.MODULE_REPL_INITIAL_COMMANDS_FILE_NAME);
+      try {
+        FileUtils.writeLines(file, StandardCharsets.UTF_8.name(), Arrays.asList(commands));
+      } catch (IOException ex) {
+        logger.error("Could not write REPL initial commands file", ex);
+      }
+      FileUtilRt.createIfNotExists(file);
+    }
+  }
```
nikke234, line 205: This should probably `private`.

Does this mean that we only support start-up commands for modules that are downloadable? An alternative is to create the file when the REPL is launched, then it would work for any module. This can be changed later too though

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
nikke234, line 176: Nit: `module` should have `@NotNull`.

Also thinking about this method, since it accesses the `module` and `project`, it should have a read action.

I think the method could also be moved inside `IntelliJModule`

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -445,6 +470,45 @@ public void unregister() {
...
+      throws MalformedCourseConfigurationFileException {
+    Map<String, String[]> replInitialCommands = new HashMap<>();
+    JSONObject replInitialCommandsJsonObject;
+    try {
+      replInitialCommandsJsonObject = jsonObject
+          .getJSONObject("repl").getJSONObject("initialCommands");
+
+      if (replInitialCommandsJsonObject == null) {
+        return replInitialCommands;
+      }
```
nikke234, line 491: `getJSONObject` throws `JSONException` so `replInitialCommandsJsonObjet` can't possibly be `null` I think. If we want to make it optional, then I would do

```
JSONObject replJsonObject = jsonObject.optJSONObject("repl");
if (replJsonObject == null) {
  return replInitialCommands;
}
JSONObject initialCommandsJsonObject = replJsonObject.optJsonObject("initialCommands");
if (initialCommandsJsonObject == null) {
  return replInitialCommands;
}
...
```

src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
```diff
@@ -34,4 +36,66 @@ object ModuleUtils {
...
+
+  def getCommandsText(@NotNull imports: Array[String]): String =
+    imports.length match {
+      case 0 => ""
+      case 1 => "Auto-imported package [" + imports(0) + "] for your convenience." // scalastyle:ignore
+      case _ => "Auto-imported packages [" + imports.mkString(", ") + "] for your convenience."
+    }
```
nikke234, line 53: Nice! Two nits:

I think replacing `imports(0)` with `imports.head` would get rid of the scalastyle warning.

Also a possibility to use string interpolation
`s"Auto-imported package [${imports.head}] for your convenience."`

src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
nikke234: Weird, I'm struggling to get it to work either

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
superseacat: and will you then face the rage of Sonar instead of me? :rage:

OlliKiljunen: What does Sonar have to say about making this non-static? 🤔

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
superseacat: AFAIK, nope, it's just a name (like a pattern) that is not shown in the project view, if it can't be found, then nothing happens. I'm certain not all the project contain those files that are listed there by default, but it still checks.

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
superseacat: sharp eye! :eye:

src/main/scala/fi/aalto/cs/apluscourses/intellij/Repl.scala
superseacat: made some improvements (a flag and a better `String` to check against), but it will remain hackish, as we're intercepting here...

src/main/java/fi/aalto/cs/apluscourses/intellij/model/IntelliJModule.java
superseacat: this is a good point, however, I think it's good as is because:
* if the module does is not downloaded (e.g. created by hand), then it won't have `replInitialCommands` in our `Module` object (at least with current prospects) - so even if there is a file it still won't have any commands
* if the module like `Ave` is meant to teach students to create a module "from scratch", then they can eventually create a file by hand and populate it with whatever commands/imports they like (and if the file has the right name the commands will be played by the REPL)
* if the intention is to go even more "from scratch", they can learn to use Scala REPL `-i` flag that will allow them to initially execute any script (file with commands)

superseacat: with `protected` I won't have to `PowerMockito` the method if I'll ever get to writing a test for it (which I probably should have done already :rofl: )

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
superseacat: `null instanceOf ...` returns `false` :nerd_face:

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
superseacat: `Change this instance-reference to a static reference. ` but no it's gone... hm...

superseacat: made the change, the issue is gone now, forget it :smile:

src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
```diff
@@ -34,4 +36,66 @@ object ModuleUtils {
...
+
+  def getModuleRoot(@NotNull moduleFilePath: String): String = {
+    val lastIndexOf = moduleFilePath.lastIndexOf("/")
+    moduleFilePath.substring(0, lastIndexOf + 1) // scalastyle:ignore
+  }
+
+  def initialReplCommandsFileExist(@NotNull replCommandsFileName: String,
+                                   @NotNull moduleFilePath: String): Boolean = {
+    val moduleDir = getModuleRoot(moduleFilePath)
+    Files.exists(Path.of(moduleDir + replCommandsFileName))
```
nikke234, line 85: Nit: this should probably be `Path.of(moduleDir, replCommandsFileName)`?

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/ReplAction.scala
```diff
@@ -45,25 +61,49 @@ class ReplAction extends RunConsoleAction {
...
-      case None =>
-        // For now, no dialog is shown for a project level REPL
-        configuration.setName("Scala REPL")
-        RunConsoleAction.runExisting(setting, runManagerEx, project)
+      case None => // For now, no dialog is shown for a project level REPL
     }
+    RunConsoleAction.runExisting(setting, runManagerEx, project)
+  }
+
+  private class ReplConfigurationFactory() extends ConfigurationFactory(getMyConfigurationType) {
```
OlliKiljunen, line 77: To get rid of a warning pointed out by @nikke234 [on Slack](https://apluslms.slack.com/archives/CSDV9CA31/p1595961149005200), this class should override `getId` so that it always returns the same string, for example, `"A+ extended Scala REPL"`.

src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
OlliKiljunen: May be relevant: https://stackoverflow.com/questions/23261075/compiling-scala-before-alongside-java-with-gradle

src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/ReplAction.scala
superseacat: aaaaaaaand, solved, thanks to the handsome gentleman from the Eastside! :kissing_heart:

PR #242, opened 2020-07-23
Fixed syntax highlight bug [#239]

### LINKED ISSUE #239: Scala syntax highlighting doesn't work
This discussion took place in Slack's #intellij channel on Mon 2020-07-14 @ 13:26:

Niklas:
> is it normal that intellij's syntax hilighting doesn't quite work with the current release version? IJ says it cannot resolve swing and similar stuff, but running stuff in REPL works

Olli:
> Are you using our plugin?

Niklas:
> yes, 0.1.813

Olli:
> It might be an issue about the automatic Scala SDK setup the plugin does. I'll check it at some point today or tomorrow.

OlliKiljunen:
# Description of the PR

Fixes #239 

The issue was that syntax highlighting didn't work if (for some reason) if `CLASSES` roots for libraries (see `.idea/libraries/scala_sdk_2_13_1.xml`  for some test project) used `file:` URL protocol instead of `jar:`. (_Interestingly_, Scala plugin uses `file:` protocol in `compiler-classpath`).

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
@@ -3,14 +3,15 @@ package fi.aalto.cs.apluscourses.intellij.model;
 import com.intellij.openapi.roots.libraries.LibraryProperties;
 import com.intellij.openapi.roots.libraries.PersistentLibraryKind;
 import com.intellij.openapi.util.io.FileUtilRt;
+import com.intellij.openapi.vfs.VfsUtil;
 import fi.aalto.cs.apluscourses.utils.DirAwareZipFile;
 import java.io.File;
 import java.io.IOException;
-import java.net.URI;
 import java.net.URL;
 import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Objects;
+import java.util.function.Function;
 import java.util.stream.Stream;
 import org.apache.commons.io.FileUtils;
 import org.jetbrains.annotations.CalledWithWriteLock;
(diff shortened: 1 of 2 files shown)


src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
```diff
@@ -80,24 +78,28 @@ public String getFileName() {
 
   @Override
   protected String[] getUris() {
-    return getUris(CLASSES);
+    return getUris(CLASSES, path -> VfsUtil.getUrlForLibraryRoot(path.toFile()));
```
OlliKiljunen, line 81: I don't really like to use platform's implementation for something that should just convert `file:///foo/bar/baz.jar` to `jar:///foo/bar/baz.jar!/`. The thing is, this format of JAR URI that is used by IntelliJ IDEA is _non-standard_ (there should be the whole URL, including the scheme, after `jar:`) and IDEA doesn't support standard JAR URLs (For standard, see [these](https://www.iana.org/assignments/uri-schemes/prov/jar) [sources](http://java.sun.com/javase/6/docs/api/java/net/JarURLConnection.html) referenced by [Wikipedia](https://en.wikipedia.org/wiki/List_of_URI_schemes)).

As I'm not aware of any existing specs for this non-standard URL version, it is not meaningful to write an own implementation based on vague guess of how it's probably meant to work. Fortunately, there is this utility function available in IntelliJ's OpenAPI, so we can use it and assume it's compatible with whatever is the "IntelliJ's way" here.

PR #257, opened 2020-07-27
Exercise list icons

nikke234:
# Description of the PR

This PR adds icons to the exercise list that are dependent on the points and number of submissions that the user has for the exercise. Points are also now fetched alongside exercises.

Credits for the icons once again belong to @superseacat 

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
```diff
@@ -18,6 +18,38 @@ public String getPresentableName() {
     return APlusLocalizationUtil.getEnglishName(getModel().getName());
   }
 
+  /**
+   * Returns {@code true} if the exercise is submittable from the plugin, {@code false} otherwise.
+   */
+  public boolean isSubmittable() {
+    String presentableName = getPresentableName();
+    return presentableName.length() > "Assignment xx (".length()
```
OlliKiljunen, line 26: Do we have a plan for how this will be eventually unhacked?

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
```diff
@@ -18,6 +18,38 @@ public String getPresentableName() {
...
+    NO_SUBMISSIONS,
+    NO_POINTS,
+    PARTIAL_POINTS,
+    FULL_POINTS,
+  }
+
+  /**
+   * Returns a {@link Status} that indicates the points status of this exercise.
+   */
+  public Status getStatus() {
```
OlliKiljunen, line 40: This method and `isSubmittable()` would make more sense in model. They do not only give information of how the state of the exercise is rendered for the user (view model things) but of the exercise object itself.

src/main/resources/META-INF/icons/noSubmissions_dark.svg
```diff
@@ -0,0 +1,488 @@
...
+     id="path919"
+     inkscape:connector-curvature="0" />
+  <path
+     style="opacity:1;fill:#ffffff;fill-opacity:1;stroke-width:0.00097656"
+     d=""
```
OlliKiljunen, line 485: Nit: This `path` seems to be empty.

src/test/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModelTest.java
```diff
@@ -20,4 +22,36 @@ public void testGetPresentableName() {
...
   }
 
+  @Test
+  public void testIsSubmittable() {
+    ExerciseViewModel ex1 = new ExerciseViewModel(
+        new Exercise(1, "|en:Assignment 13|fi:Tehtava 13|", Collections.emptyList(), 0, 0, 0));
+    ExerciseViewModel ex2 = new ExerciseViewModel(new Exercise(
+        2, "|en:Assignment 13 (Test)|fi:Tehtava 13 (Test)|", Collections.emptyList(), 0, 0, 0));
+    ExerciseViewModel ex3 = new ExerciseViewModel(new Exercise(
+        2, "|en:Assignment 1 (Piazza)|fi:Tehtava 1 (Piazzza)|", Collections.emptyList(), 0, 0, 0));
```
OlliKiljunen, line 32: Typo in the Finnish name?

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
nikke234: I initially wanted to put `isSubmittable` in the model class, but I decided against it for the following reasons:

1. The `isSubmittable` method uses a hacky solution, and therefore trying to submit a disabled exercise is still possible, just in case that `isSubmittable` produces an incorrect result.
2. This hacky method used to determine whether to show the exercise as enabled or not, may cause confusion with the `isSubmittable` method from `SubmissionInfo`, which actually looks at the submission form files.

Any opinions?

`getStatus` could definitely go in the model class, although I wonder if there is any benefit from having a separate `Status` enum, when all of the values can be figured out from the values from the exercise...

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
nikke234: Nothing concrete yet at this point. Checking submittability requires making a request to the individual exercise API end point, so most likely some sort of caching would be highly preferable at that point.

src/test/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModelTest.java
nikke234: Oops, nice catch

src/main/resources/META-INF/icons/noSubmissions_dark.svg
nikke234: How did you find this :smile: did you actually read through the entire svg??

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesTreeRenderer.java
```diff
@@ -5,12 +5,29 @@
...
+      case NO_SUBMISSIONS:
+        return PluginIcons.A_PLUS_NO_SUBMISSIONS;
+      case NO_POINTS:
+        return PluginIcons.A_PLUS_NO_POINTS;
+      case PARTIAL_POINTS:
+        return PluginIcons.A_PLUS_PARTIAL_POINTS;
+      case FULL_POINTS:
+        return PluginIcons.A_PLUS_FULL_POINTS;
+      default:
+        return null;
```
superseacat, line 27: `@NotNull`?

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/GetSubmissionsDashboardAction.java
```diff
@@ -44,47 +33,11 @@ public GetSubmissionsDashboardAction(@NotNull MainViewModelProvider mainViewMode
...
-    APlusAuthentication authentication =
-        (APlusAuthentication) exerciseDataSource.getAuthentication();
-
-    Course course = requireNonNull(mainViewModel.getCourseViewModel().get()).getModel();
-
-    SubmissionResultsList submissionResultsList = tryGetSubmissionsDashboard(
-        Long.parseLong(course.getId()), authentication, requireNonNull(project));
-    course.setSubmissionsDashboard(requireNonNull(submissionResultsList));
+    /*
+    This action will eventually be a generic update action, which updates the main view model. For
```
superseacat, line 37: why not to remove the class completely?

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -35,7 +35,7 @@
   private final List<Module> modules;
 
   @NotNull
-  private SubmissionResultsList submissionResultsList;
+  private Points submissionResultsList;
```
superseacat, line 38: `Points`? :smile:

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesTreeRenderer.java
nikke234: This should never happen, but you're right, this is silly, I'll throw `IllegalStateException`

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
nikke234: Oops, nice catch :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/GetSubmissionsDashboardAction.java
nikke234: Hmm good point, it's still in the git history should we need it again

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesTreeRenderer.java
superseacat: `throw null` :wink:

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
```diff
@@ -18,6 +18,38 @@ public String getPresentableName() {
     return APlusLocalizationUtil.getEnglishName(getModel().getName());
   }
 
+  /**
+   * Returns {@code true} if the exercise is submittable from the plugin, {@code false} otherwise.
+   */
+  public boolean isSubmittable() {
+    String presentableName = getPresentableName();
+    return presentableName.length() > "Assignment xx (".length()
+        && !"Assignment 1 (Piazza)".equals(presentableName);
```
superseacat, line 27: could you please extract both `String`s as constants (method/class level) and try coming out with some names that would give additional context (maybe like `A_BEGINNING_OF_THE_SUBMITTABLE_EXERCISE_NAME` or smth) :grin:

src/test/java/fi/aalto/cs/apluscourses/model/ExerciseTest.java
```diff
@@ -11,53 +16,101 @@
 
   @Test
   public void testExercise() {
-    Exercise exercise = new Exercise(987, "def");
+    Exercise exercise = new Exercise(987, "def", Collections.emptyList(), 0, 0, 0);
```
superseacat, line 19: an alternative could be having a constructor with fewer values and a bunch of setters to use, a least, this way it would be clear what value is being set in particular :grey_question:

superseacat:
icons look charming and eventually the code showing them is nice too! :two_hearts: 

I would like to propose one optional idea, that IMO could improve the UX a bit: the same way that the tooltip works for Modules, make a tooltip for exercises/assignments describing on-hover their status, like "Assignment (Cube) for Module (Subprograms)"

Untitled

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
OlliKiljunen: Good reasoning, let's keep hackiness out of the model. When it comes to `getStatus`, I think both of the viewpoints are valid. We can think the status resolving as a part of the business logic or just part of the presentation logic.

So no need to change.

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesTreeRenderer.java
OlliKiljunen: `IllegalArgumentException` is perhaps more accurate.

superseacat:
currently, there is a couple of icons for each status (depending on whether there are any submission attempts left or not), thus the UI shows all the icons with the "black" element (cross/tick) inside, how could I test the ones with the "white" elements in them?

the "Assignment 26 (Chess)" supposed to have a "white" cross inside (3 out of 10), right?
Screenshot from 2020-07-27 21-25-26

superseacat:
ou... :rofl: 
Screenshot from 2020-07-27 21-28-14

src/main/resources/META-INF/icons/noSubmissions_dark.svg
OlliKiljunen: We take code reviewing seriously here, don't we? :joy:

src/test/java/fi/aalto/cs/apluscourses/model/ExerciseTest.java
OlliKiljunen: We want to keep (at least certain) fields `final` so adding setters to the model class itself is not an option.

A separate builder class (preferably with 'fluent interface') could be used, though (for other classes too). It has been on my mind but in that 'Refactor after Sprint 3' section.

nikke234:
> icons look charming and eventually the code showing them is nice too! two_hearts
> 
> I would like to propose one optional idea, that IMO could improve the UX a bit: the same way that the tooltip works for Modules, make a tooltip for exercises/assignments describing on-hover their status, like "Assignment (Cube) for Module (Subprograms)"

@superseacat Good idea, will do!

nikke234:
> ou... :rofl: 

@superseacat As you noticed, I ended up using the NMS icons as the light theme icons, but I'm not sure about this, let's discuss this in tomorrows daily :smile:

PR #286, opened 2020-07-30
Token is kept in RAM if cannot store on disk  [#258]

### LINKED ISSUE #258: If token cannot be stored into persistent storage, keep it in memory

OlliKiljunen:
# Description of the PR

Fixes #258 

This is a quicker-and-dirtier solution to solve the issue.

So the token is stored in RAM in any case but written also into secure safe (keyring) if possible (keyring is unlocked). It is also read from there when course config file is fetched (the API URL is used as id in keyring, and eventually that URL is read from course config file).

If keyring is unlocked, everything just works. If keyring is locked, IDE asks to open it when
1. project is turned into A+ project
2. every time a new token is set via dialog
3. when IDE (re)starts and an A+ project is open

`PasswordStorage::store` method (that is called to store token into keyring) should return false if token was not saved. If it returns false, a notification is shown to user that token wasn't permanently stored and it must be set again once the IDE is started again.
Unfortunately, the current implementation of that method doesn't really work and it apparently returns always true. That means, the notification is not shown. Otherwise, things work as expected.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [ ] Not relevant


superseacat:
does "reset A+ Course Plugin Settings" affects the token?

src/main/java/fi/aalto/cs/apluscourses/dal/TokenAuthentication.java
```diff
@@ -0,0 +1,55 @@
+package fi.aalto.cs.apluscourses.dal;
+
+import fi.aalto.cs.apluscourses.model.Authentication;
+import java.util.Arrays;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public abstract class TokenAuthentication implements Authentication {
```
nikke234, line 8: I wonder if this could just be a concrete class with the `APlusTokenAuthentication` implementation. As far as I can tell, we use either `APlusTokenAuthentication` or `Authentication` everywhere (even in tests), so this seems a bit like adding an abstract `Mammal` class between `Animal` and `Dog`; it makes sense but is it needed

src/main/java/fi/aalto/cs/apluscourses/dal/APlusTokenAuthentication.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.apache.http.HttpRequest;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class APlusTokenAuthentication extends TokenAuthentication {
+
+  public static final String AUTHORIZATION_HEADER = "Authorization";
+
+  public static final String APLUS_USER = "A+_user";
```
nikke234, line 11: Wasn't this supposed to be the API URL of the course or was that somewhere else?

OlliKiljunen: That is another field. For one service, there could be multiple credentials (that is, different users) stored. As discussed with @superseacat today, we could, in future, store actual user name to this field (once it's fetched from the API) allowing multiple users from the same IDE (submitting the same solutions? ;) ).

nikke234: Ah right, that makes sense

src/main/java/fi/aalto/cs/apluscourses/dal/TokenAuthentication.java
OlliKiljunen: Just trying to apply single-responsibility principle here. I think `AuthenticationViewModel` (and hence the corresponding view, too) refer to this class (via its factory) instead of `APlusTokenAuthentication`. That way, if we have to change the implementation of `APlusTokenAuthentication` (for example, because of API changes) we can restrict more classes out of those that are potentially affected by that change.

But if you want, I can change this as you suggested.

OlliKiljunen:
> does "reset A+ Course Plugin Settings" affect the token?

No, I forgot about that one. Currently, it should be quite easy to implement but things get a bit trickier when there can be multiple tokens to be stored (for different servers/users). Basically we should store a list of authentications stored, and when resetting, wipe those out and tjen wipe the list itself.

What kind of implementation you'd like to see in this PR?

src/main/java/fi/aalto/cs/apluscourses/dal/TokenAuthentication.java
superseacat: don't sag that easily :wink:

nikke234: I think the public interface of `TokenAuthentication` already provides the possibility to change the implementation without changing other classes, as the new implementation just needs to keep the same public interface. Also, the factory can return some other instance of `TokenAuthentication` if needed, even if `TokenAuthentication` itself already provides an implementation.

But this is anyways a minor thing and I've approved as-is, so your call :smile:

Issue #194, opened 2020-05-28
Upgrade versions


superseacat:
- [ ] `intellij gradle` plugin (mind the [MENU_ACTIONS bug](https://github.com/Aalto-LeTech/intellij-plugin/issues/178))
- [ ] IDEA own version
- [ ] Scala plugin version to compatible with the previous one
- [ ] re-check the Scala deps are of a correct version (`2.12.10` currently)
- [ ] A+ Courses (our) plugin version to be `1.x.y` check it works on JB's plugin repo (and in Travis)
- [ ] **ensure Plugin version will be compatible with the new coming IDEA versions (say 2020.3), smth like an open ended 'supported versions' interval**
- [ ] (optional) versions of other plugins and dependencies

superseacat:
there is a broader task for this (related to the August release) #231

Issue #178, opened 2020-05-19
MENU_ACTIONS bug


OlliKiljunen:
Steps to reproduce (according to Nikolas):
```
runIde and create a new project
Import GoodStuff
File -> Invalidate Caches / Restart -> Just Restart
Import ForLoops
"Modules" and "Exercises" tool windows disappear
Exiting the IDE causes the following exception: java.lang.IllegalStateException: There's no registered action with id=fi.aalto.cs.apluscourses.intellij.actions.ActionGroups.MENU_ACTIONS
```

Stacktrace (from a similar situation):
```
java.lang.IllegalStateException: There's no registered action with id=fi.aalto.cs.apluscourses.intellij.actions.ActionGroups.MENU_ACTIONS
	at com.intellij.openapi.actionSystem.impl.actionholder.IdActionRef.getAction(IdActionRef.java:19)
	at com.intellij.openapi.actionSystem.impl.ActionMenu.fillMenu(ActionMenu.java:319)
	at com.intellij.openapi.actionSystem.impl.ActionMenu$MenuListenerImpl.menuSelected(ActionMenu.java:276)
	at java.desktop/javax.swing.JMenu.fireMenuSelected(JMenu.java:1036)
	at java.desktop/javax.swing.JMenu$MenuChangeListener.stateChanged(JMenu.java:1116)
	at java.desktop/javax.swing.DefaultButtonModel.fireStateChanged(DefaultButtonModel.java:352)
	at java.desktop/javax.swing.DefaultButtonModel.setSelected(DefaultButtonModel.java:234)
	at java.desktop/javax.swing.JMenu.setSelected(JMenu.java:291)
	at java.desktop/javax.swing.JMenu.menuSelectionChanged(JMenu.java:1187)
	at com.intellij.openapi.actionSystem.impl.ActionMenu.menuSelectionChanged(ActionMenu.java:202)
	at java.desktop/javax.swing.MenuSelectionManager.setSelectedPath(MenuSelectionManager.java:129)
	at java.desktop/javax.swing.plaf.basic.BasicMenuUI$Handler.mousePressed(BasicMenuUI.java:508)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6648)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3345)
	at java.desktop/java.awt.Component.processEvent(Component.java:6416)
	at java.desktop/java.awt.Container.processEvent(Container.java:2263)
[...]

PR #295, opened 2020-08-03
Update components

superseacat:
# Description of the PR

+ **todo: remove Travis Env variables**
+ most of the stuff listed here: https://github.com/Aalto-LeTech/intellij-plugin/issues/194
+ due to the changes to the </depends> mechanics of IDEA since 2020.2, removed most of the code (`Actions`) responsible for checking the required plugins

this might become an issue, if/when some other A+ courses would need other "required" plugins, that will be "bad" to install overall for some reason (unlikely)

+ could not reproduce the https://github.com/Aalto-LeTech/intellij-plugin/issues/178 with `0.4.21` can you @nikke234?
+ ~~feel a need for some kind of an agreement, if we leave the functionality of `requiredPlugins` in the `Course` (and course config file at all) (**if not**, then that part needs a cleanup and also TESTING.md)~~
+ did some basic manual testing on Ubuntu and Win, will do the proper one once the latest parts of the "Submissions from the IDE" are merged (pre-merge to `master` if lucky)

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [ ] Not relevant


build.gradle
```diff
@@ -29,12 +33,15 @@ dependencies {
     testCompile group: 'org.powermock', name: 'powermock-module-junit4', version: '2.0.2'
     testCompile group: 'org.powermock', name: 'powermock-api-mockito2', version: '2.0.2'
     testCompile group: 'org.scala-lang', name: 'scala-library', version: '2.12.10'
-    testCompile group: 'io.rest-assured', name: 'rest-assured', version: '4.2.0'
+    testCompile('io.rest-assured:rest-assured:4.2.0') {
+        exclude group:'org.codehaus.groovy', module: 'groovy-xml'
+    }
```
nikke234, line 38: What does this change do?

build.gradle
```diff
@@ -2,14 +2,18 @@ plugins {
...
     id 'maven-publish'
 }
 
 group 'fi.aalto.cs'
-version '0.1.' + (System.getenv('TRAVIS_BUILD_NUMBER') ?: '0')
+version = System.getenv('TRAVIS_BUILD_NUMBER')
+        //  1379 is not a "magic number" it's # of Travis last build
+        //  on the moment of creating this code :kissing_heart:
+        ? ('1.0.' + (System.getenv('TRAVIS_BUILD_NUMBER') - 1379))
+        : '1.0.0'
```
nikke234, line 16: I suggest we just get rid of the build number here and go with standard `MAJOR.MINOR.PATCH` versioning. A single bug fix may increment the build number many times, but we actually only want to add +1 to the patch number with a single bugfix.

src/main/resources/META-INF/plugin.xml
```diff
@@ -1,23 +1,36 @@
 <idea-plugin require-restart="true">
     <id>fi.aalto.cs.intellij-plugin</id>
     <name>A+ Courses</name>
-    <vendor email="intellij-plugin-letech@aalto.fi" url="https://research.cs.aalto.fi/LeTech/">LeTech Group at Aalto University</vendor>
+    <vendor email="intellij-plugin-letech@aalto.fi"
+            url="https://research.cs.aalto.fi/LeTech/">LeTech Group at Aalto University</vendor>
+    <!--  this is the range of supported IntelliJ IDEA version the plugin can be used with  -->
+    <idea-version since-build="202.6397.94"/>
```
nikke234, line 7: Does this mean versions earlier than 2020.2 won't work?

build.gradle
nikke234: Although now that I think about it, removing the build number probably would mean also removing the travis plugin publishing process

src/main/resources/META-INF/plugin.xml
superseacat: yes, my thinking is, that we explicitly aim the latest available major of IDEA. as the students are to use CE, then anyone can get 2020.2 wo any issues (or upgrade seamlessly to it)

build.gradle
superseacat: I'm glad you asked :smile: 

as you know, Gradle is built with Groovy, and once I've updated `intellij-idea gradle` plugin to the latest (4.2.1) it somehow added another version of groovy to the classpath. and as rest-assured turned out to also use gradle it led to some nasty errors and inability to compile.

the mentioned change explicitly excludes the version of `groovy-xml` that was ruining the party :balloon:

superseacat:
> I suggest we incorporate the reviewed and fixed UI strings and some minor UX improvements before we bump the major version to 1.
> 
the recent promise is to release in two weeks again and in three-four once again, so this proposal sounds like a good plan to me :+1: 

> I would also personally prefer to get rid of the build number in our versioning, we can discuss that.

I have no strong feelings about the (semantic) versioning, let's wait until @OlliKiljunen is back and then decide :question:

nikke234:
> I have no strong feelings about the (semantic) versioning, let's wait until @OlliKiljunen is back and then decide question

Sounds good :+1:

build.gradle
OlliKiljunen: I was going to suggest a versioning scheme `MAJOR.MINOR.PATCH+BUILD` like `1.0.3+319`, which is also a [standard](https://semver.org/#spec-item-10). (In **About** window, it could have been shown as **Version: 1.0.3 (build 319)**.) According to SemVer standard, two versions differing only by `+BUILD` part would be considered equal in precedence, which ideally would mean that IntelliJ _would not_ suggest an upgrade in case there's no increment in other parts.

Unfortunately, it seems that IntelliJ does not follow this standard (as far as I understand [this](https://github.com/JetBrains/intellij-community/blob/master/platform/util-rt/src/com/intellij/util/text/VersionComparatorUtil.java)), which means that the version upgrade is suggested also in case if only `+BUILD` number is incremented. That makes me think, perhaps, it's best to follow simple (manually assigned) `MAJOR.MINOR.PATCH` scheme (without `BUILD` number at all) because, as @nikke234 said, that would error publish stage of a test/master build if `PATCH` has been forgotten to change, which seems like a good thing. So I agree with @nikke234 here.

It would be possible to, nonetheless, store build number separately to be shown in the UI (in About window) if we are hyper-afraid of ending up in a situation where we don't know which version of our source code is behind some weird bug an user encounters. But as there should be guaranteed 1-to-0..1 mapping between build numbers and `MAJOR.MINOR.PATCH` strings (apart from test/master releases), that should be unnecessary.

One more thing: we could label `test` versions with some [pre-release identifier](https://semver.org/#spec-item-9) such as `1.0.4-test`. Is this something we would see useful? IntelliJ should handle that as expected (`1.0.4` < `1.0.5-test` < `1.0.5`).

build.gradle
OlliKiljunen: I can't say if this is a correct way to handle this (perhaps it is) but as long as this is only about testing and tests works, I don't care too much.

superseacat: well, I'm not sure if it's a correct way either, but it works, does int make it correct :rofl: ?

OlliKiljunen: Well, I think it is possible that, after this change, some features of `rest-assured` do not work as expected but as this concerns only testing, it's enough that our tests pass. (If this was about production code, I would call for more thorough reasoning/scrutiny on the issue.)

build.gradle
superseacat: anything is fine as long as: a) IntelliJ IDEA will propose an update to users, b) we can unambiguously understand what is the contained code

Issue #287, opened 2020-07-31
tooltips for "update Modules/Exercises" changes


superseacat:
"Refresh Modules" & "Refresh Exercises"

PR #300, opened 2020-08-06
Resource bundle

superseacat:
# Description of the PR

**this now awaits @jsorva 's review**

+ `Actions`
+ `UI`
+ `Notifications`
+ `REPL`
+ `plugin.xml`
+ `Presentation`

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


src/main/java/fi/aalto/cs/apluscourses/intellij/notifications/BetaVersionWarning.java
```diff
@@ -16,14 +18,10 @@
...
-        "You are using version " + version + " of A+ Courses plugin, "
-            + "which is a pre-release version of the plugin and still under development. "
-            + LineSeparator.getSystemLineSeparator().getSeparatorString()
-            + "Some features of this plugin are still probably missing, "
-            + "and the plugin is not yet tested thoroughly. "
-            + "Use this plugin with caution and on your own risk!",
+    super(
+        getText("general.aPlus"),
+        getText("notification.BetaVersionWarning.title"),
+        getAndReplaceText("notification.BetaVersionWarning.content", version),
```
nikke234, line 24: Can't we remove this notification since we removed the version checking activity?

superseacat: I think I did it already (maybe in another branch, but will re-check, thanks)

src/main/resources/META-INF/plugin.xml
```diff
@@ -76,19 +76,19 @@
...
+                    text="Submit A+ Assignment"
                     icon="PluginIcons.A_PLUS_UPLOAD"/>
             <action id="fi.aalto.cs.apluscourses.intellij.actions.OpenSubmissionAction"
                     class="fi.aalto.cs.apluscourses.intellij.actions.OpenSubmissionAction"
                     text="Open A+ Submission"
                     icon="PluginIcons.A_PLUS_BROWSE" />
             <action id="fi.aalto.cs.apluscourses.intellij.actions.ExercisesRefreshAction"
                     class="fi.aalto.cs.apluscourses.intellij.actions.RefreshAction"
-                    text="Update Exercises"
+                    text="Update Assignments"
```
nikke234, line 91: Perhaps this PR could also change this (and Update Modules) to "Refresh x" as described in #287

superseacat:
> 1. `test` contains one UI addition, which probably should also go into the resource file

language selection?

> 2. Should the tests that look at the contents (for an example the tests that check the titles of notifications) actually assert that the contents are equal to the contents in the resource file instead of hard-coded strings like now? This isn't super important though

don't know, but was quite helpful when extracting `Strings` this time, and not a huge burden to change

src/main/java/fi/aalto/cs/apluscourses/presentation/module/ModuleListElementViewModel.java
```diff
@@ -34,35 +36,35 @@ public String getStatus() {
...
+        return getText("presentation.dependencyStatus.depInitial");
       case Component.DEP_WAITING:
-        return "Waiting for dependencies...";
+        return getText("presentation.dependencyStatus.depWaiting");
       case Component.DEP_LOADED:
-        return "Installed";
+        return getText("presentation.dependencyStatus.depLoaded");
       default:
-        return "Error in dependencies";
+        return getText("presentation.dependencyStatus.depError");
```
OlliKiljunen, line 67: Nit: I'd perhaps prefer `getText()` taken out of the `switch` block. If we wanted to keep `return`s inside `case`s, we could introduce a private function that solves the key string using `switch ... case` and the public method woukd call that and pass its return value to `getText`.

On the other hand, it's also a good thing that key strings doesn't appear in our code elsewhere but strictly lexically surrounded by `getText(...)`. I don't even know, if compiler can actually inline those, in which case this would be a very well-justified approach.

src/main/java/fi/aalto/cs/apluscourses/intellij/notifications/ApiTokenNotSetNotification.java
```diff
@@ -8,11 +10,10 @@
    * Constructs a notification that tells authentication is not set.
    */
   public ApiTokenNotSetNotification() {
-    super("A+",
-        "API token not stored",
-        "Token could not be persistently stored. You will be requested to paste the token again "
-        + "next time you'll open the project. To allow token to be securely stored in your "
-        + "machine, check your keyring settings.",
+    super(
+        getText("general.aPlus"),
```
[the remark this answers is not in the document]
superseacat: good point:

it's both... shown in the UI as in the following screenshot, but used only for grouping the notifications (technical) and does not need localization -> can be a separate constant (say in `PluginSettings`)
Screenshot from 2020-08-12 08-34-04

PR #301, opened 2020-08-06
Add submission count and points to exercise list

nikke234:
# Description of the PR

Adds submission count and points to the exercise list, as well as some useful tooltip texts.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
@@ -64,6 +64,13 @@ public class ExerciseViewModel extends SelectableNodeViewModel<Exercise> impleme
     }
   }
 
+  @NotNull
+  public String getStatusText() {
+    Exercise exercise = getModel();
+    return "" + exercise.getSubmissionResults().size() + " of " + exercise.getMaxSubmissions()
+        + ", " + exercise.getUserPoints() + "/" + exercise.getMaxPoints();
+  }
+
   public List<SubmissionResultViewModel> getSubmissionResultViewModels() {
     return submissionResultViewModels;
   }
--- a/src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
@@ -42,6 +42,7 @@ public class ExerciseViewModel extends SelectableNodeViewModel<Exercise> impleme
   }
 
   public enum Status {
+    OPTIONAL_PRACTICE,
     NO_SUBMISSIONS,
     NO_POINTS,
     PARTIAL_POINTS,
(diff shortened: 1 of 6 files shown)


superseacat:
cool stuff :sunglasses: 

a couple of questions:
* do you think it would make sense to additionally show also a per-submission data (results), as currently there is no way to tell which of those submissions were good/bad
* is the best or the last submission result is currently being shown for assignment

superseacat:
two submissions out of zero, that give zero points out of zero points and "pass" (green - full points icon)

Screenshot from 2020-08-06 19-57-28

nikke234:
> cool stuff :sunglasses:
> 
> a couple of questions:
> 
>     * do you think it would make sense to additionally show also a per-submission data (results), as currently there is no way to tell which of those submissions were good/bad
> 
>     * is the best or the last submission result is currently being shown for assignment
> 
>     * and a hypothetical one, if a student makes a submission, which is worth 100% points, but it's overdue, so gets -100% points (zero), how is it shown?

Thanks!

To the first question, yes, it would definitely make sense, but it also involves using a different end point and adding caching, so it requires a lot more effort :smile: 

The points should always show the points of the best submission made.

Also, when points are deducted due to a late submission, it should show 0 points, as that would also be the points that the API reports.

nikke234:
> two submissions out of zero, that give zero points out of zero points and "pass" (green - full points icon)
> 
> Screenshot from 2020-08-06 19-57-28

Okay that admittedly looks a bit weird :smile: The training exercises are a bit tricky, as they have `max_points` 0 and `max_submissions` 0. I'll think about a better way to display them

superseacat:
> The training exercises are a bit tricky, as they have `max_points` 0 and `max_submissions` 0. I'll think about a better way to display them

an [image](https://github.com/JetBrains/intellij-plugins/blob/master/ide-features-trainer/res/img/featureTrainerToolWindow.svg) to underline it's a training assignment?

superseacat:
> Also, when points are deducted due to a late submission, it should show 0 points, as that would also be the points that the API reports.

but what would be the icons then: full points, zero points or partial?

nikke234:
> > The training exercises are a bit tricky, as they have `max_points` 0 and `max_submissions` 0. I'll think about a better way to display them
> 
> an [image](https://github.com/JetBrains/intellij-plugins/blob/master/ide-features-trainer/res/img/featureTrainerToolWindow.svg) to underline it's a training assignment?

Yes I think this is the way to go, good idea. Showing the points icons makes no sense anyways, as it will show the full points icon even if the submission is completely incorrect, since you still have 0/0 points :smile:

nikke234:
> > Also, when points are deducted due to a late submission, it should show 0 points, as that would also be the points that the API reports.
> 
> but what would be the icons then: full points, zero points or partial?

I have to verify this, but I'm 99% the icon will also show zero points, so the only way to know that a submission was correct after the deadline

nikke234:
> the code itself is clearly OK!
> 
> could you please extract the ideas we discussed as separate issues (tickets) for the future, and then eventually feel free to merge :1st_place_medal:

Will do :+1:

superseacat:
> > > Also, when points are deducted due to a late submission, it should show 0 points, as that would also be the points that the API reports.
> > 
> > 
> > but what would be the icons then: full points, zero points or partial?
> 
> I have to verify this, but I'm 99% sure the icon will also show zero points, so the only way to know that a submission (after the deadline) was correct, is to actually open the submission and look at the feedback.

if only there would be a way to learn that wo opening the submission :wink:

PR #306, opened 2020-08-11
Language selection

nikke234:
# Description of the PR

Adds language selection to the course project dialog. This language gets then used when submitting exercises. The language can be changed by using the "Turn project into A+ course project" feature again and selecting a different language.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


nikke234:
> This is a very fine solution. +1
> 
> I first wondered if it would make more sense that `SubmissionInfo` just had a list of `SubmittableFile`s and each `SubmittableFile` had map from languages to (localized) filenames but now when I think about it, this approach of yours feels actually better.

Thanks! Keeping the lists of files for different languages in `SubmissionInfo` also makes it possible to have different length lists for different languages, although I'm not aware of any exercise that does this, or whether it's even possible in A+ :smile:

src/main/java/fi/aalto/cs/apluscourses/ui/courseproject/CourseProjectView.form
```diff
@@ -75,18 +69,45 @@
...
+            <preferred-size width="-1" height="6"/>
+          </grid>
+        </constraints>
+      </vspacer>
+      <component id="dd1bf" class="javax.swing.JLabel">
+        <constraints>
+          <grid row="3" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="0" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
+        </constraints>
+        <properties>
+          <text value="Select the language used for the course:"/>
```
superseacat, line 89: I would suggest to be more specific here, that this is a language for assignment submissions or so... ?

nikke234: That's definitely a good point! It may not be clear what this means. However, I'm also not sure if the students know (or need to know) that A+ needs to know the submission language. I suggest we ask about it in the weekly, this will be super easy to change with the new properties file.

PR #307, opened 2020-08-11
Fix isSubmittable with an even more hacky & O1-specific solution

nikke234:
lol

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
@@ -36,9 +36,11 @@ public class ExerciseViewModel extends SelectableNodeViewModel<Exercise> impleme
    * Returns {@code true} if the exercise is submittable from the plugin, {@code false} otherwise.
    */
   public boolean isSubmittable() {
-    String presentableName = getPresentableName();
-    return presentableName.length() > "Assignment xx (".length()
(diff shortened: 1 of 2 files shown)


src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
```diff
@@ -36,9 +36,11 @@ public String getPresentableName() {
    * Returns {@code true} if the exercise is submittable from the plugin, {@code false} otherwise.
    */
   public boolean isSubmittable() {
-    String presentableName = getPresentableName();
-    return presentableName.length() > "Assignment xx (".length()
-        && !"Assignment 1 (Piazza)".equals(presentableName);
+    // O1_SPECIFIC
+    String name = getPresentableName();
+    int maxSubmissions = getModel().getMaxSubmissions();
+    return name.length() > "Assignment xx (".length() && !"Assignment 1 (Piazza)".equals(name)
```
superseacat, line 42: can this be a method that we would "all know about but pretend it does not exist and never open it again"?

superseacat: the wider question: do we need this to be a flag on A+ side instead of all this... ?

nikke234: Yes I agree :smile: as far as we know, this method doesn't exist

Issue #249, opened 2020-07-24
Course project action should use notifications instead of dialogs


nikke234:
From Intellij SDK docs:
> One of the leading design principles is avoiding the use of modal message boxes for notifying the user about errors and other situations that may warrant the user’s attention. As a replacement, the IntelliJ Platform provides multiple non-modal notification UI options.

* [ ] **check the consistency of the notification `type`s as well**

OlliKiljunen:
How to: Test different notifications of `CourseProjectAction`:
+ network errors: disable internet connection 1) before firing action, 2) before clicking OK in dialog.
+ malformed course config file: either change course configuration file to be malformed or change code (e.g. make it read property "foo" instead of "modules")
+ local course file error `touch .idea/a-plus-project.json && chmod 000 .idea/a-plus-project.json`

PR #316, opened 2020-08-14
Adjust testing guide

superseacat:
# Description of the PR

+ assignments
+ submissions
+ authentication
+ refresh
+ renumber

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [x] Yes
- [ ] Not relevant


TESTING.md
```diff
@@ -99,24 +99,17 @@ Marked with a github issue label "manual testing."
...
-
-**3.1** ASSERTION: The screen shows a notification telling
-        you are using a pre-release version of the plugin.
-
-**3.2** ASSERTION: The version mentioned in the notification is correct.
-
-![Version notification](images/version-notification.png)
-
-- You can check the version of the plugin in **File > Settings... > Plugins > Installed**
-
```
OlliKiljunen, line 112: Nit: Chapter numbers go now 1, 2, 4, 5, ... but I guess that's okay because otherwise everything should be renumbered.

I'm wondering if we should change the numbering scheme to get rid of this problem.

TESTING.md
```diff
@@ -328,3 +321,73 @@ imported packages: **o1, o1.llama, o1.randomtext**.
...
+
+![Set token](images/setToken.png)
+
+**9.2** Click on the highlighted "What's my token?" link.
+
+**9.3** ASSERTION: Browser opens your profile page in A+ LMS (login, if asked).
+
+**9.4** Copy **API Access Token** from the web page and insert it into the prompt, click "OK".
+
+**9.5** ASSERTION: The assignments tab is populated with Weekly assignments.
```
nikke234, line 337: Nit: "weekly" unnecessarily capitalized? Perhaps
"ASSERTION: The assignments tab lists weeks of the course."

TESTING.md
```diff
@@ -328,3 +321,73 @@ imported packages: **o1, o1.llama, o1.randomtext**.
...
+
+![Assignments shown](images/assignments-shown.png)
+
+**9.6** Close the IDE.
+
+**9.7** Open the IDE with the project.
+ 
+**9.8** If [safe password storage](https://jetbrains.org/intellij/sdk/docs/basics/persisting_sensitive_data.html?search=sensitive#storage) 
+is activated in the OS.
```
OlliKiljunen, line 346: ...then what?

TESTING.md
```diff
@@ -328,3 +321,73 @@ imported packages: **o1, o1.llama, o1.randomtext**.
...
+
+**9.9** ASSERTION: Assignments tab is still populated.
```
nikke234, line 348: Nit: I would combine 9.8 and 9.9, as 9.8 doesn't make much sense on its own

TESTING.md
```diff
@@ -328,3 +321,73 @@ imported packages: **o1, o1.llama, o1.randomtext**.
...
+
+#### Part I: Submitting an Assignment.
+
+**10.1** Expand **Week 1** and select **Assignment 2 (GoodStuff)**, and hit 
+**Submit A+ Assignment**.
+
+![Submit](images/submit_button.png)
+
+**10.2** ASSERTION: The "Could not find module: The A+ Courses plugin could not find the 
+module GoodStuff." warning is shown.
```
nikke234, line 360: Nit: there is a separate notification type for warnings, so this may be confusing since the notification is actually an error notification

TESTING.md
```diff
@@ -328,3 +321,73 @@ imported packages: **o1, o1.llama, o1.randomtext**.
...
+ 
+#### Part II. Checking the submission.
+ 
+**10.8** Wait a while.
+
+**10.9** ASSERTION: The "Submission feedback available Feedback for Assignment 2 (Goodstuff) is now 
+available. Click the link below to open the feedback." notification is shown.
+
+**10.10** ASSERTION: Submission is added under the corresponding Assignment.
+
```
nikke234, line 382: This isn't actually true though? Unless the "Refresh Assignments" action is used

TESTING.md
nikke234: Perhaps get rid of numbers completely and just use bullet points. The beginning of the guide can mention that the steps are to be done in order if that's potentially unclear.

TESTING.md
superseacat: actually a thing that would be nice to-re-check (added as a sub-task for https://github.com/Aalto-LeTech/intellij-plugin/issues/249)

TESTING.md
superseacat: so there's no **push**?

TESTING.md
superseacat: `ctrl + f` & `ctrl + r` are quick an easy way to renumber (had this on todo)

TESTING.md
superseacat: merged as @nikke234 advised

TESTING.md
nikke234: Yep they should be checked. "Error" is definitely correct for the missing module case IMO

TESTING.md
nikke234: By push do you mean that the submission is added to the tree? That doesn't happen at the moment.

After a refresh the new submission comes from the API and becomes visible.

nikke234:
Some tests probably need to be fixed since the UI strings were modified

superseacat:
> Some tests probably need to be fixed since the UI strings were modified

not intended, will fix, but this is why I like raw `String`s in tests

TESTING.md
```diff
@@ -99,295 +99,301 @@ Marked with a github issue label "manual testing."
...
+### 10 Refreshing Modules and Assignments
 
-**10.9** ASSERTION: The "Submission feedback available Feedback for Assignment 2 (Goodstuff) is now 
-available. Click the link below to open the feedback." notification is shown.
+**10.1** Hit the **Refresh Assignments** or **Refresh Modules** button.
+
+![Refresh](images/refresh.png)
+
+**10.2** ASSERTION: Assignments list is collapsed.
+
```
OlliKiljunen, line 381: I think this manual testing doc should reflect the desired/expected (by a user) behavior of the software rather than its actual implementation. This part makes little sense to a user as the list collapsing is more of an inconvenience than what is desired to happen when refreshing. (Of course for us, it's a sign that refresh was actually triggered but only because we know how that's been implemented.)

Instructions like this (that are based on implementation details) can also become easily outdated by some future UI improvements without developers realizing they are breaking manual testing workflow.

OlliKiljunen: So I would change the last assertion to be that the new submission has appeared there and add a note (in parentheses, maybe) that the assignment list may be collapsed when refresh button is pressed, but that's ok.

OlliKiljunen:
> > Some tests probably need to be fixed since the UI strings were modified
> 
> not intended, will fix, but this is why I like raw `String`s in tests

Do you mean that sarcastically? Because I think it would be ideal that UI strings could be re-worded without it breaking tests as long as the idea of the content is the same. So, in manual testing I prefer formats like "Dialog, which asks the user whether or not they want to restart the IDE now, is shown" because that's not dependent on how the thing is actually said word-by-word.

superseacat:
> Do you mean that sarcastically? 

nope, this helped a lot when I had to extract those strings :wink: but maybe it's not needed anymore once we're done

PR #338, opened 2020-08-20
Course project action notifications #249

OlliKiljunen:
# Description of the PR

+ Changed course project action to use notifications on errors (instead of dialogs)

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
@@ -2,6 +2,7 @@ package fi.aalto.cs.apluscourses.intellij.actions;
 
 import static fi.aalto.cs.apluscourses.utils.PluginResourceBundle.getText;
 
+import com.intellij.notification.Notifications;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.application.ApplicationManager;
@@ -9,6 +10,10 @@ import com.intellij.openapi.application.ex.ApplicationEx;
 import com.intellij.openapi.project.Project;
 import fi.aalto.cs.apluscourses.intellij.model.IntelliJModelFactory;
 import fi.aalto.cs.apluscourses.intellij.model.SettingsImporter;
+import fi.aalto.cs.apluscourses.intellij.notifications.CourseConfigurationError;
+import fi.aalto.cs.apluscourses.intellij.notifications.CourseFileError;
+import fi.aalto.cs.apluscourses.intellij.notifications.NetworkErrorNotification;
+import fi.aalto.cs.apluscourses.intellij.notifications.Notifier;
 import fi.aalto.cs.apluscourses.intellij.services.PluginSettings;
 import fi.aalto.cs.apluscourses.intellij.utils.CourseFileManager;
 import fi.aalto.cs.apluscourses.model.ComponentInstaller;
@@ -20,10 +25,15 @@ import fi.aalto.cs.apluscourses.ui.InstallerDialogs;
 import fi.aalto.cs.apluscourses.ui.courseproject.CourseProjectActionDialogs;
 import fi.aalto.cs.apluscourses.ui.courseproject.CourseProjectActionDialogsImpl;
 import fi.aalto.cs.apluscourses.utils.PostponedRunnable;
+import fi.aalto.cs.apluscourses.utils.async.Awaitable;
 import fi.aalto.cs.apluscourses.utils.async.SimpleAsyncTaskManager;
+import fi.aalto.cs.apluscourses.utils.async.TaskManager;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.Objects;
+import java.util.concurrent.Executor;
+import java.util.concurrent.FutureTask;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
@@ -53,6 +63,11 @@ public class CourseProjectAction extends AnAction {
   @NotNull
   private final InstallerDialogs.Factory installerDialogsFactory;
 
+  @NotNull
+  private final Notifier notifier;
+
+  private final Executor executor;
+
   /**
    * Construct a course project action with the given parameters.
    *
@@ -72,6 +87,7 @@ public class CourseProjectAction extends AnAction {
    *                         Since the installation of automatically installed components may take
    *                         quite a while, it is advisable for this to show the user a confirmation
    *                         dialog regarding the restart.
+   * @param notifier         Notification bus.
    */
   public CourseProjectAction(@NotNull CourseFactory courseFactory,
                              boolean createCourseFile,
@@ -79,7 +95,9 @@ public class CourseProjectAction extends AnAction {
                              @NotNull ComponentInstaller.Factory installerFactory,
                              @NotNull PostponedRunnable ideRestarter,
                              @NotNull CourseProjectActionDialogs dialogs,
-                             @NotNull InstallerDialogs.Factory installerDialogsFactory) {
+                             @NotNull InstallerDialogs.Factory installerDialogsFactory,
+                             @NotNull Notifier notifier,
+                             @NotNull Executor executor) {
     this.courseFactory = courseFactory;
     this.createCourseFile = createCourseFile;
     this.settingsImporter = settingsImporter;
@@ -87,6 +105,8 @@ public class CourseProjectAction extends AnAction {
     this.ideRestarter = ideRestarter;
     this.dialogs = dialogs;
     this.installerDialogsFactory = installerDialogsFactory;
+    this.notifier = notifier;
+    this.executor = executor;
   }
 
   /**
@@ -104,6 +124,8 @@ public class CourseProjectAction extends AnAction {
       }
     });
     this.installerDialogsFactory = InstallerDialogs::new;
+    this.notifier = Notifications.Bus::notify;
+    this.executor = runnable -> new Thread(runnable).start();
   }
 
   @Override
@@ -114,7 +136,7 @@ public class CourseProjectAction extends AnAction {
       return;
     }
 
-    URL selectedCourseUrl = getSelectedCourseUrl();
+    URL selectedCourseUrl = getSelectedCourseUrl(project);
     if (selectedCourseUrl == null) {
       return;
     }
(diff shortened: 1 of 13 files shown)


src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
```diff
@@ -130,25 +152,33 @@ public void actionPerformed(@NotNull AnActionEvent e) {
...
+    if (!courseProjectViewModel.userOptsOutOfSettings() && !tryImportIdeSettings(project, course)) {
+      return;
     }
 
     if (createCourseFile) {
       // The course file not created in testing.
       PluginSettings.getInstance().createUpdatingMainViewModel(project);
     }
+
+    executor.execute(() -> {
```
superseacat, line 176: OK, I understand the part with `dialog` goes and `notification` comes, but what's all the fuzz with `awaitable` and **async**?

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
```diff
@@ -130,25 +152,33 @@ public void actionPerformed(@NotNull AnActionEvent e) {
...
+    Awaitable installDone = startAutoInstallsWithRestart(course, project);
+    Objects.requireNonNull(installDone); // because of checkstyle stupidity
 
     if (!tryImportProjectSettings(project, course)) {
       return;
     }
 
-    if (!courseProjectViewModel.userOptsOutOfSettings()) {
-      tryImportIdeSettings(course);
+    if (!courseProjectViewModel.userOptsOutOfSettings() && !tryImportIdeSettings(project, course)) {
```
superseacat, line 167: "remove this explession that always evaluates to `false`" says sonar for `!tryImportIdeSettings(project, course)`

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
superseacat: more I read it more I get a feeling this PR does something extra :stuck_out_tongue_winking_eye:

superseacat (commented):
overall PR is good :rocket: thus I would still love to clarify the "extra value" it adds to the process of auto-installing components.
I suggest we don't rush with it and include it into the next "release" (the one coming next week)

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
OlliKiljunen: In a way... or it is more like that transition to notifications make a certain problem apparent, and that problem I tried to fix here.

So, previously do-you-wanna-restart-dialog was shown after the auto-install modules were (asynchronously) installed *regardless of* whether settings were successfully imported or not. So even if settings import failed, the user was requested to restart. In practice, it wasn't a big issue as error dialog made it clear if something went wrong. With notifications, it's very likely that user wouldn't pay attention to an error notification that is shown at the same time with do-you-wanna-restart-dialog. So user just clicks "yes, restart now" and will never know that there was an error with settings import as restarting also clears notifications.

(I'm not sure if it was even possible that user clicks "do restart" before the import was completed. I guess, it wasn't, though.)

So now, do-you-wanna-restart-dialog is only shown after the settings have successfully imported (otherwise, a notification is shown). This is implemented (not perhaps best way but I tried to keep footprint small) so that after everything has gone well, line 176 spawns a new thread which 1) waits for auto-installment being completed, 2) shows the do-you-wanna-restart-dialog if needed.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
```diff
@@ -258,26 +284,13 @@ private boolean tryImportProjectSettings(@NotNull Project project, @NotNull Cour
...
    * @return True if IDE settings were successfully imported, false otherwise.
    */
-  private boolean tryImportIdeSettings(@NotNull Course course) {
+  private boolean tryImportIdeSettings(@NotNull Project project, @NotNull Course course) {
     try {
       settingsImporter.importIdeSettings(course);
       return true;
     } catch (IOException e) {
-      notifyNetworkError();
+      notifier.notify(new NetworkErrorNotification(e), project);
```
nikke234, line 292: Perhaps also log this one so we have more information if needed?

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
nikke234: > (I'm not sure if it was even possible that user clicks "do restart" before the import was completed. I guess, it wasn't, though.)

I think the action (at least when launched from the A+ menu) runs on the UI thread, so the auto-installs (which use a write action) cannot finish before the action. Otherwise we would have an interesting situation where the user restarts before the action is completed.

However, I'm not sure if there are ways to use the action where it doesn't run on the UI thread, so I think this addition is a good guarantee against that (and of course showing the restart dialog when the IDE settings import failed is a bit silly)

src/main/java/fi/aalto/cs/apluscourses/utils/PostponedRunnable.java
```diff
@@ -2,6 +2,7 @@
 
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.ModalityState;
+import fi.aalto.cs.apluscourses.utils.async.Awaitable;
```
nikke234, line 5: Seems to be an unused import :smile:

src/test/java/fi/aalto/cs/apluscourses/intellij/notifications/CourseFileErrorTest.java
```diff
@@ -0,0 +1,23 @@
...
+import org.junit.Test;
+
+public class CourseFileErrorTest extends TestCase {
+
+  @Test
+  public void testCourseFileError() {
+    String errorMessage = "This is just a test! :D";
+    Exception e = new Exception(errorMessage);
+    Notification notification = new CourseFileError(e);
+    assertEquals(PluginSettings.A_PLUS, notification.getTitle());
```
nikke234, line 19: Probably meant notification group instead of notification title?

src/main/java/fi/aalto/cs/apluscourses/intellij/notifications/CourseFileError.java
```diff
@@ -0,0 +1,23 @@
...
+
+  /**
+   * Error to be shown when course settings file cannot be accessed.
+   *
+   * @param e Exception.
+   */
+  public CourseFileError(Exception e) {
+    super(PluginSettings.A_PLUS,
+        getText("notification.CourseFileError.title"),
+        getAndReplaceText("notification.CourseFileError.content", e.getLocalizedMessage()),
```
nikke234, line 20: I'm curious if there's any specific reason to use `getLocalizedMessage` here? Hasn't been used elsewhere as far as I know

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
```diff
@@ -192,25 +223,19 @@ private Course tryGetCourse(@NotNull Project project, @NotNull URL courseUrl) {
...
-      notifyMalformedCourseConfiguration();
+      notifier.notify(new CourseConfigurationError(e), project);
       return null;
     }
   }
 
-  private void startAutoInstallsWithRestart(@NotNull Course course,
-                                            boolean restartWhenFinished,
-                                            @NotNull Project project) {
+  private Awaitable startAutoInstallsWithRestart(@NotNull Course course, @NotNull Project project) {
```
nikke234, line 235: perhaps rename this method to `startAutoInstalls` after this change?

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
nikke234: Btw is the separate waiting thread in the end needed for this reason, i.e. that the action completes?

nikke234:
> overall PR is good :rocket: thus I would still love to clarify the "extra value" it adds to the process of auto-installing components.
> I suggest we don't rush and include it into the next "release" (the one coming next week)

In theory I agree since this adds a bit more than just the notifications, but in practice this also helps us debug the IDE settings bug if it still appears, so I would very much prefer to merge this first

superseacat:
> > overall PR is good rocket thus I would still love to clarify the "extra value" it adds to the process of auto-installing components.
> > I suggest we don't rush and include it into the next "release" (the one coming next week)
> 
> In theory I agree since this adds a bit more than just the notifications, but in practice this also helps us debug the IDE settings bug if it still appears, so I would very much prefer to merge this first

it's me just being wimpy again :stuck_out_tongue_closed_eyes: 

some points here:
* the merge supposed to happen yesterday
* what supposed to be a low-priority one-liner is not now
* the additional change covers one critical feature of ours (`autoInstall`), that can easily twist all the further processes
* proposed implementation is of "we reason about it, not auto-test"
* we are in no rush yet, so again, I suggest to consider a well-tested feature next week vs. quick-push (it's still getting there after all)

nikke234:
> > > overall PR is good rocket thus I would still love to clarify the "extra value" it adds to the process of auto-installing components.
> > > I suggest we don't rush and include it into the next "release" (the one coming next week)
> > 
> > 
> > In theory I agree since this adds a bit more than just the notifications, but in practice this also helps us debug the IDE settings bug if it still appears, so I would very much prefer to merge this first
> 
> it's me just being wimpy again stuck_out_tongue_closed_eyes
> 
> some points here:
> 
>     * the merge supposed to happen yesterday
> 
>     * what supposed to be a low-priority one-liner is not now
> 
>     * the additional change covers one critical feature of ours (`autoInstall`), that can easily twist all the further processes
> 
>     * proposed implementation is of "we reason about it, not auto-test"
> 
>     * we are in no rush yet, so again, I suggest to consider a well-tested feature next week vs. quick-push (it's still getting there after all)

You are correct, we shouldn't rush this. One option would be to just add a single log statement for the `IOException` for the current release and then leave this for the next week once properly tested

OlliKiljunen:
I strip this from extra functionality so it is notifications only. Gimme 30 min :sweat:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
OlliKiljunen: Sonar is wrong. It doesn't understand `try-catch`es.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
OlliKiljunen: So this was reverted but I comment here:

> I think the action (at least when launched from the A+ menu) runs on the UI thread, so the auto-installs (which use a write action) cannot finish before the action.

This was my reasoning too but as said, it's vague and this guarantee is not clearly visible in the code.

> Btw is the separate waiting thread in the end needed for this reason, i.e. to let the action complete?

We don't want to wait in EDT. Yes, this action makes EDT wait on other parts (was the freezing of UI even reported by some of our testers) and we should get rid of those completely, so we don't want to add more waiting.

src/main/java/fi/aalto/cs/apluscourses/intellij/notifications/CourseFileError.java
OlliKiljunen: No reason. I somehow assumed/remembered we used localized messages in the UI. Changed that.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
OlliKiljunen: I'll keep that in mind for the future.

OlliKiljunen:
> proposed implementation is of "we reason about it, not auto-test"

As it should be. :wink:

nikke234:
> > proposed implementation is of "we reason about it, not auto-test"
> 
> As it should be. :wink:

I agree :laughing:

PR #339, opened 2020-08-21
Update build.gradle

superseacat:
update version for **1.1** release


nikke234 (changes requested):
I suggest we wait for #338 before we merge :+1:

PR #342, opened 2020-08-21
Update README.md

superseacat:
# Description of the PR

+ add code of conduct
+ and a reference to the TESTING.md and the wiki

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,11 @@ The project is developed under [Aalto Le-Tech research group](https://research.c
 
 ## Code of Conduct
 
-The team follows Aalto University general [Code of Conduct](https://www.aalto.fi/en/aalto-university/code-of-conduct).
+The team follows Aalto University general [Code of Conduct](https://www.aalto.fi/en/aalto-university/code-of-conduct) principles.
+
+## Features and further development
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,6 +13,14 @@ This repository hosts the code for [IntelliJ IDEA](https://www.jetbrains.com/ide
 
 The project is developed under [Aalto Le-Tech research group](https://research.cs.aalto.fi/LeTech/) and is steered by **[@jsorva](https://github.com/jsorva)** and **[@oseppala](https://github.com/oseppala)**.
 
+## Code of Conduct
+
+The team follows Aalto University general [Code of Conduct](https://www.aalto.fi/en/aalto-university/code-of-conduct) principles.
```
OlliKiljunen, line 18: Nit: Is there link to this that is available for non-Aalto users?

OlliKiljunen (approved):
:+1:

README.md
nikke234: What do you mean? I think that link works for everyone

nikke234 (approved):
I'm not sure if we need to mention the Aalto CoC in the README, but I don't have a strong opinion either way on this

README.md
OlliKiljunen: Oh sorry, now I see it does. I only assumed it was restricted as it took me to login page but I guess that was because I was previously logged in but the session had timed out. For non-logged-in users, it apparently goes right to the page.

PR #343, opened 2020-08-24
Course project action shows restart dialog only after import is done 

OlliKiljunen:
# Description of the PR

+ I don't like the code much, I'd prefer abstracting threading away from the business logic and provide unit tests another way to join spawned threads than just shutting down the executor. But I tried to keep footprint small for your convenience :D Let's refactor after Sprint 3...

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,3 @@ deploy:
 
 env:
   - CI=true
-
-notifications:
-  slack: $SLACK
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ before_install:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
 jdk:
-  - oraclejdk11
+  - openjdk11
 
 addons:
   sonarcloud:
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
@@ -275,6 +275,7 @@ public class CourseProjectAction extends AnAction {
       settingsImporter.importProjectSettings(project, course);
       return true;
     } catch (IOException e) {
+      logger.error("Failed to import project settings", e);
       notifier.notify(new NetworkErrorNotification(e), project);
       return false;
     }
@@ -291,6 +292,7 @@ public class CourseProjectAction extends AnAction {
       settingsImporter.importIdeSettings(course);
       return true;
     } catch (IOException e) {
+      logger.error("Failed to import IDE settings", e);
       notifier.notify(new NetworkErrorNotification(e), project);
       return false;
     }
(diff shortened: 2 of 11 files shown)


.travis.yml
```diff
@@ -28,4 +28,4 @@ env:
   - CI=true
 
 notifications:
-  slack: $SLACK
\ No newline at end of file
+  slack: $SLACK
```
nikke234, line 31: Could we just remove this, since we're not using it atm? :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
```diff
@@ -145,20 +154,33 @@ public void actionPerformed(@NotNull AnActionEvent e) {
...
+    Future<?> autoInstallDone = executor.submit(() -> startAutoInstalls(course, project));
 
-    if (!courseProjectViewModel.userOptsOutOfSettings() && !tryImportIdeSettings(project, course)) {
-      return;
-    }
+    Future<Boolean> projectSettingsImported =
+        executor.submit(() -> tryImportProjectSettings(project, course));
+
+    Future<Boolean> ideSettingsImported =
+        executor.submit(() -> importIdeSettings && tryImportIdeSettings(project, course));
```
nikke234, line 165: `SettingsImporter` should probably now use read actions where needed as it gets run on a background thread.

Also, I assume you've reasoned through that importing project settings and IDE settings in parallel is safe? (Can't see a reason why it wouldn't be)

.travis.yml
superseacat: not using it because we don't need it or because it has been broken for a while?

superseacat (approved):
it's not that bad :+1:

.travis.yml
nikke234: Mainly because it has been broken, but I don't think getting build notifications in slack is useful anyways

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
OlliKiljunen: > SettingsImporter should probably now use read actions where needed as it gets run on a background thread.

A good point but I didn't find anything that should be write-locked (that is, any PSI, VFS, or project root model things).

> Also, I assume you've reasoned through that importing project settings and IDE settings in parallel is safe?

I don't see any reason why it wouldn't be, either.

nikke234 (approved):
Nice! Thanks for taking the concerns into account :+1:

PR #363, opened 2020-08-31
Version -> 1.3

OlliKiljunen:
Version 1.3 (with one bugfix)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ plugins {
 }
 
 group 'fi.aalto.cs'
-version '1.2'
+version '1.3'
 
 sourceCompatibility = 1.8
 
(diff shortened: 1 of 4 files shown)


nikke234 (approved):

superseacat (approved):
it's good, merge it :runner:

superseacat:
> Version 1.3 (with one bugfix)

the issue (bug) discover was, that IntelliJ IDEA still does not instantly save the contents of the file, that has been edited right now. and so, when submitting to A+, the non-modified original version is being sent. this has now been replaced with a "save all" emulation programmatically. in the future, this might be changed into a fine-graned implementation affection e.g. only the files of the module or even submittable ones. the trigger to redo this might also be performance issues, which are possible but unlikely to happen :hankey:

PR #372, opened 2020-09-03
Log exceptions from exercise tree update

nikke234:


superseacat (changes requested):
are you sure theses log records are visible in log for the local IntelliJ IDEA instance :nerd_face:

superseacat (approved):

PR #378, opened 2020-09-07
Filter exercises

OlliKiljunen:
# Description of the PR

- [x] Unit tests
- [x] Manual testing doc
- [x] UI string handling
- [x] Filtering to be moved into a bg thread (just set the correct executor).

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [x] Yes
- [ ] Not relevant


superseacat:
one proposal is to move the filter to the right (align right), what do you think?

src/main/java/fi/aalto/cs/apluscourses/utils/CollectionUtil.java
```diff
@@ -0,0 +1,54 @@
...
+   * Index-supporting mapping for lists.
+   *
+   * @param list Source list.
+   * @param func Function from source element and its index to result element.
+   * @param startIndex The beginning of indexing.
+   * @param <T> Type of source list elements.
+   * @param <R> Type of result list elements.
+   * @return List of result elements.
+   */
+  public static <T, R> List<R> mapWithIndex(@NotNull List<T> list,
```
superseacat, line 26: we're gonna have kotlin someday :stuck_out_tongue_winking_eye:

superseacat:
"refresh" resets the "filter", intended?

src/main/java/fi/aalto/cs/apluscourses/presentation/base/TreeViewModel.java
```diff
@@ -1,9 +1,9 @@
...
 
 import java.util.List;
-import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.NotNull;
 
 public interface TreeViewModel {
-  @Nullable
-  List<? extends TreeViewModel> getSubtrees();
+  @NotNull
+  List<? extends TreeViewModel> getChildren();
```
superseacat, line 8: without going deeper into the details why `@Nullable` -> `@NotNull`?

OlliKiljunen:
Coincidentally, this now also closes #321

OlliKiljunen:
The main changes that were made when merging `test` to this branch:
- Renamed and moved `TreeViewModel` to `Tree`.
- Added method `traverseAndFind` into `Tree`. That is now used when finding selected submissions/exercises.
- To the needs of that method, added class `ConsList` to support list constructs similar to those in Lisp `(cons first rest)`.
- Changed `Exercise` and `SubmissionResult` classes so that `Exercise` objects are created first and `SubmissionResult` objects are then created and added to the `Exercise`.

src/main/java/fi/aalto/cs/apluscourses/presentation/base/TreeViewModel.java
OlliKiljunen: Previously leaves returned `null`. To streamline things, they now return an empty list.

OlliKiljunen:
> "refresh" resets the "filter", intended?

This will be fixed in the separate upcoming PR where option status is preserved between restarts.

src/main/java/fi/aalto/cs/apluscourses/utils/ConsList.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.List;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Lisp-like list implementation.
+ * Inspired by https://stackoverflow.com/a/2634421
+ *
+ * @param <T> Type of list elements.
+ */
+public class ConsList<T> extends AbstractList<T> {
```
superseacat, line 13: could you please elaborate on what is a benefit of this approach over the simple creation of a new `List` with that desired element as (to be added) as the first one?

src/main/java/fi/aalto/cs/apluscourses/presentation/filter/Filter.java
```diff
@@ -0,0 +1,17 @@
...
+import org.jetbrains.annotations.NotNull;
+
+public interface Filter {
+  /**
+   * Resolves whether or not the filter applies to an item.
+   *
+   * @param item An item to check.
+   * @return True: this filter applies to the item.
+   *         False: this filter could apply to the item, but does not.
+   *         Empty: this filter is not applicable to the item.
```
superseacat, line 13: feels a bit unnatural, maybe `false` and emty should be switched?

src/main/java/fi/aalto/cs/apluscourses/utils/ConsList.java
OlliKiljunen: You mean like creating a new `ArrayList` or `LinkedList` or other standard implementations? The problem with those is that a creating a new list based on an old one requires all the elements to be copied from the old to the new (in case of `ArrayList`, even extra copying is needed to insert an element in the beginning of the list).

This class is used in recursive computation where each step produces a new list with a new item in the beginning of the list. If we used, say Java's `LinkedList`, we would be copying items again and again from a temporary list to another which would lead to quite a rampant and difficult-to-predict time complexity (possibly memory complexity too) .

It's quite surprising that Java's standard library doesn't have a class like this ready-made. On the other hand, the code that uses this, is not very Java-like but following the patterns of functional programming languages such as Lisp. It could be refactored to carry out the computation in more Java-like (imperative) manner but it would then be more difficult to understand it and reason about its correctness.

src/main/java/fi/aalto/cs/apluscourses/presentation/filter/Filter.java
OlliKiljunen: The explanation is very poor but I guess the values are chosen right. What this means: if our filter is say "strings that end with a period":

`"hello."` -> true
`"hi!"` -> false
`0.123` -> empty

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/FilterOptionAction.java
```diff
@@ -0,0 +1,30 @@
+package fi.aalto.cs.apluscourses.intellij.actions;
+
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.actionSystem.ToggleAction;
+import fi.aalto.cs.apluscourses.presentation.filter.Option;
+import org.jetbrains.annotations.NotNull;
+
+public class FilterOptionAction extends ToggleAction {
```
superseacat, line 8: some javadoc/comment here on somewhere explaining the overall construct?

src/main/java/fi/aalto/cs/apluscourses/presentation/filter/LogicalFilter.java
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  @Override
+  @NotNull
+  public Optional<Boolean> apply(Object item) {
+    return filters.stream()
+        .map(filter -> filter.apply(item))
+        .filter(Optional::isPresent)
+        .map(Optional::get)
+        .reduce(operator);
```
superseacat, line 24: abstract stuff is a bit hard to read, but guess there's no other way, or... maybe a comment? :open_mouth:

src/main/java/fi/aalto/cs/apluscourses/presentation/filter/Filter.java
```diff
@@ -0,0 +1,17 @@
+package fi.aalto.cs.apluscourses.presentation.filter;
+
+import java.util.Optional;
+import org.jetbrains.annotations.NotNull;
+
+public interface Filter {
```
superseacat, line 6: I'm a bit lost here...

Screenshot from 2020-09-15 18:34:57

src/main/java/fi/aalto/cs/apluscourses/presentation/filter/Option.java
```diff
@@ -0,0 +1,48 @@
...
+  private final String name;
+  @NotNull
+  private final Filter filter;
+  @Nullable
+  private final Icon icon;
+
+  public final ObservableProperty<Boolean> isSelected = new ObservableReadWriteProperty<>(true);
+
+  /**
+   * An option, that is, a filter that when not selected, filters out those that match to the filter
```
superseacat, line 22: arrrrgh... ! :pirate_flag:

src/main/java/fi/aalto/cs/apluscourses/presentation/filter/Option.java
```diff
@@ -0,0 +1,48 @@
+package fi.aalto.cs.apluscourses.presentation.filter;
+
+import fi.aalto.cs.apluscourses.utils.observable.ObservableProperty;
+import fi.aalto.cs.apluscourses.utils.observable.ObservableReadWriteProperty;
+import java.util.Optional;
+import javax.swing.Icon;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class Option implements Filter {
```
superseacat, line 10: `Optional<Option>` :stuck_out_tongue_closed_eyes:

src/main/java/fi/aalto/cs/apluscourses/utils/Tree.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.function.Predicate;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public interface Tree {
+  @NotNull
+  List<? extends Tree> getChildren();
+
+  /**
+   * Traverses the tree in depth-first pre-order and finds the first node that satisfies the
```
superseacat, line 15: what does `depth-first pre-order` mean? :nerd_face:

superseacat (approved):
this looks nice, approving on the condition @OlliKiljunen promises to have a small guide-through session for me on this :+1:

src/main/java/fi/aalto/cs/apluscourses/utils/ConsList.java
superseacat: alright, your answer helped me to (find more details) understand! :bow:

src/main/java/fi/aalto/cs/apluscourses/presentation/filter/Filter.java
OlliKiljunen: `Filter` is like Java's `Predicate<T>` (with the exception that it returns true/false/empty as discussed elsewhere).

`TypedFilter<T>` is an abstract class that can be extended to implement such filters that return empty if the item is not type of `T` and otherwise true/false based on logic written into subclass.

`Option` is what we use for these kind of filtering options: "show non-submittable exercises", "show exercises with submission count exceeded" etc. They are given as a constructor parameter a filter that defines those target objects (e.g. for "show non-submittable exercises" option is given such a filter that returns true for non-submittable exercises, false for other exercises and empty for non-exercises). Option has property `isSelected`. If it's true, it always returns empty (that is, it doesn't make any restrictions to what should be shown). If `isSelected` is false, it returns false for its target objects and otherwise true/empty. E.g. when "show non-submittable exercises" option is not selected, it hides non-submittable exercises (by returning false for them).

`LogicalFilter` is just a class I wrote to be base for "and-filter" (`LogicalFilter.Conjunction`) and "or-filter" (turned out there was no need for it), that is filters that compose multiple filters and applies a logical operator (AND/OR) to them. As there's only one such (AND i.e. conjunction) currently in use, the whole class could be replace with more specific `AndFilter` class (which could also has proper short-circuiting rules to improve performance).

`Options` is a collection of `Option` objects and applies AND-logic to them (thus inheriting `LogicalFilter.Conjunction`). An `Options` object is what we apply to the exercise tree and it makes sure that each of its `Option`s is applied to the tree elements.

src/main/java/fi/aalto/cs/apluscourses/utils/Tree.java
OlliKiljunen:
```
    A
   / \
  B   C
 / \
D   E
```
Depth first search with pre-order for this tree is: A, B, D, E, C.
Compare this to DFS with post-order (where parent nodes are visited after their children): D, E, B, C, A.
There's also breadth first search: A, B, C, D, E.

src/main/java/fi/aalto/cs/apluscourses/presentation/filter/Filter.java
nikke234: In terms of filtering, what is the difference between an item "passing" the filter or the filter not being applicable to the item?

nikke234:
> Changed Exercise and SubmissionResult classes so that Exercise objects are created first and SubmissionResult objects are then created and added to the Exercise.

Would you mind explaining the reasoning behind this?

src/main/java/fi/aalto/cs/apluscourses/utils/CollectionUtil.java
```diff
@@ -0,0 +1,59 @@
...
+   * @param iterator An iterator.
+   * @param itemToFind A item that is wanted to be found.
+   * @param <T> Type of items.
+   * @return The index of the item.
+   */
+  public static <T> int indexOf(@NotNull Iterator<T> iterator, @Nullable T itemToFind) {
+    int index = 0;
+    while (iterator.hasNext()) {
+      T current = iterator.next();
+      if (current == itemToFind) {
```
nikke234, line 52: Wouldn't equals be more appropriate here? (taking care of the possibility that `current` may be null of course

src/main/java/fi/aalto/cs/apluscourses/presentation/filter/Filter.java
nikke234: Replacing `LogicalFilter` with the more concrete `AndFilter` has my strong support. Our code already has plenty of base classes which aren't really necessary :smile:

src/main/java/fi/aalto/cs/apluscourses/utils/ConsList.java
nikke234: A cons list has horrible performance when iterating (compared to `ArrayList`), but the recursive solution in `traverseAndFind` with cons list is very neat indeed

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/FilterOptionsActionGroup.java
```diff
@@ -0,0 +1,40 @@
...
+
+  // Caution: Array covariance in the return type.
+  @Override
+  public @NotNull FilterOptionAction[] getChildren(@Nullable AnActionEvent e) {
+    return Optional.ofNullable(e)
+        .map(AnActionEvent::getProject)
+        .map(mainViewModelProvider::getMainViewModel)
+        .map(MainViewModel::getExercises)
+        .map(BaseTreeViewModel::getFilterOptions)
+        .map(Streamable::stream)
```
nikke234, line 35: :joy:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/FilterOptionsActionGroup.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public FilterOptionsActionGroup() {
+    this(PluginSettings.getInstance());
+  }
+
+  public FilterOptionsActionGroup(MainViewModelProvider mainViewModelProvider) {
+    this.mainViewModelProvider = mainViewModelProvider;
+  }
+
+  // Caution: Array covariance in the return type.
```
nikke234, line 27: What's the array covariance here? Doesn't the method create an array `FilterOptionAction[]`?

src/main/java/fi/aalto/cs/apluscourses/utils/ConsList.java
OlliKiljunen: Right. It should have similar performance in iteration and random-access as `LinkedList` (well, it is a linked list). Which is of course poor compared to arrays.

src/main/java/fi/aalto/cs/apluscourses/utils/CollectionUtil.java
OlliKiljunen: Yes, that makes sense, I'll change it. In case where this is used, it doesn't matter.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/FilterOptionsActionGroup.java
OlliKiljunen: Yes it does. What I meant was that `FilterOptionsActionGroup` overrides a method returning `AnAction[]` with a method returning `FilterOptionAction[]` (and why is that? for convenience in testing). In Java, that is allowed, and it shouldn't cause any problems here because the caller is not expected to modify the array.

OlliKiljunen:
> 
> 
> > Changed Exercise and SubmissionResult classes so that Exercise objects are created first and SubmissionResult objects are then created and added to the Exercise.
> 
> Would you mind explaining the reasoning behind this?

Sure. I needed `SubmissionResult` objects to keep reference to the `Exercise`. Instead of adding it after the `SubmissionResult` object have been instantiated, I thought it makes more sense that `SubmissionResult`s are created knowing their parent but added to the parent after being created. (If we ever need to change to a design where new `SubmissionResult` objects are added to the tree without the whole tree being rebuilt, this would be the natural order of instructions).

PR #379, opened 2020-09-07
Troubleshooting

superseacat:
# to be sent to all the TAs to combat facing issues

+ first part is meant to validate their installation (please read carefully, aim is to be concise but rich)
+ second part is questionable: on one hand, it can make students think there are plenty of bugs or mislead them (say confusing one bug for another one and therefore complicate the troubleshooting), on the other hand, we could get more feedback on bugs' occurrence and potential solutions


TROUBLESHOOTING.md
```diff
@@ -0,0 +1,62 @@
...
+
+**0.** ensure the latest compatible version of IntelliJ IDEA Community Edition is installed (**2020.2.** or later).
+
+**1.** update plugin to the latest version (you can check the latest version on [Plugin's official page](https://plugins.jetbrains.com/plugin/13634-a-courses/versions)).
+
+**2.** create a new project: **File -> New -> Project...** and then **Empty Project**.
+
+**3.** Turn project into an A+ Project: **A+ -> Turn Project Into A+ Project** (and choose the rest of the settings asked for in the pop-up dialog).
+
+**4.** check the IDE for not having errors (you **should not** see the blinking red icon as shown). If there are errors, please share the "stacktrace" text with the course stuff.
```
nkroger, line 14: typo: "stuff"

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,62 @@
...
+|-----|-----------|----------------------------------------------------------------------------------------------------|
+| 1.  | **problem**   | [~~IDE complaining about not able to see the SDK (whitespaces and special characters in naming (path))~~](https://github.com/Aalto-LeTech/intellij-plugin/issues/360)|
+|     | **status**    | solved                                                                                             |
+|     | **solution**  | update to the latest plugin version                                                                |
+| 2.  | **problem**   | [~~assignments are not visible on Aalto Linux (VDI)~~](https://github.com/Aalto-LeTech/intellij-plugin/issues/371)                                                   |
+|     | **status**    | solved                                                                      |
+|     | **solution**  | it's an environment configuration issue and was solved by Aalto IT                                               |
+| 3.  | **problem**   | [continuously being asked to re-enter the A+ token]()                                                  |
```
OlliKiljunen, line 47: Missing link?

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,62 @@
...
+|     | **status**    | not possible to solve due to the configuration of Aalto Linux on VDI                               |
+|     | **solution**  | none                                                                                               |
+| 4.  | **problem**   | [keymap imported by the plugin interferes with macOS's normal keyboard shortcuts]()                    |
```
OlliKiljunen, line 50: Missing link here too?

OlliKiljunen (approved):
Two links are missing (as far as I understand) but as those are not that important, I don't mind. Very nice start for this document which hopefully stays up to date. Very useful for the project! 👍

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Verifying the setup
+
+**0.** ensure the latest compatible version of IntelliJ IDEA Community Edition is installed (**2020.2.** or later).
+
+**1.** update plugin to the latest version (you can check the latest version on [Plugin's official page](https://plugins.jetbrains.com/plugin/13634-a-courses/versions)), but at least **1.4**.
+
+**2.** create a new project: [**File -> New -> Project...** and then **Empty Project**](https://plus.cs.aalto.fi/o1/2020/w01/ch02/#launch-intellij).
+
+**3.** Turn project into an A+ Project: [**A+ -> Turn Project Into A+ Project**](https://plus.cs.aalto.fi/o1/2020/w01/ch02/#the-a-courses-plugin) (and choose the rest of the settings asked for in the pop-up dialog).
```
nikke234, line 12: Is this meant to be done always? I think it would be good to first try to fix the issues in the existing project

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,70 @@
...
+![run](images/Run.png)
+
+**7.** start REPL on a **GoodStuff** module by right-clicking **Scala REPL...**
+![repl](images/REPL.png)
+
+### [Known issues](https://github.com/Aalto-LeTech/intellij-plugin/labels/user-bug)
+
+| #   | status    | description                                                                                        |
+|-----|-----------|----------------------------------------------------------------------------------------------------|
+| 1.  | **problem**   | [~~IDE complaining about not able to see the SDK (whitespaces and special characters in naming (path))~~](https://github.com/Aalto-LeTech/intellij-plugin/issues/360)|
```
nikke234, line 41: Isn't the actual error something like 
> scalac: Scala compiler JARs not found (module: '<module name>')

Also, since we have "status", let's not use strikethrough here since it's a bit difficult to read

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,70 @@
...
+| 5.  | **problem**   | [~~yet another NPE in REPL~~](https://github.com/Aalto-LeTech/intellij-plugin/pull/380)            |
+|     | **status**    | solved                                                                                             |
+|     | **affected**  | :angry: < 10                                                                                       |
+|     | **solution**  | since Plugin version **1.5**                                                                       |
+| 6.  | **problem**   | [network error about A+ token containing whitespaces](https://github.com/Aalto-LeTech/intellij-plugin/issues/377)                                                |
+|     | **status**    | under investigation                                                                                |
+|     | **affected**  | :angry: < 10                                                                                       |
+|     | **solution**  | none                                                                                               |
```
nikke234, line 62: "submit from the web browser for now"?

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,70 @@
...
+
+**Modules**
+![modules 3](images/ts_modules_3.png)
+![modules 1](images/ts_modules_1.png)
+![modules 2](images/ts_modules_2.png)
+
+**Libraries (Scala SDK)**
+![libraries](images/ts_libs.png)
+
+**6.** download a **GoodStuff** module and run it by right-clicking and choosing **Run**
```
nikke234, line 31: Nit: "download a" -> download the"

superseacat: I see you take "proof-reading" seriously, thanks, fixed.

nikke234:
I would also add the case where the blinking comes up when pressing OK on the course project action dialog and the stacktrace has something about the course file missing language or being malformed in another way. The only fix (this fix doesn't always help since the error is pretty random) is to invalidate caches and restart and try again with a fresh project

TROUBLESHOOTING.md
superseacat: this is our analogy to the "please restart your computer", so yes, I'd like this to be here. there is a myriad of possibilities why their initial setup went sour, missed a step, some network issues or somesuch (that you'll never get to reproduce). whereas, running the script through verifies, that the whole thing can be run in that particular environment. once done, then it would make sense to a) use this new project as a base for the course, b) if curious/have time - research what was the cause of trouble

superseacat:
> I would also add the case where the blinking comes up when pressing OK on the course project action dialog and the stacktrace has something about the course file missing language or being malformed in another way. The only fix (this fix doesn't always help since the error is pretty random) is to invalidate caches and restart and try again with a fresh project

the name is just different because the issue appeared quite a while ago, but suggesting invalidate and restart, should it come on that same project or creating a new one is safeR?

nikke234:
> > I would also add the case where the blinking comes up when pressing OK on the course project action dialog and the stacktrace has something about the course file missing language or being malformed in another way. The only fix (this fix doesn't always help since the error is pretty random) is to invalidate caches and restart and try again with a fresh project
> 
> the name is just different because the issue appeared quite a while ago, but suggesting invalidate and restart, should it come on that same project or creating a new one is safeR?

Since this happens immediately after the "Turn Project Into A+ Course Project", there's nothing valuable in the project yet, so I would create a new one just to be safe

PR #380, opened 2020-09-08
Fix REPL null pointer exception

nikke234:
# Description of the PR

The issue is caused by the fact, that `getModuleOfEditorFile` could actually return `Some(null)` in certain situations due to an oversight of mine. This leads to a situation where `Some(null).filter(hasScalaSdk)` is evaluated, which causes an error since `hasScalaSdk` has a `@NotNull` annotation on its module parameter. This PR fixes both `getModuleOfEditorFile` and `getModuleOfSelectedFile` to always either return `None` or `Some(module)`, but never `Some(null)`.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
+++ b/src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
@@ -27,15 +27,15 @@ object ModuleUtils {
     FileUtilRt.toSystemIndependentName(ModuleUtilCore.getModuleDirPath(module))
 
   def getModuleOfEditorFile(@NotNull project: Project,
-                            @NotNull dataContext: DataContext): Option[Module] = for {
-    editor <- Option(CommonDataKeys.EDITOR.getData(dataContext))
-    openFile <- Option(FileDocumentManager.getInstance.getFile(editor.getDocument))
-  } yield ModuleUtilCore.findModuleForFile(openFile, project)
+                            @NotNull dataContext: DataContext): Option[Module] =
+    Option(CommonDataKeys.EDITOR.getData(dataContext))
+        .flatMap(editor => Option(FileDocumentManager.getInstance.getFile(editor.getDocument)))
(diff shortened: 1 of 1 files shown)


OlliKiljunen (approved):
Nice catch. I think I understand the Scala code. which looks correct now. 👍

src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
```diff
@@ -27,15 +27,15 @@ object ModuleUtils {
     FileUtilRt.toSystemIndependentName(ModuleUtilCore.getModuleDirPath(module))
 
   def getModuleOfEditorFile(@NotNull project: Project,
-                            @NotNull dataContext: DataContext): Option[Module] = for {
-    editor <- Option(CommonDataKeys.EDITOR.getData(dataContext))
-    openFile <- Option(FileDocumentManager.getInstance.getFile(editor.getDocument))
-  } yield ModuleUtilCore.findModuleForFile(openFile, project)
+                            @NotNull dataContext: DataContext): Option[Module] =
```
superseacat, line 30: a quick test case to ensure the behavior? :nerd_face:

superseacat (approved):

src/main/scala/fi/aalto/cs/apluscourses/intellij/utils/ModuleUtils.scala
nikke234: That's unfortunately very difficult here due to the calls to `FileDocumenttManager.getInstance.getFile` and `ModuleUtilCore.findModuleForFile`

superseacat: would I otherwise ask? :wink:

PR #390, opened 2020-09-15
Course file fix

nikke234:
# Description of the PR

This PR uses the project path as the key of a map instead of the project itself. It also prevents a missing `downloadedAt` or `language` in the course file from causing an uncaught exception.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
@@ -65,15 +65,46 @@ public class PluginSettings implements MainViewModelProvider {
   private final PropertiesComponent applicationPropertiesManager = PropertiesComponent
       .getInstance();
 
-  @NotNull
-  private final ConcurrentMap<Project, MainViewModel> mainViewModels = new ConcurrentHashMap<>();
+  private static class ProjectKey {
+    @NotNull
+    public final Project project;
+
+    public ProjectKey(@Nullable Project project) {
+      if (project == null) {
+        this.project = ProjectManager.getInstance().getDefaultProject();
+      } else {
+        this.project = project;
+      }
+    }
+
+    @Override
+    public boolean equals(Object other) {
+      if (other == null || !(other instanceof ProjectKey)) {
+        return false;
+      }
+      return getPath().equals(((ProjectKey) other).getPath());
+    }
+
+    @Override
+    public int hashCode() {
+      return getPath().hashCode();
+    }
+
+    @NotNull
+    private String getPath() {
+      return project.isDefault() ? "" : project.getBasePath();
+    }
+  }
 
   @NotNull
-  private final ConcurrentMap<Project, MainViewModelUpdater> mainViewModelUpdaters
+  private final ConcurrentMap<ProjectKey, MainViewModel> mainViewModels = new ConcurrentHashMap<>();
+
+  @NotNull
+  private final ConcurrentMap<ProjectKey, MainViewModelUpdater> mainViewModelUpdaters
       = new ConcurrentHashMap<>();
 
   @NotNull
-  private final ConcurrentMap<Project, CourseFileManager> courseFileManagers
+  private final ConcurrentMap<ProjectKey, CourseFileManager> courseFileManagers
       = new ConcurrentHashMap<>();
 
   @NotNull
@@ -86,12 +117,13 @@ public class PluginSettings implements MainViewModelProvider {
   private final ProjectManagerListener projectManagerListener = new ProjectManagerListener() {
     @Override
     public void projectClosed(@NotNull Project project) {
-      courseFileManagers.remove(project);
-      MainViewModelUpdater updater = mainViewModelUpdaters.remove(project);
+      ProjectKey key = new ProjectKey(project);
+      courseFileManagers.remove(key);
+      MainViewModelUpdater updater = mainViewModelUpdaters.remove(key);
       if (updater != null) {
         updater.interrupt();
       }
-      MainViewModel mainViewModel = mainViewModels.remove(project);
+      MainViewModel mainViewModel = mainViewModels.remove(key);
       if (mainViewModel != null) {
         mainViewModel.dispose();
       }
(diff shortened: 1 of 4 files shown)


src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
```diff
@@ -58,26 +58,58 @@ public String getName() {
...
+    public int hashCode() {
+      return getPath().hashCode();
+    }
+
+    @NotNull
+    private String getPath() {
+      return project.isDefault() ? "" : project.getBasePath();
+    }
+  }
+
```
OlliKiljunen, line 91: I like this approach a lot. I'm only uncertain if it's fine to a) keep references to possibly disposed `Project` objects (as hash map keys), b) call `isDefault()` and `getBasePath()` methods on them (hash map calls `equals` on it's stored keys). So would it be safer to store just the string returned by `getBasePath`, not the `Project` object itself, what do you think? But it's very good design that the implementation is encapsulated inside a class nonetheless.

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/CourseFileManager.java
```diff
@@ -192,9 +203,18 @@ private void loadFromJsonObject(@NotNull JSONObject jsonObject) throws IOExcepti
...
+
       String moduleId = moduleObject.getString(MODULE_ID_KEY);
-      ZonedDateTime downloadedAt
-          = ZonedDateTime.parse(moduleObject.getString(MODULE_DOWNLOADED_AT_KEY));
+
+      String downloadedAtStr = moduleObject.optString(MODULE_DOWNLOADED_AT_KEY);
+      ZonedDateTime downloadedAt;
+      if (downloadedAtStr == null || downloadedAtStr.trim().isEmpty()) {
+        logger.error("Module " + moduleName + " missing 'downloadedAt' in course file");
+        downloadedAt = Instant.EPOCH.atZone(ZoneId.systemDefault());
```
OlliKiljunen, line 213: Would it be safer to default to a distant day in the past? Because we use this information to test whether or not files have been changed since the module was downloaded, false positives (asking user if they want to overwrite even though there are no changes) cause less harm than false negatives (not asking user when there actually are changes).

OlliKiljunen (commented):
Quick and well done. A couple of questions I'd like to hear your thoughts of before approving.

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/CourseFileManager.java
nikke234: `Instant.EPOCH` is a constant for `1970-01-01T00:00:00Z`, so it's pretty distant in the past :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
nikke234: There is a `projectClosed` listener which removes the entries from the maps, but I agree that it makes more sense to just store the base path (assuming that we don't access the project from the key, which I'll check).

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/CourseFileManager.java
OlliKiljunen: Oh sorry, you're right. 🤦‍♂️  My mind only read "Instant" and automatically thought it means like "instantly now". I was almost to suggest the epoch day to be chosen as default... 😛

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
```diff
@@ -58,26 +58,58 @@ public String getName() {
...
+      if (project == null) {
+        this.project = ProjectManager.getInstance().getDefaultProject();
+      } else {
+        this.project = project;
+      }
+    }
+
+    @Override
+    public boolean equals(Object other) {
+      if (other == null || !(other instanceof ProjectKey)) {
```
superseacat, line 75: IJ: "remove `other == null`"

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/CourseFileManager.java
superseacat, line 213: `ZonedDateTime.now()`?

superseacat (approved):
:fireworks:

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/CourseFileManager.java
OlliKiljunen: This is same misunderstanding that I had: https://github.com/Aalto-LeTech/intellij-plugin/pull/390#discussion_r488571846

`downloadedAt` is NOT set to the current date (and it shouldn't). Instant.EPOCH refers to `1970-01-01`, that is "the instant of epoch", not "this instant second in epoch time" as both of us, apparently, read it. :D

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
nikke234: Ah indeed, fixed!

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
nikke234: Fixed :+1:

OlliKiljunen (approved):
Great! 👍 

I realized the next step will be replacing `Project`s with `ProjectKey`s pretty much everywhere in our code (well, then `ProjectKey` will no loger be just a key for a project but our proxy object or something for an IntelliJ project, so it will be renamed too). But it is definitely out of the scope of this PR.

Go and merge! :shipit:

PR #391, opened 2020-09-16
Specified user-agent header of http requests

### LINKED ISSUE #355: some extra payload for A+ API calls
as proposed by @raphendyr and discussed [here](https://apluslms.slack.com/archives/CQ92SAHNY/p1598430777065200) it would make sense to add some extra data for the calls made for A+ API by the plugin, looking something like this: `intellij/123 o1-plugin/0.4 aalto-o1/0.1 Java/1.10.0`

OlliKiljunen:
# Description of the PR

Closes #355 

Example:
```
User-Agent: intellij/2020.2 (IDEA; Community Edition; Windows 10) apluscourses/1.5 Apache-HttpClient/4.5.12 (Java/11.0.7)
```

This is used in API requests to
- A+
- grader.cs.hut.fi when getting the course config file

This user-agent string is not used when
- ~downloading modules/libraries/settings~ (it is!)
- opening browser (for API token, feedback, etc.)

Having information related to course there too (as suggested [here](https://github.com/Aalto-LeTech/intellij-plugin/issues/355#issuecomment-680832876)) would be more complicated thing to implement. I also don't think it should be part of user-agent string. In server side, the course can be easily inferred from the URL.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
@@ -9,6 +9,7 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URL;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import org.apache.commons.io.FileUtils;
 import org.apache.http.HttpEntity;
@@ -40,6 +41,8 @@ import org.json.JSONTokener;
  * response is needed.
  */
 public class CoursesClient {
+
+  private static final AtomicReference<String> userAgent = new AtomicReference<>(null);
 
   /**
    * Makes a GET request to the given URL and returns the response body in a
@@ -300,6 +303,11 @@ public class CoursesClient {
 
   @NotNull
   private static String getUserAgent() {
+    String userAgentString = userAgent.get();
+    if (userAgentString != null) {
+      return userAgentString;
+    }
+
     ApplicationInfo appInfo = ApplicationInfo.getInstance();
     ApplicationNamesInfo appNamesInfo = ApplicationNamesInfo.getInstance();
     String intellijVersion = appInfo.getFullVersion();
@@ -309,7 +317,8 @@ public class CoursesClient {
     String pluginVersion = BuildInfo.INSTANCE.version.toString();
     String apacheUserAgent =
         VersionInfo.getUserAgent("Apache-HttpClient", "org.apache.http.client", VersionInfo.class);
-    return String.format(
+
+    userAgentString = String.format(
         "intellij/%s (%s; %s; %s) apluscourses/%s %s",
         intellijVersion,
         product,
@@ -317,5 +326,7 @@ public class CoursesClient {
         os,
         pluginVersion,
         apacheUserAgent);
+
+    return userAgent.compareAndSet(null, userAgentString) ? userAgentString : userAgent.get();
   }
 }
--- a/src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
@@ -1,5 +1,7 @@
 package fi.aalto.cs.apluscourses.utils;
 
+import com.intellij.openapi.application.ApplicationInfo;
+import com.intellij.openapi.application.ApplicationNamesInfo;
 import fi.aalto.cs.apluscourses.model.InvalidAuthenticationException;
 import fi.aalto.cs.apluscourses.model.UnexpectedResponseException;
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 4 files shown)


src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
```diff
@@ -289,4 +292,30 @@ private static void requireResponseEntity(@NotNull HttpResponse response)
       throw new UnexpectedResponseException(response, "Response is missing body");
     }
   }
+
+  @NotNull
+  private static CloseableHttpClient createClient() {
+    return HttpClients.custom().setUserAgent(getUserAgent()).build();
```
nikke234, line 298: I wonder if `custom` also changes other things from `createDefault`, such as timeouts or something else

src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
```diff
@@ -289,4 +292,30 @@ private static void requireResponseEntity(@NotNull HttpResponse response)
...
+    String apacheUserAgent =
+        VersionInfo.getUserAgent("Apache-HttpClient", "org.apache.http.client", VersionInfo.class);
+    return String.format(
+        "intellij/%s (%s; %s; %s) apluscourses/%s %s",
+        intellijVersion,
+        product,
+        edition,
+        os,
+        pluginVersion,
+        apacheUserAgent);
```
nikke234, line 319: I assume this is all fail-safe if any of those return null?

nikke234: Also just a minor thing: all of this information is now fetched for every request, but it shouldn't change during the runtime of the application

src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
nikke234: An alternative would be to add the user agent to the request instead of the client

nikke234:
> This user-agent string is not used when downloading modules

I think it is actually? And I think it should as well

OlliKiljunen:
> 
> 
> > This user-agent string is not used when downloading modules
> 
> I think it is actually? And I think it should as well

Wait, you are correct... again! :D I didn't remember how downloading modules works but yeah, it uses the same class.

src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
OlliKiljunen: `String.format("Hello %s!", null)` returns `"Hello null!", so it would leave "null"s to the user agent but not fail.

Yes, the whole user agent string could be computed only once (perhaps in a lazy way?), unless we add course information there (which would need also more changes). If we are sure we are not gonna need course info in user-agent, I'll make it a static lazy-initialized string.

src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
OlliKiljunen: Correct. I was wondering the same thing when coding this; I couldn't find no clear documentation but look at the decompiled code reveals they are the same thing:

```java
public class HttpClients {
  // ...
  public static HttpClientBuilder custom() {
    return HttpClientBuilder.create();
  }

  public static CloseableHttpClient createDefault() {
    return HttpClientBuilder.create().build();
  }
  // ...
}

src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
nikke234: We can switch it again if want to add course info, so let's go with static lazy-initialized for now then

src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
nikke234: Nice find! It's good as is in that case

superseacat:
> ```Java/11.0.7 ```

do you think there could be a benefit in having the name of the exact SDK (Adopt... )?

superseacat (approved):
good to go

OlliKiljunen:
> 
> 
> > `Java/11.0.7 `
> 
> do you think there could be a benefit in having the name of the exact SDK (Adopt... )?

This part of the user agent comes straight from the Apache library so I'd say it's best to stick with the info they provide.

PR #397, opened 2020-09-24
Filter for completed assignments

Atlanex:
# Description of the PR
Adds filter, which hides assignments with maximum user points

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [x] Yes
- [ ] Not relevant

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -400,7 +400,7 @@ available. Click here to open the feedback.** notification is shown.
 
 **12.1** Click **Filter Assignments** button (pink funnel icon) above the assignments tree.
 
-**12.2** ASSERTION: **Non-submittable** menu item is checked.
+**12.2** ASSERTION: **Non-submittable** and **Completed** menu items are checked.
 
 **12.3** Uncheck **Non-submittable** menu item.  Note: This makes the exercise tree to collapse.
 
@@ -423,3 +423,5 @@ available. Click here to open the feedback.** notification is shown.
 **12.12** ASSERTION: Non-submittable assignments are shown again in the tree.
 
 **12.12** ASSERTION: **Non-submittable** menu item is checked.
+
+**12.13** Repeat 12.3-12.12 for the **Completed** filter
--- a/TESTING.md
+++ b/TESTING.md
@@ -400,7 +400,7 @@ available. Click here to open the feedback.** notification is shown.
 
 **12.1** Click **Filter Assignments** button (pink funnel icon) above the assignments tree.
 
-**12.2** ASSERTION: **Non-submittable** and **Completed** menu items are checked.
+**12.2** ASSERTION: **Non-submittable** menu item is checked.
 
 **12.3** Uncheck **Non-submittable** menu item.  Note: This makes the exercise tree to collapse.
 
@@ -423,5 +423,3 @@ available. Click here to open the feedback.** notification is shown.
 **12.12** ASSERTION: Non-submittable assignments are shown again in the tree.
 
 **12.12** ASSERTION: **Non-submittable** menu item is checked.
-
-**12.13** Repeat 12.3-12.12 for the **Completed** filter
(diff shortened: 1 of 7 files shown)


src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
```diff
@@ -31,6 +31,15 @@ public boolean isSubmittable() {
...
   }
 
+  /**
+   * Returns {@code true} if the user has gotten maximum points, {@code false} otherwise.
+   */
+  public boolean isCompleted() {
+    int maxPoints = getModel().getMaxPoints();
+    int userPoints = getModel().getUserPoints();
+    return userPoints == maxPoints;
+  }
```
nikke234, line 41: Does this mark all optional practice assignments as completed, since they have 0 max points? Perhaps an additional condition that the exercise has at least one submission

TESTING.md
```diff
@@ -400,7 +400,7 @@ available. Click here to open the feedback.** notification is shown.
 
 **12.1** Click **Filter Assignments** button (pink funnel icon) above the assignments tree.
 
-**12.2** ASSERTION: **Non-submittable** menu item is checked.
+**12.2** ASSERTION: **Non-submittable** and **Completed** menu items are checked.
```
OlliKiljunen, line 403: This can be a bit difficult to test because tester doesn't necessarily have any completed assignments. Thus, I'd leave this away from the TESTING.md

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
OlliKiljunen: Furthermore, this logic would suit better in the model class (`Exercise`) because it's not about presentation but a property of the domain object (an exercise/assignment) itself. (Yeah, I know, we haven't been completely consistent in these matters ourselves. :smile:)

OlliKiljunen: And because @superseacat is on vacation, I say this on his behalf: Could we have unit tests? :laughing:

OlliKiljunen (approved):
@nikke234 raised a good point of optional assignments but I'm afraid I'm not the correct person to take a stand on whether or not optional assignments should be considered always "completed".

I left a couple of nits but overall, this is a very good implementation of a feature that, I think, would also be useful. I'm personally fine with the code as-is, so I approve this PR with pleasure.

Thank you very much, @Atlanex for your contributions! They are more than welcome. If you're interested in getting more involved in the project, don't hesitate to contact us. Also, if you're in Otaniemi some day, I'd be happy to buy you a student lunch in acknowledgement of your work. :)

nikke234:
> @nikke234 raised a good point of optional assignments but I'm afraid I'm not the correct person to take a stand on whether or not optional assignments should be considered always "completed".

From the perspective of O1, we don't want to consider optional assignments completed unless they have actually been completed by the student. This gets a bit tricky though, as it's not possible to say whether the optional assignment has been truly completed or not --- it gets zero points regardless of whether the submission was correct or not. Therefore considering an optional assignment complete if it has at least one submission is not ideal, but I think as a "compromise" it's good enough.

Based on these discussions I suggest the following:
  + Move `isComplete` to `Exercise.java` and add a test for it in `ExerciseTest.java`
  + Modify `isComplete` to not return `true` by default for optional assignments

Atlanex:
I'll make the changes based on the suggestions

OlliKiljunen (approved):
Perfect! :+1:

nikke234:
Just fix the checkstyle complaints and this is good to go :+1:

OlliKiljunen:
This is gonna be included in the next release that will be out probably in the next week (or week after that). Currently, we are not publishing intervening versions anywhere but if you wanna use this feature, you can follow the steps in section **1.2.B** of **TESTING.md** to install it to your IDE.

Thank you @Atlanex for an excellent PR.

PR #398, opened 2020-09-24
Add github workflow for publishing plugin

nikke234:
The current CI process is broken, since we constantly try to publish again with the same version number leading to failed builds. This is a test/attempt at creating a GitHub workflow that is tied to prerelease tags.

Based on:
https://github.com/JetBrains/intellij-platform-plugin-template/blob/main/.github/workflows/release.yml

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,27 @@
+name: Release
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  release:
+    name: Publish Plugin
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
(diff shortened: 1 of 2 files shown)


OlliKiljunen:
Sounds like a good idea but I'm afraid I don't totally understand how this works. Would this replace deployment via Travis? What will be version number and is it set manually or automatically. And is this triggered when we push git tags or when?

nikke234:
The idea is this:

  + Remove publishing from the Travis pipeline
  + We tag those version increment and change note commits with a prerelease tag
  + This workflow gets triggered and publishes to the test channel

I thought that it would be safest to test if this even works at all on the test channel, as I don't have much experience with GitHub actions :smile:

OlliKiljunen (approved):
I'm fine with trying this out! :+1: Let's see what happens. :D

PR #399, opened 2020-09-27
Hide upper-level item in exercise tree if all its children are hidden

OlliKiljunen:
# Description of the PR

As both @nikke234 and @Atlanex have noted, exercise group items ("weeks") are still shown in the exercise tree even if all their child nodes are hidden. This PR fixes the issue.

`SelectableNodeViewModel::applyFilter` is set to return `Optional<Boolean>` instead of plain `boolean` value following the similar logic as `Filter::apply` but returning non-empty value if any of its descendants return non-empty (and applying or-logic to boolean values). That is,
**a)** if node passes the filter OR it's not applicable but one of it descendants returns **true**, node is **visible** (and returns true)
**b)** else if node doesn't pass the filter OR it's not applicable but one of its descendants returns **false**, node is **invisible** (and returns false)
**c)** else (if node is not applicable to the filter and all of its descendants returns **empty**), node is **visible** (and returns empty)

This solution is fully unaware of domain semantics.

In practice, there is one caveat: in O1 course, there are some exercise groups / weeks that simply don't have any children (why?). This semantically-oblivious implementation is unable to distinguish them from Submissions (both of them are leaf nodes and inapplicable to our filters that only handle exercises, and thus they fall the **c)** category above). I suggest that we just hide those exercise groups (weeks) altogether (or is there reason to show them anyways?).

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


src/test/java/fi/aalto/cs/apluscourses/presentation/base/SelectableNodeViewModelTest.java
```diff
@@ -101,25 +99,30 @@ public void testApplyFilterWhenChildReturnsTrue() {
     when(filter.apply(child1)).thenReturn(Optional.of(true));
     when(filter.apply(child2)).thenReturn(Optional.empty());
 
-    assertTrue(node.applyFilter(filter));
+
+    Optional<Boolean> result = node.applyFilter(filter);
+    assertTrue(result.isPresent());
+    assertTrue(result.get());
```
nikke234, line 105: Perhaps add the no-sonar-comments to these since sonar is too stupid to recognize the previous call to `isPresent`

src/main/java/fi/aalto/cs/apluscourses/presentation/base/SelectableNodeViewModel.java
```diff
@@ -35,18 +35,19 @@ public SelectableNodeViewModel(@NotNull T model,
...
+  public Optional<Boolean> applyFilter(Filter filter) {
+    Optional<Boolean> result = filter.apply(this);
+    for (SelectableNodeViewModel<?> child : children) {
+      if (Thread.currentThread().isInterrupted()) {
+        return Optional.empty();
+      }
+      Optional<Boolean> childResult = child.applyFilter(filter);
+      if (childResult.isPresent()) {
+        result = Optional.of(result.orElse(false) || childResult.get());
+      }
```
nikke234, line 47: Am I understanding the logic here correctly, that if this node does NOT pass the filter (`false`) but one of the children does pass the filter (`true`), then this node becomes visible after all?

OlliKiljunen: You were right. I changed it now so that the descendants matter only if `filter.apply(this)` returns empty.

src/main/java/fi/aalto/cs/apluscourses/presentation/base/SelectableNodeViewModel.java
```diff
@@ -31,18 +32,18 @@ public SelectableNodeViewModel(@NotNull T model,
...
-    if (Thread.currentThread().isInterrupted()) {
-      return true;
+  public Optional<Boolean> applyFilter(Filter filter) {
+    Optional<Boolean> myResult = filter.apply(this);
+    Optional<Boolean> result = myResult;
+    for (SelectableNodeViewModel<?> child : children) {
+      if (Thread.currentThread().isInterrupted()) {
+        return Optional.empty();
+      }
+      result = OptionalBooleanLogic.or(result, child.applyFilter(filter));
```
nikke234, line 42: I personally liked the non-abstracted version more :smile: It's not completely obvious how `OptionalBooleanLogic` deals with empty optionals, so now this adds an additional indirection which has to be checked every time the code is read

OlliKiljunen: Removed `OptionalBooleanLogic` I also changed it to skip child nodes if the result is false (node is hidden). Furthermore, I made it throw `InterruptedException` if interrupted to follow the convention.

OlliKiljunen:
We should make filters return something else than `Optional<Boolean>` because those are PITA to deal with. Something like [this](https://thedailywtf.com/articles/What_Is_Truth_0x3f_)...

nikke234:
> We should make filters return something else than `Optional<Boolean>` because those are PITA to deal with. Something like [this](https://thedailywtf.com/articles/What_Is_Truth_0x3f_)...

Let's not :laughing:

nikke234 (approved):
Looks good!

PR #400, opened 2020-09-30
Fix notifications appearing in wrong windows

Atlanex:
# Description of the PR
Some notifiers used null instead of project, so the notifications could appear in wrong windows

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/model/SubmissionStatusUpdater.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/model/SubmissionStatusUpdater.java
@@ -1,12 +1,17 @@
 package fi.aalto.cs.apluscourses.model;
 
 import com.intellij.notification.Notifications;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.ui.Messages;
 import fi.aalto.cs.apluscourses.intellij.notifications.FeedbackAvailableNotification;
 import fi.aalto.cs.apluscourses.intellij.notifications.Notifier;
 import java.io.IOException;
 import org.jetbrains.annotations.NotNull;
 
 public class SubmissionStatusUpdater {
+
+  @NotNull
+  private final Project project;
 
   @NotNull
   private final ExerciseDataSource dataSource;
@@ -45,7 +50,8 @@ public class SubmissionStatusUpdater {
   /**
    * Construct a submission status updater with the given parameters.
    */
-  public SubmissionStatusUpdater(@NotNull ExerciseDataSource dataSource,
+  public SubmissionStatusUpdater(@NotNull Project project,
+                                 @NotNull ExerciseDataSource dataSource,
                                  @NotNull Authentication authentication,
                                  @NotNull Notifier notifier,
                                  @NotNull String submissionUrl,
@@ -53,6 +59,7 @@ public class SubmissionStatusUpdater {
                                  long interval,
                                  long increment,
                                  long timeLimit) {
+    this.project = project;
     this.dataSource = dataSource;
     this.authentication = authentication;
     this.notifier = notifier;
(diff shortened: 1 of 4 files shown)


src/main/java/fi/aalto/cs/apluscourses/model/SubmissionStatusUpdater.java
```diff
@@ -45,14 +49,16 @@
   /**
    * Construct a submission status updater with the given parameters.
    */
-  public SubmissionStatusUpdater(@NotNull ExerciseDataSource dataSource,
+  public SubmissionStatusUpdater(@NotNull Project project,
```
nikke234, line 52: I would perhaps make this `@Nullable` instead, since `ActionEvent#getProject` is also `@Nullable`. Currently of course `SubmitExerciseAction` only works if the project isn't null, but since `SubmissionStatusUpdater` only uses the project for the notification, I don't think it should require it to be non-null.

src/main/java/fi/aalto/cs/apluscourses/model/SubmissionStatusUpdater.java
```diff
@@ -1,13 +1,17 @@
...
+import com.intellij.openapi.project.Project;
 import fi.aalto.cs.apluscourses.intellij.notifications.FeedbackAvailableNotification;
 import fi.aalto.cs.apluscourses.intellij.notifications.Notifier;
 import java.io.IOException;
 import org.jetbrains.annotations.NotNull;
 
 public class SubmissionStatusUpdater {
 
+  @NotNull
+  private final Project project;
```
OlliKiljunen, line 13: Remark: we've faced some issues with long-living objects with references to IntelliJ's `Project` objects but the lifetime of this object is so short, it shouldn't be an issue here.

src/main/java/fi/aalto/cs/apluscourses/model/SubmissionStatusUpdater.java
Atlanex: Good point. Should I leave SubmissionStatusUpdaterTest as is?

src/test/java/fi/aalto/cs/apluscourses/model/SubmissionStatusUpdaterTest.java
```diff
@@ -67,13 +69,14 @@ public void testSubmissionStatusUpdater() throws InterruptedException {
 
     assertEquals("The submission results are not fetched anymore after feedback is available",
         3, dataSource.getSubmissionResultFetchCount());
-    verify(notifier).notify(any(FeedbackAvailableNotification.class), isNull());
+    verify(notifier).notify(any(FeedbackAvailableNotification.class), any(Project.class));
```
OlliKiljunen, line 72: Nit: could check that the `Project` object is the same that was given to the `SubmissionStatusUpdater`. But this is a truly minor thing.

src/main/java/fi/aalto/cs/apluscourses/model/SubmissionStatusUpdater.java
nikke234: Yep the mock project is fine :+1:

OlliKiljunen (approved):
Very good. 👍 Approved with delight.

src/main/java/fi/aalto/cs/apluscourses/model/SubmissionStatusUpdater.java
nikke234: In this case the reference is luckily gone after at most 3 minutes. Which reminds me that we should add (at some point) some sort of "Failed to get feedback from A+" notification instead of just silently stopping... :thinking:

nikke234 (approved):
Great stuff!

PR #406, opened 2020-09-30
Add possibility to select default group

nikke234:
# Description of the PR

This PR adds a check box that makes the submission group the default value of the combo box in future submissions. Importantly, the group combo box is still there in future submissions, so the group can still be changed. The check box is also always visible, and it is checked by default if a default group exists. Unchecking the check box manually removes the default group from future submissions (analogous to a settings reset).

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [ ] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
@@ -215,11 +215,27 @@ public class SubmitExerciseAction extends AnAction {
     groups.add(0, new Group(-1, Collections
         .singletonList(getText("ui.toolWindow.subTab.exercises.submission.submitAlone"))));
 
-    SubmissionViewModel submission =
-        new SubmissionViewModel(exercise, submissionInfo, history, groups, files, language);
+    // Find the group from the available groups that matches the default group ID.
+    // A group could be removed, so this way we check that the default group ID is still valid.
+    Optional<Long> defaultGroupId = PluginSettings.getInstance().getDefaultGroupId();
+    Group defaultGroup = defaultGroupId
+        .flatMap(id -> groups
+            .stream()
+            .filter(group -> group.getId() == id)
+            .findFirst())
+        .orElse(null);
+
+    SubmissionViewModel submission = new SubmissionViewModel(
+        exercise, submissionInfo, history, groups, defaultGroup, files, language);
 
     if (!dialogs.create(submission, project).showAndGet()) {
       return;
+    }
+
+    if (Boolean.TRUE.equals(submission.makeDefaultGroup.get())) {
+      PluginSettings.getInstance().setDefaultGroupId(submission.selectedGroup.get().getId());
+    } else {
+      PluginSettings.getInstance().clearDefaultGroupId();
     }
 
     String submissionUrl = exerciseDataSource.submit(submission.buildSubmission(), authentication);
--- a/src/test/java/fi/aalto/cs/apluscourses/ui/exercise/SubmissionDialogTest.java
+++ b/src/test/java/fi/aalto/cs/apluscourses/ui/exercise/SubmissionDialogTest.java
@@ -5,15 +5,21 @@ import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 
 import com.intellij.testFramework.LightIdeaTestCase;
+import fi.aalto.cs.apluscourses.model.Exercise;
 import fi.aalto.cs.apluscourses.model.Group;
+import fi.aalto.cs.apluscourses.model.SubmissionHistory;
+import fi.aalto.cs.apluscourses.model.SubmissionInfo;
 import fi.aalto.cs.apluscourses.model.SubmittableFile;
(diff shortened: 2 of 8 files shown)


src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
```diff
@@ -215,13 +215,29 @@ private void trySubmit(@NotNull Project project)
...
+        exercise, submissionInfo, history, groups, defaultGroup, files, language);
 
     if (!dialogs.create(submission, project).showAndGet()) {
       return;
     }
 
+    if (Boolean.TRUE.equals(submission.makeDefaultGroup.get())) {
+      PluginSettings.getInstance().setDefaultGroupId(submission.selectedGroup.get().getId());
+    } else {
+      PluginSettings.getInstance().clearDefaultGroupId();
```
OlliKiljunen, line 238: Or should the default just to stay being what it was.

src/test/java/fi/aalto/cs/apluscourses/ui/exercise/SubmissionDialogTest.java
```diff
@@ -43,31 +53,37 @@ public String getSubmissionNumberText() {
...
-        .getSubmissionCountText();
-    return viewModel;
+  private SubmissionViewModel createViewModel(@NotNull String exerciseName,
+                                              @NotNull List<Group> availableGroups,
+                                              @Nullable Group defaultGroup,
+                                              @NotNull List<SubmittableFile> files,
+                                              int numberOfSubmissions,
+                                              int maxNumberOfSubmissions) {
+    return new SubmissionViewModel(
+        new Exercise(1, exerciseName, "http://www.fi", 0, 0, maxNumberOfSubmissions),
```
OlliKiljunen, line 63: I haven't seen that URL for a long long time. IIRC, it used to be some quite popular Finnish search engine in the late 90s - early 2000s. 😃 Thanks for a nostalgic moment!

OlliKiljunen (approved):
Great! 👍 One question about how we want it to function. But for me, this goes as-is.

OlliKiljunen:
And BIG thanks for fixing TESTING.md! :relaxed:🙏

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
nikke234: That's a good point. So it could be used to make another group the default value, but "undefaulting" a group requires a settings reset.

My only concern is whether it's unintuitive that unchecking it actually has no effect once you already have a default group

src/test/java/fi/aalto/cs/apluscourses/ui/exercise/SubmissionDialogTest.java
nikke234: I myself only very recently discovered that URL when I realized that there's nothing special about `www` in a URL :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/SubmitExerciseAction.java
nikke234: Perhaps the correct solution here is disable the checkbox if the selected group is the default group?

OlliKiljunen: I think the current way is ok.

Disabling checkbox could be a bit confusing for users even though it would kinda make sense to me.

More complex logic I thought is that when group selection is changed by dropdown list, the status of the checkbox auto-updares to reflect whether or not that is the current default.  If user doesn't change that, the default remains. If user unchecks automatically checked checkbox, default is reset. If user checks auto-unchecked checkbox, default is changed to be the new value.

IF selected == default XOR makeDefaultChecked THEN default = makeDefaultChecked ? selected : null

nikke234: So just to make sure I understood what you suggested correctly: otherwise equivalent to the current solution (unchecking resets), but selecting a non-default group automatically unchecks the check box?

OlliKiljunen: Almost just as you said but with an additional rule that if the selected group is not the default group and the checkbox is unchecked (which is automatically is after selecting that group from dropdown list), then the default group is not changed nor cleared (it keeps the same).

But I see now that this is quite confusing logic too. Perhaps just go with what it is now. That's good and sensible, after all.

nikke234: Sounds good, will keep it as is so it's free of too much complexity

PR #407, opened 2020-10-02
Fixed filtering bug

OlliKiljunen:
# Description of the PR

+Fixed rendering issues with exercise tree filtering by rewriting `TreeNode` subclass. This follows a suggestion made by @nikke234 in Slack.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


src/main/java/fi/aalto/cs/apluscourses/ui/base/CompositeTreeNode.java
```diff
@@ -0,0 +1,79 @@
...
+import fi.aalto.cs.apluscourses.presentation.base.FilterableTree;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.stream.Collectors;
+import javax.swing.tree.TreeNode;
+import jdk.internal.jline.internal.Nullable;
+import org.jetbrains.annotations.NotNull;
+
+public class CompositeTreeNode implements TreeNode {
```
nikke234, line 12: This implementation looks good, but now that there is no longer the additional visibility property, I wonder if we need this class at all? Since it's only used internally in `TreeView`, we could use `DefaultMutableTreeNode` there to avoid reinventing the wheel, and the mutability wouldn't matter, since the mutable tree node is just an interal implementation detail of `TreeView`.

OlliKiljunen: That's sounds reasonable enough, even though this class is so basic it barely brings any complexity.

I'd like to keep the tree construction logic out of `TreeView` class for better unit-testability but that shouldn't be a problem; a util class or something will do.

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
```diff
@@ -42,24 +56,38 @@ public TreeView() {
...
+      }
     }
-    for (Tree subtree : children) {
-      node.add(createNode(subtree));
+  }
+
+  private void registerViewModel() {
+    synchronized (viewModelLock) {
+      if (viewModel != null) {
+        viewModel.filtered.addListener(this, TreeView::update);
```
nikke234, line 80: Could this possibly call `update` (which modifies the UI tree model) outside of the EDT?

nikke234 (approved):
Very nice changes! Also, I can't reproduce the bug anymore so the fix seems to work :+1: Just one question regarding modifying the tree from outside of the EDT

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
OlliKiljunen: That's a valid point. It probably could be reasoned out whether or not that method is ever called outside of the EDT but it's better to be sure and add `invokeLater` there somewhere.

What is more, if that method is called outside of EDT, it could, in theory at least, be possible that of two subsequent filterings, the latter would be rewritten by the former (if it happened very soon after the other). That would result in a situation where tree is filtered by a different filter than what option buttons display. But that is a very minor concern, anyways.

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
```diff
@@ -42,24 +57,38 @@ public TreeView() {
...
+      }
     }
-    for (Tree subtree : children) {
-      node.add(createNode(subtree));
+  }
+
+  private void registerViewModel() {
+    synchronized (viewModelLock) {
+      if (viewModel != null) {
+        viewModel.filtered.addListener(this, TreeView::update, SwingUtilities::invokeLater);
```
OlliKiljunen, line 81: [IntelliJ doc](https://github.com/JetBrains/intellij-community/blob/cf665c226927d79da572338d860b5ff7ff5aa812/platform/core-api/src/com/intellij/openapi/application/ModalityState.java#L44)
> SwingUtilities#invokeLater(Runnable) [...] may be used in the purely UI-related code, but not with anything that deals with PSI or VFS.

I'd say `SwingUtilities::invokeLater` is the correct choice here. This class is independent of IntelliJ framework; introducing `Application` reference here would make it dependent on the framework (and kinda justify itself). `update` method is not expected to call any framework-critical code, so there's no danger in that.

As our project grows, the modularity of the code becomes more and more important factor to keep up maintainability. Being strict with component dependencies and interfaces is the best approach for that. So even though calling `Application::invokeLater` here wouldn't hurt much on modularity itself, it would be yet another step on the slippery slope towards more component-model-breaking cross-references that ultimately turn the code into unmanageable spaghetti.

PR #408, opened 2020-10-05
Do not collapse exercise tree [#319]

### LINKED ISSUE #319: When refreshing the assignment pane, it would be nice if it didn’t fold every item in the tree view
needs further investigation

OlliKiljunen:
# Description of the PR

Closes #319 

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [x] Yes
- [ ] Not relevant


OlliKiljunen:
It should be noted that if a node is hidden and then re-displayed by changed filtering, its expanded-status is not preserved from the previous time it was visible. I don't think this is a severe issue, though.

src/main/java/fi/aalto/cs/apluscourses/ui/utils/TreePathCoder.java
```diff
@@ -0,0 +1,23 @@
+package fi.aalto.cs.apluscourses.ui.utils;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import javax.swing.tree.TreePath;
+import org.jetbrains.annotations.Nullable;
+
+public abstract class TreePathCoder<T> {
```
nikke234, line 8: What's a tree path coder?

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
```diff
@@ -130,4 +157,17 @@ public void mouseClicked(@NotNull MouseEvent e) {
       }
     }
   }
+
+  private static class MyTreePathCoder extends TreePathCoder<String> {
```
nikke234, line 161: ..... `StringTreePathCoder`?

src/main/java/fi/aalto/cs/apluscourses/ui/utils/TreeModelTraversal.java
```diff
@@ -0,0 +1,60 @@
...
+
+import fi.aalto.cs.apluscourses.utils.Streamable;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.stream.Stream;
+import javax.swing.tree.TreeModel;
+import javax.swing.tree.TreePath;
+import org.jetbrains.annotations.NotNull;
+
+public class TreeModelTraversal {
```
nikke234, line 11: Is there a reason that this isn't a utility class with `TreeModel` as a method parameter instead of an instance variable?

src/main/java/fi/aalto/cs/apluscourses/ui/utils/TreePathCoder.java
```diff
@@ -0,0 +1,23 @@
...
+  private final ConcurrentMap<TreePath, T> map = new ConcurrentHashMap<>();
+
+  public T code(@Nullable TreePath treePath) {
+    return treePath == null ? emptyCode()
+        : map.computeIfAbsent(treePath, this::codeInternal);
```
nikke234, line 13: lambda-phobia :smile:

src/main/java/fi/aalto/cs/apluscourses/ui/utils/TreePathCoder.java
nikke234: I think I get it now, but would the more correct word be `Encoder` then?

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
nikke234: Or even better, `TreePathString(En)coder`

src/main/java/fi/aalto/cs/apluscourses/ui/utils/TreePathCoder.java
OlliKiljunen: Well, when I was young, there were no lambdas in Java.  We had to actually implement ourselves each and every interface we ever wanted to use.  And then we skied 50 km to school each morning in -30 °C blizzard while wolves were howling.

But the most incredible thing is that, on those days of old, Java was run in a web browser and _it just worked_. On every platform.  And absolutely no lambdas were needed for that.

src/main/java/fi/aalto/cs/apluscourses/ui/utils/TreeModelTraversal.java
OlliKiljunen: Well, I'd say this is more OOP way to do it - no need to pass the same object reference first to `traverse(TreePath)` then to `getChildNodeStream(Object)`, then to `NodeIterator`. Is there a reason to do it in other way?

The name of the class is, however, a bit misleading. Perhaps `TreeModelTraverser` would be better. After all, the same object can be used for more than one traversals.

src/main/java/fi/aalto/cs/apluscourses/ui/utils/TreePathCoder.java
OlliKiljunen: Yes. English difficult.

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
OlliKiljunen: Hmm, maybe, but... it's not a generic to-string-encoder but also aware of the semantics of the node objects. It's a `TreePathEncoder` used by that class, `TreeView`'s own `TreePathEncoder` (hence, "My").

src/main/java/fi/aalto/cs/apluscourses/ui/utils/TreePathCoder.java
nikke234: "Can't teach an old dog new tricks" or something like that... :smile:

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
nikke234: Well, yes it's not generic, but it's also private so no one is going to use it by mistake, I do consider this "My" and "Our" prefixing quite indescriptive, and it's quite similar to the "Impl" suffix naming perversion

src/main/java/fi/aalto/cs/apluscourses/ui/utils/TreeModelTraversal.java
nikke234: It's not used in `getChildNodeStream` which is why I initially asked, it seemed like an overly complicated way of passing a parameter. Now I see that the iterator uses it, so this way makes sense. I agree that `Traverser` is a better name :+1:

nikke234 (approved):
Important feature with a good implementation! :+1:

PR #413, opened 2020-10-09
Update MainViewModel after receiving FeedbackAvailableNotification

### LINKED ISSUE #320: List of submissions should be refreshed when new feedback is available
automatically (remembering the state of the "openness")

Atlanex:
# Description of the PR

Closes #320

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


OlliKiljunen (approved):
Nice solution. 👍 Thank you!

What I don't like is that this PR cements model components' dependency of code in `apluscoures.intellij` package. We'd like to get rid of those dependencies at some point (hopefully sooner than later). But it's a wider problem and not expected to be solved here.

So I'm happy to approve this. ☺️

nikke234:
I think there is unfortunately a thread safety issue here. `updateMainViewModel` isn't synchronized and it calls `MainViewModelUpdater#restart`, which isn't a thread safe method. Therefore it's possible that a nasty race condition occurs with `SubmissionStatusUpdater` doing an update from a background thread and the user doing an update from the UI thread.

I'm not sure what the "most correct" solution here is, but perhaps making `MainViewModelUpdater#restart` `synchronized` is the way to go? I guess we could make `MainViewModelUpdater#start` synchronized as well?

nikke234:
Actually `MainViewModelUpdater#start` shouldn't be synchronized, because starting the same thread twice would be an error anyways. But `restart` should be synchronized

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
```diff
@@ -229,7 +229,7 @@ public void start() {
    * Interrupts the updater and starts it again. This can be used to trigger an instant main view
    * model update, skipping the sleep interval.
    */
-  public void restart() {
+  public synchronized void restart() {
```
OlliKiljunen, line 232: Now when I think of it, it actually looks a bit suspicious that `start` ain't synchronized because it accesses a field that is set in a synced method `restart`. However, as the idea is that `restart` can only be called after `start` has been called ('after' in the sense of Java's _happens-before_ relation).

But is it so? I guess `start` is called almost immediately after the object is created but is there still a small time window where an external `restart` call could happen? What I suggest (not as a part of this PR, though - it is fine) is that we'd get rid of `start`/`restart` duality and only have one synced method with the following code:
```java
 Optional.ofNullable(thread).ifPresent(Thread::interrupt);
thread = new Thread(this::run);
thread.start();
```
And `thread` would be just `null` initially.

nikke234: That small time frame doesn't exist, because `start` is called in the `computeIfAbsent` function in `PluginSettings`, so the reference to the updater becomes "publicly available" only after `start` has already been called.

That being said, the current interface does allow that risk if the code using it isn't careful. I really like your suggested solution, and I think it's the way to go :+1:

nikke234 (approved):
Nice! Thanks again

PR #415, opened 2020-10-12
Unify plugin publishing for test and stable version

### LINKED ISSUE #310: fix the CI/CD process
does not break when you merge to test (no changes in version numbers)

nikke234:
# Description of the PR

Adds a workflow that gets triggered by a new release. This is very similar to the existing workflow that gets triggered by a pre-release, except that it publishes to the stable channel instead of the test channel. Closes #310.
 
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -1,8 +1,8 @@
-name: Release
+name: Pre-release
 
 on:
   release:
-    types: [released]
+    types: [prereleased]
 
 jobs:
   release:
@@ -22,5 +22,6 @@ jobs:
       - name: Publish Plugin
         env:
           INTELLIJ_PUBLISH_TOKEN: ${{ secrets.PUBLISH_TOKEN }}
+          INTELLIJ_PUBLISH_CHANNEL: test
         run: ./gradlew publishPlugin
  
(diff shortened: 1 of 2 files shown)


OlliKiljunen (approved):
As far as I understand something of this, looks good. 👍

Can we go through once more, how are releases triggered? It's git tags, right? And is naming of tags significant in whether or not they are considered pre-release?

nikke234:
> As far as I understand something of this, looks good. +1
> 
> Can we go through once more, how are releases triggered? It's git tags, right? And is naming of tags significant in whether or not they are considered pre-release?

Sure!

Yep, it's Git tags. If you open the GitHub menu for creating new tags, there is an option for whether the release is a pre-release or a normal release (you can look at the `v1.7` tag to see what a pre-release looks like), so naming should not have anything to do with it. (I'm not sure what this pre-release option corresponds to in Git tags). Then simply editing the tag and changing that option from pre-release to normal release should trigger the release event. 

My only uncertainty here is that checkout action, but from it's documentation:
>  When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event. Otherwise, uses the default branch.

As I understand it, we are checking out from the repository that triggered the workflow, so the reference should be the (pre-)release event.

OlliKiljunen:
> Sure!
> [...]

Perfect. Thank you! 👍

PR #419, opened 2020-10-14
Show notification for module updates available only once

### LINKED ISSUE #369: notifications of modules having updates is shown once only during the session
instead of every 15 minutes

Atlanex:
# Description of the PR

Closes #369 

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
@@ -13,9 +13,9 @@ import fi.aalto.cs.apluscourses.intellij.services.PluginSettings;
 import fi.aalto.cs.apluscourses.model.Course;
 import fi.aalto.cs.apluscourses.model.Module;
 import java.net.URL;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 import org.jetbrains.annotations.NotNull;
@@ -40,7 +40,7 @@ public class MainViewModelUpdater {
 
   private Notification newModulesVersionsNotification = null;
 
-  private final Set<String> notifiedModules = new HashSet<>();
+  private final Set<String> notifiedModules = ConcurrentHashMap.newKeySet();
 
   /**
    * Construct a {@link MainViewModelUpdater} with the given {@link MainViewModel}, project, and
--- a/src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
@@ -13,8 +13,9 @@ import fi.aalto.cs.apluscourses.intellij.services.PluginSettings;
 import fi.aalto.cs.apluscourses.model.Course;
 import fi.aalto.cs.apluscourses.model.Module;
 import java.net.URL;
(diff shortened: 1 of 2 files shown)


src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
```diff
@@ -36,6 +39,8 @@
 
   private Notification newModulesVersionsNotification = null;
 
+  private final List<String> notifiedModules = new ArrayList<>();
```
OlliKiljunen, line 42: Should this be
1. a (hash)set?
2. synced / "concurrent" collection?

OlliKiljunen (commented):
Left one question, I'd like to discuss before approval.

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
OlliKiljunen: If you had `Set<String> notifiedModules = ConcurrentHashMap.newKeySet()`, your filtering method could be just:
```java
return modules.stream()
  .filter(m -> notifiedModules.add(m.getName())
  .collect(Collectors.toList());
```
because `Set.add(elem)` returns true iff `elem` wasn't there in the first place.

Yes, side effects in stream filters are bad but this is one of those rare situations where we accept them.

OlliKiljunen (commented):
Added a suggestion.

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModelUpdater.java
Atlanex: That's better so I did that

nikke234: I get shivers just looking at that filter :smile: 

But more importantly, I see no reason to use a set with concurrency support here. Just a standard `HashSet` will do

nikke234: Argh, there is of course the case where `restart` is called, and it is then possible that the `run` method of a single updater instance is executing on two different threads simultaneously, so it should be synced after all. Sorry, my bad.

But that has me thinking now, is it even safe for `run` to be running on two threads at the same time?

@OlliKiljunen I remember we had some `IllegalStateException` stack traces from `IntelliJCourse#unregister`, but I can't find them anymore. `run` calls `unregister` for the course stored in `MainViewModel`, so this could actually be the root cause of the illegal state where unregister is called multiple times.

OlliKiljunen: I'll take a look at that.

> I get shivers just looking at that filter 😄

Those are good shivers. :+1: But it's important to notice that the side effects here are _idempotent_ which makes it less scary.

nikke234: I think it might make more sense to make `run` synchronized instead of using a synced set here.

OlliKiljunen:
May I suggest we revert this PR to f10a25a and merge?

PR #424, opened 2020-10-25
Number of tests failed/passed shown for optional assignments

jaakkonarhi:
# Description of the PR

Test results for optional assignments are parsed from the feedback and cached to a-plus-project.json. Parsing from the feedback is probably course specific, but I could add support for more courses in the future. If the parsing isn't successful, [0/0 points] is shown instead of [n tests passed/failed]. Amount of tests passed/failed is also shown in the feedback notification

Now when all the tests have passed, the optional assignment is considered completed. It would be good to update the icon in the assignments tree.

More unit tests need to be made, and test caching probably needs to be moved from MainViewModel

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [ ] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/notifications/FeedbackAvailableNotification.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/notifications/FeedbackAvailableNotification.java
@@ -17,7 +17,7 @@ public class FeedbackAvailableNotification extends Notification {
   /**
    * Construct a notification that notifies the user that feedback is available for a submission.
    * The notification contains a link that can be used to open the feedback and the amount of
-   * points the submission got.
+   * points, or the test results the submission got.
    */
   public FeedbackAvailableNotification(@NotNull SubmissionResult submissionResult,
                                        @NotNull Exercise exercise) {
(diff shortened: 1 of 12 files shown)


OlliKiljunen (commented):
Thank you for this code! I like the way you solved the problem, good! 👍

As it often happens, only after seeing a possible solution, I realize that I actually would prefer something else than what I initially thought. In this case, I'd like to change the design so that
- the raw data from the API is what gets stored in the local file;
- any model object (exercise, exgroup, submission, ...) can store their data there in the same way; the implementation of the data-access layer should be very generic (in this PR, test results get quite a special treatment in DAL, which I see non-optimal).
- it should be each model object's responsibility to call DAL to populate itself with the required data.

That said, I don't oppose merging this PR (though I need to re-read it before that because now I only focused on higher-level design choices). As I said, this seems to be a good solution for the feature. It however introduces some extra complexity to our code, which we have to refactor away before building new things in top of this. So if this gets merged, we must make sure that the refactoring is also done at some point soon and not when it's already too late.

superseacat:
> It however introduces some extra complexity to our code, which we have to refactor away before building new things in top of this. So if this gets merged, we must make sure that the refactoring is also done at some point soon and not when it's already too late.

the refactoring might only come after Sprint 3, so I'd be careful here 😉

OlliKiljunen:
Let's close this for now but reopen when it becomes topical again.

Issue #248, opened 2020-07-24
Exercise list should support submitting using the mouse right-click key


nikke234:
This is similar to how pressing the enter key with a module selected in the module list triggers an installation.

superseacat:
overall, check that the approach for the now united "A+ Courses" subtabs is unified more-or-less

superseacat:
submit on `Ctrl + D-click`?

PR #447, opened 2021-01-08
Control clicking exercise opens submit window

jaakkonarhi:
# Description of the PR

Ctrl + left clicking an assignment opens the submission window. Also works when clicking submissions, which I think is fine. Part of #248

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [ ] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
@@ -10,8 +10,8 @@ import java.awt.event.ActionListener;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.swing.SwingUtilities;
@@ -34,7 +34,8 @@ public class TreeView extends com.intellij.ui.treeStructure.Tree {
         }
       };
 
-  private final transient HashMap<String, ActionListener> nodeAppliedListeners = new HashMap<>();
+  private final transient Set<ActionListener> ctrlClickListeners = ConcurrentHashMap.newKeySet();
+  private final transient Set<ActionListener> doubleClickListeners = ConcurrentHashMap.newKeySet();
 
   @NotNull
   public static SelectableNodeViewModel<?> getViewModel(Object node) {
@@ -98,12 +99,20 @@ public class TreeView extends com.intellij.ui.treeStructure.Tree {
     restoreExpandedState(expandedState);
   }
 
-  public void addNodeAppliedListener(String key, ActionListener listener) {
-    nodeAppliedListeners.put(key, listener);
+  public void addDoubleClickListener(ActionListener listener) {
+    doubleClickListeners.add(listener);
   }
 
-  public void removeNodeAppliedListener(String key) {
-    nodeAppliedListeners.remove(key);
+  public void removeDoubleClickListener(ActionListener listener) {
+    doubleClickListeners.remove(listener);
+  }
+
+  public void addCtrlClickListener(ActionListener listener) {
+    ctrlClickListeners.add(listener);
+  }
+
+  public void removeCtrlClickListener(ActionListener listener) {
+    ctrlClickListeners.remove(listener);
   }
 
   @NotNull
@@ -153,10 +162,10 @@ public class TreeView extends com.intellij.ui.treeStructure.Tree {
     public void mouseClicked(@NotNull MouseEvent e) {
       if (e.getClickCount() == 1 && e.isControlDown() && selectedItem != null) {
         ActionEvent actionEvent = new ActionEvent(this, ActionEvent.ACTION_PERFORMED, null);
-        nodeAppliedListeners.get("submitExercise").actionPerformed(actionEvent);
+        ctrlClickListeners.forEach(listener -> listener.actionPerformed(actionEvent));
       } else if (e.getClickCount() == 2 && selectedItem != null) {
         ActionEvent actionEvent = new ActionEvent(this, ActionEvent.ACTION_PERFORMED, null);
-        nodeAppliedListeners.get("openSubmission").actionPerformed(actionEvent);
+        doubleClickListeners.forEach(listener -> listener.actionPerformed(actionEvent));
       }
     }
   }
--- a/src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
@@ -10,8 +10,8 @@ import java.awt.event.ActionListener;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.swing.SwingUtilities;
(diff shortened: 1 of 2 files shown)


src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
```diff
@@ -151,9 +151,12 @@ public void valueChanged(@NotNull TreeSelectionEvent e) {
...
     @Override
     public void mouseClicked(@NotNull MouseEvent e) {
-      if (e.getClickCount() == 2 && selectedItem != null) {
+      if (e.getClickCount() == 1 && e.isControlDown() && selectedItem != null) {
         ActionEvent actionEvent = new ActionEvent(this, ActionEvent.ACTION_PERFORMED, null);
-        nodeAppliedListeners.forEach(listener -> listener.actionPerformed(actionEvent));
+        nodeAppliedListeners.get("submitExercise").actionPerformed(actionEvent);
+      } else if (e.getClickCount() == 2 && selectedItem != null) {
+        ActionEvent actionEvent = new ActionEvent(this, ActionEvent.ACTION_PERFORMED, null);
+        nodeAppliedListeners.get("openSubmission").actionPerformed(actionEvent);
```
OlliKiljunen, line 159: TreeView should be a domain-ignorant class; it shouldn't deal with concepts such as "submission". It's a general purpose tree widget.

So if anything, the keys should be something like "doubleClick" and "ctrlClick", or maybe more preferably something abstract like "nodePrimaryAction", "nodeSecondaryAction".

I wouldn't, however, introduce any magic strings into the public interface of this class (they can be used internally, if you like). Instead, I would make a separate add\*Listener and remove\*Listener methods for different kind of events.

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
```diff
@@ -34,7 +34,7 @@
         }
       };
 
-  private final transient Set<ActionListener> nodeAppliedListeners = ConcurrentHashMap.newKeySet();
+  private final transient HashMap<String, ActionListener> nodeAppliedListeners = new HashMap<>();
```
OlliKiljunen, line 37: Should probably be ConcurrentHashMap

OlliKiljunen (commented):
In general, this looks pretty nice! :+1: I have a couple of concerns I discuss in my comments. See them and tell your thoughts of them! :)

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
```diff
@@ -98,12 +98,12 @@ private void update() {
     restoreExpandedState(expandedState);
   }
 
-  public void addNodeAppliedListener(ActionListener listener) {
-    nodeAppliedListeners.add(listener);
+  public void addNodeAppliedListener(String key, ActionListener listener) {
+    nodeAppliedListeners.put(key, listener);
```
OlliKiljunen, line 102: In my previous comments, I overlooked one issue. As implemented here, this allows only one listener per key. It's possible that we never need more than one but at least in that case the name of this method should reflect that.

Anyways, I would probably just instead of key-listener map have to separate set of listeners, one per an event type. And add and remove methods for both.

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
jaakkonarhi: I think separate add*Listener and remove*Listener methods for different kind of events sounds good, I'll do that

OlliKiljunen (commented):
One thing

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
nikke234: Sounds good to me too, but in that case I would prefer two sets for primary and secondary listeners respectively, instead of a map (even if only used internally).

nikke234: Oops, Olli already suggested this in his later comment, but yes, I agree

nikke234: The javadocs for `add*Listener` could also mention when primary and secondary actions are triggered, since "primary" and "secondary" don't make it obvious

nikke234 (commented):
This is a nice addition and a good idea! :+1: 

Eventually we'll have to figure out how to let users know of this feature, but it is nice to have for "power users" even as is.

Btw, does double-clicking (without `ctrl`) a submission also open the submission dialog, or does that open the submission in the browser?

jaakkonarhi:
Yes, double clicking submissions still opens them in a browser @nikke234

jaakkonarhi:
Ctrl-clicking a selected assignment doesn't work since it unselects it

nikke234:
So if I understood correctly:

1. ctrl-clicking an unselected assignment opens the submission dialog
2. ctrl-clicking a selected assignment unselects it
3. ctrl-clicking a submission opens the submission dialog for that submission
4. double-clicking a submission opens the submission in a browser.

Number 3 and 4 is potentially a weird combination, but then again I think the "submit assignment" icon also works when a submission (instead of an assignment) is selected, so it makes sense.

Number 2 however seems undesirable/unintuitive? I think it would be nice if ctrl-click would always submit, although I wouldn't say it's worth it if it requires significant effort

superseacat:
can you please link [as described](https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#:~:text=Manually%20linking%20a%20pull%20request%20to%20an%20issue,-Anyone%20with%20write&text=On%20GitHub%2C%20navigate%20to%20the,right%20sidebar%2C%20click%20Linked%20issues.) @jaakkonarhi

nikke234:
It's related, but this doesn't directly close it, so let's unlink the issue for now.

superseacat:
> 1. ctrl-clicking an unselected assignment opens the submission dialog
> 2. ctrl-clicking a selected assignment unselects it
> 3. ctrl-clicking a submission opens the submission dialog for the corresponding assignment
> 4. double-clicking a submission opens the submission in a browser.

I suggest thinking about this approach a bit more... Currently, in **modules** subtab, `ctrl+click` regardless of the previous selection state selects and unselects correspondingly, however, allows for multiple selections. And hence, the multiple selections do not make that much sense in **assignments** subtab, how much do we want to retain the uniform approach between subtabs?

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
```diff
@@ -151,9 +160,12 @@ public void valueChanged(@NotNull TreeSelectionEvent e) {
 
     @Override
     public void mouseClicked(@NotNull MouseEvent e) {
```
superseacat, line 162: do you think the logic inside the method can be tested fully or maybe at least to the extent of the conditional statements?

jaakkonarhi:
I'm starting to feel that the ctrl clicking is a bit useless, since you can submit with a single click with the popup menu, so I'm focusing on that now

superseacat:
> I'm starting to feel that the ctrl clicking is a bit useless, since you can submit with a single click with the popup menu, so I'm focusing on that now

let's discuss after the daily!
A13usaonutL _AC_CLa_2140,2000_61+2E4yqnzL png_0,0,2140,2000+0 0,0 0,2140 0,2000 0_SX466 _SX _UX _SY _UY_

nikke234:
> I'm starting to feel that the ctrl clicking is a bit useless, since you can submit with a single click with the popup menu, so I'm focusing on that now

Closing this, as we agreed in a meeting that this is the way to go

PR #465, opened 2021-01-12
Add course selection dialog to course project action

nikke234:
# Description of the PR

This adds a course selection dialog when the user selects "Turn Project Into A+ Course Project". The dialog is only shown if the project isn't already a course project. If the project is a course project already, then the same course will be used again without showing the selection, since it would be undesirable to turn the same project into a course project of a different course.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/presentation/CourseItemViewModel.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/presentation/CourseItemViewModel.java
@@ -1,0 +1,35 @@
+package fi.aalto.cs.apluscourses.presentation;
+
+import org.jetbrains.annotations.NotNull;
+
+public class CourseItemViewModel {
+
+  private final String name;
+  private final String semester;
+  private final String url;
+
+  /**
+   * Construct a view model with the given parameters.
+   */
+  public CourseItemViewModel(@NotNull String name, @NotNull String semester, @NotNull String url) {
+    this.name = name;
+    this.semester = semester;
+    this.url = url;
+  }
+
+  @NotNull
+  public String getName() {
+    return name;
+  }
+
+  @NotNull
+  public String getSemester() {
+    return semester;
+  }
+
+  @NotNull
+  public String getUrl() {
+    return url;
+  }
+
+}
--- a/src/main/java/fi/aalto/cs/apluscourses/presentation/CourseSelectionViewModel.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/presentation/CourseSelectionViewModel.java
@@ -1,0 +1,37 @@
+package fi.aalto.cs.apluscourses.presentation;
+
+import fi.aalto.cs.apluscourses.utils.observable.ObservableProperty;
+import fi.aalto.cs.apluscourses.utils.observable.ObservableReadWriteProperty;
+import fi.aalto.cs.apluscourses.utils.observable.ValidationError;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.List;
+
+public class CourseSelectionViewModel {
+
+  private List<CourseItemViewModel> courses;
+
+  public final ObservableProperty<String> selectedCourseUrl
+      = new ObservableReadWriteProperty<>("", CourseSelectionViewModel::validateCourseUrl);
+
+  public CourseSelectionViewModel(List<CourseItemViewModel> courses) {
+    this.courses = courses;
+  }
+
+  public CourseItemViewModel[] getCourses() {
+    return courses.toArray(new CourseItemViewModel[0]);
+  }
+
+  private static ValidationError validateCourseUrl(String courseUrl) {
+    if (courseUrl.trim().isEmpty()) {
+      return () -> "Select a course";
+    }
+    try {
+      new URL(courseUrl);
+      return null;
+    } catch (MalformedURLException e) {
+      return () -> "Invalid URL";
+    }
+  }
+
+}
--- a/src/main/java/fi/aalto/cs/apluscourses/ui/base/TextField.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/ui/base/TextField.java
@@ -1,0 +1,27 @@
+package fi.aalto.cs.apluscourses.ui.base;
+
+import fi.aalto.cs.apluscourses.ui.utils.TwoWayBindable;
+import java.awt.event.KeyAdapter;
+import java.awt.event.KeyEvent;
+import javax.swing.JTextField;
+
+/**
+ * A class deriving from {@link JTextField} that works with the {@link TwoWayBindable} class to
+ * reflect its text input to the source property.
+ */
+public class TextField extends JTextField {
+  public final transient TwoWayBindable<TextField, String> textBindable =
+      new TwoWayBindable<>(this, TextField::setText, TextField::getText);
+
+  /**
+   * Construct a TextField.
+   */
+  public TextField() {
+    addKeyListener(new KeyAdapter() {
+      @Override
+      public void keyReleased(KeyEvent e) {
+        textBindable.updateSource();
+      }
+    });
+  }
+}
--- a/src/main/java/fi/aalto/cs/apluscourses/ui/base/TextField.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/ui/base/TextField.java
@@ -1,9 +1,9 @@
 package fi.aalto.cs.apluscourses.ui.base;
 
 import fi.aalto.cs.apluscourses.ui.utils.TwoWayBindable;
-import java.awt.event.KeyAdapter;
-import java.awt.event.KeyEvent;
 import javax.swing.JTextField;
+import javax.swing.event.DocumentEvent;
+import javax.swing.event.DocumentListener;
 
 /**
  * A class deriving from {@link JTextField} that works with the {@link TwoWayBindable} class to
@@ -17,10 +17,19 @@ public class TextField extends JTextField {
    * Construct a TextField.
    */
   public TextField() {
-    addKeyListener(new KeyAdapter() {
+    getDocument().addDocumentListener(new DocumentListener() {
       @Override
-      public void keyReleased(KeyEvent e) {
+      public void insertUpdate(DocumentEvent documentEvent) {
         textBindable.updateSource();
+      }
+
+      @Override
+      public void removeUpdate(DocumentEvent documentEvent) {
+        textBindable.updateSource();
+      }
+
+      @Override
+      public void changedUpdate(DocumentEvent documentEvent) {
       }
     });
   }
--- a/src/main/java/fi/aalto/cs/apluscourses/ui/courseproject/CourseSelectionView.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/ui/courseproject/CourseSelectionView.java
@@ -1,0 +1,117 @@
+package fi.aalto.cs.apluscourses.ui.courseproject;
+
+import com.intellij.openapi.project.Project;
+import com.intellij.ui.ColoredListCellRenderer;
+import com.intellij.ui.SimpleTextAttributes;
+import com.intellij.ui.components.JBList;
+import fi.aalto.cs.apluscourses.presentation.CourseItemViewModel;
+import fi.aalto.cs.apluscourses.presentation.CourseSelectionViewModel;
+import fi.aalto.cs.apluscourses.ui.base.OurDialogWrapper;
+import fi.aalto.cs.apluscourses.ui.base.TextField;
+import fi.aalto.cs.apluscourses.utils.PluginResourceBundle;
+import icons.PluginIcons;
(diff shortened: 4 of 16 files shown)


src/main/java/fi/aalto/cs/apluscourses/ui/courseproject/CourseSelectionView.java
```diff
@@ -0,0 +1,117 @@
...
+    this.viewModel = viewModel;
+    setTitle(PluginResourceBundle.getText("ui.courseProject.courseSelection.title"));
+
+    courseList.setSelectionMode(ListSelectionModel.SINGLE_SELECTION);
+    courseList.addListSelectionListener(listSelectionEvent -> {
+      CourseItemViewModel selected = courseList.getSelectedValue();
+      if (selected != null) {
+        viewModel.selectedCourseUrl.set(selected.getUrl());
+      }
+    });
```
OlliKiljunen, line 55: This logic could perhaps be moved to view model with `selectedValueBindable` and `selectedCourseProperty`.

src/main/java/fi/aalto/cs/apluscourses/ui/courseproject/CourseSelectionView.java
```diff
@@ -0,0 +1,117 @@
...
+        PluginResourceBundle.getText("ui.courseProject.courseSelection.textField"));
+    urlField.addFocusListener(new FocusListener() {
+      @Override
+      public void focusGained(FocusEvent focusEvent) {
+        courseList.clearSelection();
+      }
+
+      @Override
+      public void focusLost(FocusEvent focusEvent) {
+      }
```
OlliKiljunen, line 85: This too (see my previous comment)

src/main/java/fi/aalto/cs/apluscourses/ui/base/TextField.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+public class TextField extends JTextField {
+  public final transient TwoWayBindable<TextField, String> textBindable =
+      new TwoWayBindable<>(this, TextField::setText, TextField::getText);
+
+  /**
+   * Construct a TextField.
+   */
+  public TextField() {
+    addKeyListener(new KeyAdapter() {
```
OlliKiljunen, line 20: I guess this is fine. 👍

src/main/java/fi/aalto/cs/apluscourses/presentation/CourseSelectionViewModel.java
```diff
@@ -0,0 +1,37 @@
...
+    this.courses = courses;
+  }
+
+  public CourseItemViewModel[] getCourses() {
+    return courses.toArray(new CourseItemViewModel[0]);
+  }
+
+  private static ValidationError validateCourseUrl(String courseUrl) {
+    if (courseUrl.trim().isEmpty()) {
+      return () -> "Select a course";
```
OlliKiljunen, line 27: A couple of nested classes would fit so well in this class but okay...

src/main/java/fi/aalto/cs/apluscourses/presentation/CourseSelectionViewModel.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public final ObservableProperty<String> selectedCourseUrl
+      = new ObservableReadWriteProperty<>("", CourseSelectionViewModel::validateCourseUrl);
+
+  public CourseSelectionViewModel(List<CourseItemViewModel> courses) {
+    this.courses = courses;
+  }
+
+  public CourseItemViewModel[] getCourses() {
+    return courses.toArray(new CourseItemViewModel[0]);
```
OlliKiljunen, line 22: Should we store `courses` as an array?

OlliKiljunen (approved):
Looks good! 👍 I wrote down some ideas but approved also as-is.

src/main/java/fi/aalto/cs/apluscourses/ui/base/TextField.java
nikke234: Argh, I meant to use a document listener, this was a left-over. I'll change it

src/main/java/fi/aalto/cs/apluscourses/presentation/CourseSelectionViewModel.java
nikke234: Yeah good idea, not used as a list anywhere anyways

src/main/java/fi/aalto/cs/apluscourses/presentation/CourseSelectionViewModel.java
nikke234: :joy: I was so close to changing the existing nested classes to lambdas as well...

src/main/java/fi/aalto/cs/apluscourses/ui/base/TextField.java
nikke234: `updateSource` is still additionally called when OK is selected in the dialog, so it should work, but document listener is clearly the more correct choice here

src/main/java/fi/aalto/cs/apluscourses/presentation/CourseItemViewModel.java
```diff
@@ -0,0 +1,35 @@
...
+public class CourseItemViewModel {
+
+  private final String name;
+  private final String semester;
+  private final String url;
+
+  /**
+   * Construct a view model with the given parameters.
+   */
+  public CourseItemViewModel(@NotNull String name, @NotNull String semester, @NotNull String url) {
```
superseacat, line 14: is the intent to have the class immutable or it just happens to have some characteristics of such?

OlliKiljunen: I don't see any reason why this wouldn't be immutable. None of the fields are expected to change.

superseacat:
should the course selection be based on branch `multicourse-support-base`, where we could first aggregate all the changes and then merge to `test`? @nikke234

nikke234:
> should the course selection be based on branch `multicourse-support-base`, where we could first aggregate all the changes and then merge to `test`? @nikke234

Ah, I forgot about that one and merged language selection already to test. However, there isn't an automatic release from test anymore so it shouldn't cause issues

src/main/java/fi/aalto/cs/apluscourses/presentation/CourseSelectionViewModel.java
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  private static ValidationError validateCourseUrl(String courseUrl) {
+    if (courseUrl.trim().isEmpty()) {
+      return () -> "Select a course";
+    }
+    try {
+      new URL(courseUrl);
+      return null;
```
superseacat, line 31: can you explain a bit why such an approach is chosen: why `null`, why `return` instead of `throw`, why `ValidationError` is an error (but not an exception at all)?

superseacat: also, I'm curious what would happen if a) `URL` is incorrect, b) `URL` is formally (syntactically) correct, but there's no course config file at "that end"?

src/main/java/fi/aalto/cs/apluscourses/ui/base/TextField.java
```diff
@@ -0,0 +1,27 @@
...
+import java.awt.event.KeyEvent;
+import javax.swing.JTextField;
+
+/**
+ * A class deriving from {@link JTextField} that works with the {@link TwoWayBindable} class to
+ * reflect its text input to the source property.
+ */
+public class TextField extends JTextField {
+  public final transient TwoWayBindable<TextField, String> textBindable =
+      new TwoWayBindable<>(this, TextField::setText, TextField::getText);
```
superseacat, line 14: @Taikelenn @jaakkonarhi @stellatsv 

please check this and ensure you understand what is happening here (in terms of the language and also in terms of logcs), ask if unclear

src/main/java/fi/aalto/cs/apluscourses/presentation/CourseSelectionViewModel.java
nikke234: > can you explain a bit why such an approach is chosen: why null, why return instead of throw, why ValidationError is an error (but not an exception at all)?

This reuses the validation approach that `OurDialogWrapper` uses. @OlliKiljunen is probably better than me at explaining the design decisions there - I personally like the consistency in that our dialogs all use the same validation mechanism.

> also, I'm curious what would happen if a) URL is incorrect, b) URL is formally (syntactically) correct, but there's no course config file at "that end"?

An incorrect URL doesn't pass the validation, so for a) the "OK" option of the dialog "refuses to work". For b), the dialog will happily accept the URL, but then a notification is shown that something is wrong with the "other end" of the URL. It would maybe be nice to check the course configuration already in the dialog, but that would be quite a heavy operation, and the course configuration is already checked in the course project action later, so this keeps the change smaller.

OlliKiljunen: This is about validating user input. `null` means no validation errors -> input is valid. It is customary not to use exceptions in user input validation (as its kinda expected to be invalid).

If URL is incorrect this returns non-null value causing validation error message to be shown to the user. Whether or not URL points to a correct document is out of this class' scope as that can only be known after the request over network is made and at that time, the form this view model is associated with, is already gone.

superseacat:
please update TESTING.md

src/main/java/fi/aalto/cs/apluscourses/presentation/CourseSelectionViewModel.java
superseacat: fiddling with the error messages a bit more I see it works nicely:

```log
13.21	The A+ Courses plugin encountered a network error: Please check your network connection and try again. Error message: 404 Not Found.

13.23	Failed to parse the course configuration file
			The A+ Courses plugin for IntelliJ could not set up the target course, perhaps due to a mistake by the course staff.
			As a result, many features of the plugin won't work as expected. If the problem persists, please contact the course staff.
			Error message: Course configuration file should consist of a valid JSON object.`

src/main/java/fi/aalto/cs/apluscourses/ui/base/TextField.java
OlliKiljunen: I could try to explain this a bit.

So we use Model-View-ViewModel MVVM patttern. That is, we have model classes (that represent our domain objects) and views (pieces of UI) just as in perhaps more familiar MVC pattern. Like in MVC, models should concern only in business logic and be totally ignorant of everything related to presentation of their data. On the other hand, views should be as dumb as possible in the sense that no logic should be located in view classes.

In MVC, we have controllers that work between models and views and take care of the presentation logic. In MVVM, we have view models (a.k.a presentation models) in the place of controllers. View models are, well, models for views: they can be thought as abstract representations of views. For example. if there is a view for personal info and there is text field for person's name then PersonalInfoViewModel has a (string) property "name" whose value is the content of the text field. So, the program doesn't "read" the person's name from the text field but a corresponding property in the view model.

Now, presentation logic (how data is shown, input validation, and so on) can be done in view models that are plain old java classes, and thus, totally unit testable. As said, it's "a representation of the UI" but without any UI framework or widgets messing thing up.

All that a view has to do is to keep its widgets in sync with the view model's properties. This keeping in sync we call _binding_. So we make UI elements' properties (such as text fields' text) bindable in a way presented in this code chunk. Then, in personal info view (in our hypothetical example), we create a binding by calling `nameTextField.bindToSource(viewModel.name)` and that's it (`viewModel.name` is an `ObservableProperty<String>` that is our implementation of observer pattern). Now the view model's "name" property is synced with view's "nameTextField"'s text content so that when one of them changes (either by UI input or by some logic in code), that change is immediately propagated to the other. Neat!

Please feel free to discuss and challenge this approach of ours. And don't hesitate to ask me if there's something that is difficult to understand (I bet there is). If you are not familiar with MVVM pattern and wanna search web there're quite a lot written about it (search either "MVVM" or "Presentation Model". Note that there's a variety in the details of the pattern).

superseacat (approved):
GoodStuff! 🤩

src/main/java/fi/aalto/cs/apluscourses/ui/courseproject/CourseSelectionView.java
nikke234: Agreed, and added a note about this, but merging for now, as we need to release soon

PR #467, opened 2021-01-13
Fix (once again) ordering of assignments

nikke234:
# Description of the PR

The new ordering works well with O1 and Studio 2, and it should work reasonably well with most courses. 

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseGroupViewModel.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseGroupViewModel.java
@@ -4,8 +4,6 @@ import fi.aalto.cs.apluscourses.model.ExerciseGroup;
 import fi.aalto.cs.apluscourses.presentation.base.SelectableNodeViewModel;
 import fi.aalto.cs.apluscourses.utils.APlusLocalizationUtil;
 import java.util.Comparator;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.jetbrains.annotations.NotNull;
 
@@ -20,7 +18,7 @@ public class ExerciseGroupViewModel extends SelectableNodeViewModel<ExerciseGrou
         .values()
         .stream()
         .map(ExerciseViewModel::new)
-        .sorted(EXERCISE_COMPARATOR) // O1_SPECIFIC
+        .sorted(EXERCISE_COMPARATOR)
         .collect(Collectors.toList()));
   }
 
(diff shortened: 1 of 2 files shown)


superseacat:
me, me, don't forget about me!

https://github.com/Aalto-LeTech/intellij-plugin/blob/sorting/src/test/java/fi/aalto/cs/apluscourses/integration/ApiTest.java

nikke234:
> me, me, don't forget about me!
> 
> https://github.com/Aalto-LeTech/intellij-plugin/blob/sorting/src/test/java/fi/aalto/cs/apluscourses/integration/ApiTest.java

I agree that these need to be improved, but that's a separate ticket :smile:

superseacat:
> I agree that these need to be improved, but that's a separate ticket

I'd suggest ensuring the hierarchical names are there and match with expected (inside existing or as a separate) case

nikke234:
> > I agree that these need to be improved, but that's a separate ticket
> 
> I'd suggest ensuring the hierarchical names are there and match with expected (inside existing or as a separate) case

This doesn't use the hierarchical names after all, so it's not needed

superseacat (approved):
nice!

nikke234:
Although now that you mention that, it _might_ be that with hierarchical name we could just sort lexicographically and it would always work, but we don't currently get the hierarchical name from the API, so let's make that a future consideration :+1:

superseacat (changes requested):
integration tests can still be there... otherwise nice job! 👍🏻

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseGroupViewModel.java
```diff
@@ -20,69 +18,44 @@ public ExerciseGroupViewModel(@NotNull ExerciseGroup exerciseGroup) {
...
-    if (chapter2 > chapter1) {
-      return -1;
+    /*
+     * Sorting lexicographically by the HTML URL almost works, except for the assignment numbers,
+     * since "...w01_ch04_9/" > "...w01_ch04_10/". So we compare the HTML URLs without the
+     * assignment numbers, and if they aren't equal, then there is a clear "winner". If they are
+     * equal, we compare the HTML URLs by length first, so that "...10/" becomes greater than
+     * "...9/". This conveniently also makes feedback exercises the last exercises in their
+     * chapters.
+     */
```
OlliKiljunen, line 38: Are those codes like `w01_ch04_9` generated by A+ (and kinda part of the "protocol") or are they chosen by teachers of the courses? If the latter, this is still a solution specific to certain courses only, right?

I was wondering if something like [natural sort](https://blog.codinghorror.com/sorting-for-humans-natural-sort-order/) would give a generic solution to the problem. Probably an overkill, though.

OlliKiljunen (approved):
Great, approved! 👍 Just one question I wasn't sure of.

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseGroupViewModel.java
nikke234: Yeah that is a good question, which I've been thinking about as well :smile: I'll try to dig deeper.

Natural sort is cool, although it would still stumble on feedback assignments I think :thinking:

superseacat (approved):

Issue #165, opened 2020-05-18
About A+ Plugin dialog improvement


superseacat:
+ [ ] **version number**
+ [x] Should be renamed ("A+ Courses")
+ [ ]  Layout (look and feel) of the dialog should be improved: margins and text alignment + logo
+ [x]  Hyperlink to somewhere else: external to 
  + [ ]  A+'s own website - Aalto A+ (like the LMS one)/
  + [ ] A+ general pages/
  + [ ]  JB's plugin page)
+ [ ] Authors' names
+ [ ] Check required 3rd party library attributions

superseacat:
**j**: low priority, but can be done

superseacat:
almost done

Issue #403, opened 2020-09-30
Opening a browser doesn't work in some (not officially supported) platforms


OlliKiljunen:
Opening a browser from a link in the event viewer (probably elsewhere too) causes error in some non-desktop environments. The links should be handled in the same way as IntelliJ's own "Open in browser" actions.

superseacat:
prefer Intellij's `BrowserUtil.browse()` as in https://github.com/Aalto-LeTech/intellij-plugin/pull/470#discussion_r557126840

PR #470, opened 2021-01-13
About dialog

stellatsv:
# Description of the PR
This relates to #165 . It contains improvements to the About dialog. Some extra adjustments may be needed to accommodate the new logo. 

So far it looks like this:

about

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


nikke234:
Nice! I really like the new look and additions, and the version is also very well highlighted, which is good. The screenshot is also nice, so I can skip the effort of actually running the code myself :smile: 

I'll add more specific code comments later, but a few general notes:

- Checkstyle complains about `.*` imports, but most likely IntelliJ creates those automatically for you whenever you save a file. To disable that behavior, see `File` -> `Settings...` -> `Editor` -> `Code Style` -> `Java` -> `Imports`.
-  Checkstyle has some other complaints as well, running the gradle "check" task is a good idea to avoid those (in IntelliJ's gradle window it's under "verification", or just `./gradlew check` on the command line.
- It could be nice to have symmetry in the inner padding of the window. That is, there would be the same amount of space between the left end of the text and the left edge of the window, and the right end of the text and the right edge of the window. However, that will probably significantly change when that logo is changed to the bigger new logo, so let's not worry about that for now.
- Your "catch & print" approach with `IOException` is correct: it doesn't need any special handling, since something is already badly wrong if that happens. However, instead of printing the stack trace, we use logging for these situations, which also works nicely with the IntelliJ UI. I'll add an example later, but searching for `logger.error(` somewhere in our code should given an idea of how it's done.

Overall great work, and I'm very happy to see how well you've gotten up to speed with our code and practices, etc. :+1:

src/main/resources/resources.properties
```diff
@@ -125,7 +125,12 @@ ui.authenticationView.noEmptyToken=The token cannot be empty
...
 
 ### Misc
-ui.aboutDialog.description=<html>The A+ Courses plugin supports the educational use of IntelliJ (and its Scala plugin) in<br>programming courses that rely on the A+ learning management system, which has been<br>developed at Aalto University. The plugin accesses programming assignments<br>and automated grading services provided by A+ and otherwise enhances<br>the student experience.</html>
-ui.aboutDialog.title=About the A+ Courses Plugin
+ui.aboutDialog.description=<html>The A+ Courses plugin supports the educational use of IntelliJ (and its Scala plugin)<br>in programming courses that rely on the A+ learning management system, which<br>has been developed at Aalto University.<br><br>The plugin accesses programming assignments and automated grading services<br>provided by A+ and otherwise enhances the student experience.</html>
+ui.aboutDialog.title=A+ Courses
 ui.aboutDialog.website=A+ Courses Plugin website
 ui.localHistory.submission.tag={0}, {1}, Submission {2}
+ui.aboutDialog.version=<html><b>Version: {0}</b></html>
+ui.aboutDialog.authors=<html>Developed by the department of Computer Science at Aalto University.</html>
```
nikke234, line 133: The `<html>` tags are most likely unnecessary in this line, since there's no styling like `<b>` or anything

src/main/resources/resources.properties
```diff
@@ -125,7 +125,12 @@ ui.authenticationView.noEmptyToken=The token cannot be empty
...
+ui.aboutDialog.APlusWebsite=A+ website
+ui.aboutDialog.GithubWebsite=A+ Courses Plugin Github
```
nikke234, line 135: This is a nitpicky minor detail, but since this is our "very official and professional" about window, let's capitalize the 'H' in GitHub

src/main/resources/resources.properties
```diff
@@ -125,7 +125,12 @@ ui.authenticationView.noEmptyToken=The token cannot be empty
...
+ui.aboutDialog.attributes=<html><div style='font-size:8px'>3<sup>rd</sup> Party Attributions:<ul><li><a href="https://commons.apache.org/proper/commons-io/">Apache Commons IO</a></li><li><a href="https://plugins.jetbrains.com/plugin/1347-scala">IntelliJ Scala Plugin</a></li><li><a href="https://www.json.org/json-en.html">json.org</a></li><li><a href="https://www.scala-lang.org/api/current/">Scala Standard Library 2.13.4</a></li><li><a href="https://github.com/srikanth-lingala/zip4j">zip4j</a></li></ul></div></html>
```
nikke234, line 136: I would suggest "The plugin uses the following 3<sup>rd</sup> party libraries:"

src/main/java/fi/aalto/cs/apluscourses/ui/AboutDialog.java
```diff
@@ -24,27 +29,47 @@ private AboutDialog() {
...
         getText("ui.aboutDialog.title"),
-        JOptionPane.INFORMATION_MESSAGE);
+        JOptionPane.INFORMATION_MESSAGE, icon);
   }
 
   @NotNull
-  private static JPanel createAboutInnerContainer() {
+  private static JPanel createAboutInnerContainer() throws IOException {
     JPanel panel = new JPanel();
     panel.setLayout(new BoxLayout(panel, BoxLayout.Y_AXIS));
```
OlliKiljunen, line 44: I like this way of building the view. We have elsewhere used mainly IntelliJ's visual GUI designer but it's a bit clumsy and XML it produces is messy. This is more structured and VCS friendly way.

I see no remarkable problem in having different view-building approaches in our code base; they're after all just implementation details of each view class. :+1:

src/main/resources/resources.properties
nikke234: The existing link also had "Plugin" with a uppercase 'p', but to be consistent with the first paragraph "The A+ Courses plugin...", I suggest we change to lowercase 'p' in both of the links. (_very_ minor detail again, but I can't help myself)

src/main/java/fi/aalto/cs/apluscourses/ui/AboutDialog.java
```diff
@@ -58,11 +83,52 @@ private static JPanel createAboutInnerContainer() {
...
+    linkLabel.setIconTextGap(0);
+    linkLabel.setHorizontalTextPosition(SwingConstants.LEFT);
+
+    return linkLabel;
+  }
+  @NotNull
+  private static JBLabel createVersionTextLabel() throws IOException {
+    JBLabel label = new JBLabel();
+    Properties properties = new Properties();
+    properties.load(AboutDialog.class.getResourceAsStream("/build-info.properties"));
```
nikke234, line 115: I wonder if it's possible to use the existing `BuildInfo` class here (`src/main/java/fi/aalto/cs/apluscourses/utils/BuildInfo.java`), something like this (untested):

```
Version version = BuildInfo.INSTANCE.version;
```

`BuildInfo` also takes care of the IOException, which would be nice, since this code wouldn't have to worry about it and the `throws` could be removed

src/main/resources/resources.properties
superseacat, line 136: as the lines' lengths are growing longer and longer, I suggest we start breaking lines. How about using the checkstyle's length of:
```xml
  <module name="LineLength">
            <property name="max" value="100"/>
            ...
   </module>
```
❓❓❓

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/AboutAction.java
```diff
@@ -3,12 +3,17 @@
...
 
 public class AboutAction extends DumbAwareAction {
 
   @Override
   public void actionPerformed(@NotNull AnActionEvent e) {
-    AboutDialog.display();
+    try {
+      AboutDialog.display();
+    } catch (IOException ioException) {
+      ioException.printStackTrace();
```
superseacat, line 16: I suggest considering if this is the correct level to handle the exception. In other words, perhaps it should be handled in the `display()` method already (or even deeper)?

superseacat: However, if you wish to still handle it here, why don't you try the:
```java
//top level declaration
private static final Logger logger = LoggerFactory.getLogger(AboutAction.class);

//instead of ioException.printStackTrace();
logger.error("An exception was thrown in an asynchronous call", ioException);
```

src/main/java/fi/aalto/cs/apluscourses/ui/AboutDialog.java
```diff
@@ -58,11 +83,52 @@ private static JPanel createAboutInnerContainer() {
     return linkLabel;
   }
 
+  @NotNull
+  private static LinkLabel<Object> createAPlusWebsiteLinkLabel() {
+    LinkLabel<Object> linkLabel = new LinkLabel<>(
+            getText("ui.aboutDialog.APlusWebsite"),
+            AllIcons.Ide.External_link_arrow,
+            (first, second) ->
+                    BrowserUtil.browse(A_PLUS_PAGE));
```
superseacat, line 92: question to @Aalto-LeTech/idea-plugin-dev:
was this something you meant when discussing the #403

src/main/java/fi/aalto/cs/apluscourses/ui/AboutDialog.java
```diff
@@ -1,21 +1,26 @@
...
+
+import java.io.IOException;
+import java.util.Properties;
 import org.jetbrains.annotations.NotNull;
 
 public class AboutDialog {
 
   private static final String A_COURSES_PLUGIN_PAGE = "https://plugins.jetbrains.com/plugin/13634-a-courses";
+  private static final String A_PLUS_PAGE = "https://plus.cs.aalto.fi/";
+  private static final String GITHUB_PAGE = "https://github.com/Aalto-LeTech/intellij-plugin";
```
superseacat, line 23: now when there's more of them, I wonder, if we should also externalize the URLs around the code into the `resources.properties` (as one of the features of Java props is that you can replace them without recompiling the code - not very useful for us but say Otto or some other teacher might really benefit from this when testing)? @Aalto-LeTech/idea-plugin-dev

superseacat:
> about

Nit: but I wonder, if it's possible to align all the paragraphs' text in the same manner, say `text-align: justify;` or something, and perhaps have equal margins?

superseacat:
Nit: the About window is mentioned in the [TESTING.md](https://github.com/Aalto-LeTech/intellij-plugin/blob/test/TESTING.md#3-about-window), so you may wanna add something there (or not, totally optional)

superseacat:
please link the GH issue to the PR as [described](https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#manually-linking-a-pull-request-to-an-issue) 😉

superseacat:
great code, almost ready to go! 🤩

superseacat:
there are basically two approached to writing (G)UIs' (Swing/Vaadin) code: one is more traditional, like in this PR, and has certain benefits like full control over the process and then there's UI-designer based, which hides some complexity and produces XML code. Both are welcome!

Examples:

- [Vaadin Designer](https://vaadin.com/docs/designer/getting-started/designer-installing-idea.html)
- [Swing Designer](https://www.jetbrains.com/help/idea/designing-gui-major-steps.html)

I strongly suggest you to try both approaches to see which of them is the best for you! 🦜

src/main/resources/resources.properties
OlliKiljunen: That would require change in properties format (XML or [mprops](https://github.com/mprops/mprops-java) or something). This format doesn't support multiline values.

src/main/java/fi/aalto/cs/apluscourses/ui/AboutDialog.java
nikke234: Yes, and `BrowserUtil.browse` worked, whereas the standard Java option of `Desktop.browse` did not, although that student had some obscure setup so we don't need to be very concerned with that issue

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/AboutAction.java
OlliKiljunen: Exception handling can probably be avoided altogether; see @nikke234's comment: https://github.com/Aalto-LeTech/intellij-plugin/pull/470#discussion_r556722429

src/main/java/fi/aalto/cs/apluscourses/ui/AboutDialog.java
OlliKiljunen: I agree we should. Perhaps not in the same resource file though with the UI strings (also because strings can be localized in the future).

src/main/java/fi/aalto/cs/apluscourses/ui/AboutDialog.java
stellatsv: You're right, it works fine and there is no need to handle a possible IOException.

src/main/java/fi/aalto/cs/apluscourses/ui/AboutDialog.java
nikke234: I am lightly in favor of the status quo (keeping them in the code), but I don't have any strong opinions this

src/main/resources/resources.properties
stellatsv: Well spotted! :smile:

src/main/java/fi/aalto/cs/apluscourses/ui/AboutDialog.java
OlliKiljunen: In any case, there's no need to address that in this PR. This is fine for now.

nikke234:
Great! I'll "undraftify" this, since I think this is ready and I'm curious to see what sonar has to say

nikke234 (approved):
So looks like sonar is off the table ATM, so no need to consider its feedback. Great stuff and merge-ready as far as I'm concerned :+1: Feel free to explore the logo+margin issue if you want to ofc

superseacat (approved):

Issue #246, opened 2020-07-24
Improve file list in submission dialog


nikke234:
- [ ] timestamp in a format "last modified 1 minute ago"
- [ ] the bullet points of the submitted files list in the submissions dialog are ever so slightly cut off from the left.

superseacat:
![2021-01-08_15-04](https://user-images.githubusercontent.com/9991098/104018669-dee53280-51c2-11eb-9b1d-746a9aedcfd9.png)
here!

PR #472, opened 2021-01-14
Display submitted files' last modification dates

Taikelenn:
# Description of the PR
Related to issue #246.

The submission window now displays how much time has passed since each file's last modification date.
untitled

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


src/main/java/fi/aalto/cs/apluscourses/utils/FileDateFormatter.java
```diff
@@ -0,0 +1,86 @@
...
+          ChronoUnit.DAYS,
+          ChronoUnit.MONTHS,
+          ChronoUnit.YEARS
+  );
+
+  private static String formatWithTimeUnit(ZonedDateTime fileTime,
+                                           ZonedDateTime currentTime,
+                                           ChronoUnit timeUnit) {
+    long timeUnitCount = timeUnit.between(fileTime, currentTime);
+    String timeUnitText = timeUnit.toString().toLowerCase(Locale.ROOT);
```
nikke234, line 30: Interesting, didn't know about this before :+1:

src/main/java/fi/aalto/cs/apluscourses/utils/FileDateFormatter.java
```diff
@@ -0,0 +1,86 @@
...
+import java.nio.file.attribute.FileTime;
+import java.time.Clock;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
+
+public class FileDateFormatter {
```
nikke234, line 15: I think there are two things going on here: formatting the date and getting the last modification time of a file. I could easily see the date formatting being useful in general also for other dates that don't come from files.

I would prefer it if we had a generic `DateFormatter` that does the formatting (regardless of whether the date comes from a file or something else). The `getFileModificationTime` would then either be a convenience method of `DateFormatter`, or `Files.getLastModifiedTime(...)` could be moved to `SubmissionViewModel`, or something else could work as well.

src/main/java/fi/aalto/cs/apluscourses/utils/FileDateFormatter.java
```diff
@@ -0,0 +1,86 @@
...
+
+    return timeUnitCount + " " + timeUnitText + " ago";
+  }
+
+  private static String formatWithLargestTimeUnit(ZonedDateTime fileTime, Clock currentTimeClock) {
+    ZonedDateTime currentTime = ZonedDateTime.now(currentTimeClock);
+
+    if (ChronoUnit.SECONDS.between(fileTime, currentTime) <= 0) {
+      return "just now";
+    }
```
nikke234, line 44: Does this mean what I think that it means? :smile: As soon as one full second has passed "just now" isn't shown anymore?

OlliKiljunen (commented):
Very good, I like this a lot.

Two things I'm thinking about:

1)

**FileDateFormatter** is really cool but a "problem" I see with it is that it has two somewhat separate responsibilities: getting information of a file's modification date, and, on the other hand, calculating a difference between two time points.

I'd suggest there would be a class that is only about the latter, that is, given two `ZonedDateTime` objects it would return their difference in that "X chronounits" format. This could be achieved either by A) moving the related logic in a new class, or, B) moving file-related code from `FileDateFormatter` to the caller method and renaming the class to something not file-related.

Now, the class could be unit-tested with arbitrary time points without having to mock up file system stuff.

2)

Preferably, strings visible in the UI would be read from `resources.properties` file. Here we have "modified", "ago", "just now" and then those time units. One possibility would be moving all but time units to resources and handle time units as you have done. Other possibility would be that there was string like "seconds|minutes|hours|days|months|years" (and maybe another like "second|minute|...") in the resources so the correct string could be achieved with an index from the array returned by `String::split`.

We don't currently have to support internationalization in the UI but less hard coded English the better.

--------

But all in all, nice code and design. I'm not against merging this as-is (even though I'd like to see some changes) because it still implements the feature it should, does it well, and doesn't degrade the code base by any means. Though, since this is still marked as draft, I don't approve yet.

src/main/java/fi/aalto/cs/apluscourses/utils/FileDateFormatter.java
OlliKiljunen: Yeah, I didn't see this comment but addressed the same thing in [my comment](https://github.com/Aalto-LeTech/intellij-plugin/pull/472#pullrequestreview-568367600).

src/main/java/fi/aalto/cs/apluscourses/utils/FileDateFormatter.java
OlliKiljunen: It could as well be "0 seconds ago" if that would make the logic simpler.

nikke234: Or the seconds option could perhaps also be removed and everything < 1 minute shown as "just now". A minor issue with seconds is that it makes it super obvious that the value isn't updated while the dialog is visible, since it constantly says "5 seconds ago"

But this is quite a small thing, so any of these options (including the current one) are quite fine with me

src/main/java/fi/aalto/cs/apluscourses/utils/FileDateFormatter.java
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * Same as {@link FileDateFormatter#getFileModificationTime(Path, Clock)} with the clock
+   * being always set to the system's current time.
+   *
+   * @see FileDateFormatter#getFileModificationTime(Path, Clock)
+   */
+
+  public static String getFileModificationTime(Path filePath) throws IOException {
+    return getFileModificationTime(filePath, Clock.system(ZoneOffset.UTC));
```
nikke234, line 63: This isn't a comment on the code per se, but just making sure I understand this correctly: is it correct that it wouldn't matter what zone offset you use here, since the file time is also converted to the time zone of the `currentTimeClock` parameter?

nikke234 (commented):
The implementation is very nice! I didn't initially expect this issue to be solvable in such a neat way :+1: The code is also tested, which is great

Looks like Olli and I independently arrived at the same conclusion about `FileDateFormatter`, so that's probably a good sign that it could be changed :smile: Other than that, good stuff

src/main/java/fi/aalto/cs/apluscourses/utils/FileDateFormatter.java
nikke234: My initial confusion arises from the fact that I would imagine it would be almost impossible for students to click "submit" so quickly that it would be less than a full second

OlliKiljunen: Good points. Though it sounds a bit ridiculous (and not necessarily look good in the UI), I think showing seconds is useful for a user. I'd guess it's most often just seconds between a student being ready with editing the file and them making a submission.

Imagine you're a student who writes some code and is making a submission. It shows "just now" (meaning less than a minute ago). Then you remember that you didn't remove your "debug print" so you press cancel, remove that line and press "submit" button again. It took you about 10 seconds. It still shows "just now". Was the file saved? Is the latest change included in the submission? (Okay this was stupid but, still, my point holds. 😛)

nikke234: Well, the file is definitely saved since we save every file in the project on assignment submission :wink: Of course the student probably doesn't know that...

But I don't actually think the main benefit of this feature is "was the file saved?". I would say the main benefit of this feature is preventing submissions to the incorrect assignment. So when a student selects the wrong assignment and sees "last modified 2 hours ago" even though they just worked on it, then it becomes obvious that something is wrong. Selecting the incorrect assignment has happened in practice BTW, and if I recall correctly, feedback from a student who did so was specifically what prompted us to consider adding this feature in the first place.

src/main/java/fi/aalto/cs/apluscourses/utils/FileDateFormatter.java
Taikelenn: Correct, no matter which time zone you use here, it will work the same. It's just that time zone is a mandatory parameter, so I went with the most "standard" one - UTC. Speaking of which, I just found out that there's already a wrapper function which does the same: `Clock.systemUTC()`

nikke234: Sounds good :+1: Might even add a comment like `// any time zone would work here`, my initial reaction to seeing UTC was "surely this can't be correct"

src/main/java/fi/aalto/cs/apluscourses/utils/FileDateFormatter.java
Taikelenn: To be honest, I was targeting some other obscure case when doing this check for `<= 0 seconds` - a possibility that the file date is actually slightly in the future compared to current system time 😛
There are so many possible filesystems and each of them is handling last modification time in their own way so I wouldn't be surprised if this behaviour would be sometimes possible (file data/metadata caching or something) - this check would avoid a "modified -1 seconds ago" situation
Anyway, there seem to be following possibilities:
- drop "just now" altogether in favour of "0 seconds/minutes ago"
- reserve "just now" for a larger span of time (5 seconds? 30 seconds? a minute?)
- keep it as is

nikke234: The prosecution and the defense have presented their cases, you can be the judge on which is the best option of the three

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/SubmissionViewModel.java
```diff
@@ -104,6 +108,25 @@ public String getSubmissionCountText() {
...
+   */
+  @NotNull
+  public String getFileInformationText(SubmittableFile file) {
+    StringBuilder fileInfoText = new StringBuilder(file.getName());
+    try {
+      String lastModificationTime =
+              FileDateFormatter.getFileModificationTime(filePaths.get(file.getKey()));
+      fileInfoText.append(" (modified ").append(lastModificationTime).append(")");
+    } catch (IOException e) {
+      // don't print anything about the last modification time if a very unlikely exception happened
```
superseacat, line 125:
```java
//top level declaration
private static final Logger logger = LoggerFactory.getLogger(AboutAction.class);

//instead of ioException.printStackTrace();
logger.error("An exception was thrown for some mysterious reason", e);
```

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/SubmissionViewModel.java
```diff
@@ -104,6 +108,25 @@ public String getSubmissionCountText() {
...
   }
 
+  /**
+   * Formats a descriptive string for the submission dialog about a submittable file. The string
+   * includes the file name and its last modification date.
+   *
+   * @param file The submittable file in question.
+   */
+  @NotNull
+  public String getFileInformationText(SubmittableFile file) {
```
superseacat, line 118: would this benefit from `...(@NotNull Submittable file)`?

superseacat, line 118: I suggest to consider even more descriptive name for a method, say `getFileNameAndLastModificationTimeText()` or something?

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/SubmissionViewModel.java
```diff
@@ -104,6 +108,25 @@ public String getSubmissionCountText() {
     return submissionCountText.toString();
   }
 
+  /**
+   * Formats a descriptive string for the submission dialog about a submittable file. The string
+   * includes the file name and its last modification date.
```
superseacat, line 113: the doc says "... last modification data", but is it really a date or rather a dateTime?

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/SubmissionViewModel.java
OlliKiljunen: I kinda disagree. The contract of this method is to return a human-readable description of the file to be shown in the file list of the dialog. It is this class' choice what exactly is shown there. A subclass could choose to return some other info (as long as it would satisfy the contract).

superseacat: agreed, my bad 🙄

nikke234 (approved):
Great! :+1:

src/main/java/fi/aalto/cs/apluscourses/utils/FileDateFormatter.java
```diff
@@ -4,80 +4,24 @@
...
 import java.nio.file.attribute.FileTime;
-import java.time.Clock;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
-import java.time.temporal.ChronoUnit;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Locale;
 
 public class FileDateFormatter {
```
OlliKiljunen, line 10: Nit: this class could be renamed to `FileUtils` or something to allow other utility functions to be added here in the future. As such as it is currently, the purpose of this class seems a bit questionable to me.

OlliKiljunen (approved):
Great work! 👍 Please, go and merge. It's nice we got this included in the next release! :shipit:

PR #476, opened 2021-01-18
Convert text to path

jaakkonarhi:
Fixes the text on the README logo


nikke234 (approved):

OlliKiljunen (approved):
Oh, so it never was intended to be serif font. :D I kinda liked that but I have to agree the actual (sans serif) is more professional looking. 👍 

image

PR #477, opened 2021-01-18
"Fix" exercise submittability

### LINKED ISSUE #436: something is fishy with the assignments SUBMITTABILITY in Round 15 - part A
almost sure, that some of these assignments (**Round 15 - part A) must be submittable**

assignments_ordering+

nikke234:
This fixes #436 by looking at `exerciseModules` from the course configuration to determine whether an assignment is submittable or not. This is only a temporary solution in lieu of a proper solution, that determines submittability from the individual exercise end point in the A+ API. I didn't want to add any course <-> exercise link to the the classes for a temporary solution, so to keep the impact minimal, I have added the information to the `Points` class, which is already available when constructing an `Exercise` object from the JSON corresponding to the exercise.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/model/Exercise.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/model/Exercise.java
@@ -26,6 +26,8 @@ public class Exercise {
 
   private final int maxSubmissions;
 
+  private final boolean submittable;
+
   /**
    * Construct an exercise instance with the given parameters.
    *
@@ -41,13 +43,15 @@ public class Exercise {
                   @NotNull String htmlUrl,
                   int userPoints,
                   int maxPoints,
-                  int maxSubmissions) {
+                  int maxSubmissions,
+                  boolean submittable) {
     this.id = id;
     this.name = name;
     this.htmlUrl = htmlUrl;
     this.userPoints = userPoints;
     this.maxPoints = maxPoints;
     this.maxSubmissions = maxSubmissions;
+    this.submittable = submittable;
   }
 
   /**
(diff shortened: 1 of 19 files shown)


superseacat (approved):
I'd merge and start the pre-S2 testing _asap_. Also, I suggest keeping/create afresh the issue for fixing the challenge properly, and, perhaps, we need a new endpoint on the A+ API side so as not to have to traverse each exercise endpoint every time?

nikke234:
Yes, will merge once more approves come. :+1: 

I'll create an issue for fixing it properly.

A new A+ end point probably isn't needed. We use the exercise end points anyways when making a submission, so it's just a matter of caching the results so that we don't need to do it all the time.

OlliKiljunen (approved):
This is fine. Architecturally, I see no fundamental problem with this solution (except the actual "meaning" of `Points` class should be re-thought) but I understand we don't ultimately want to make it teacher's responsibility to list submittable exercises in course config file.

But nicely implemented even though temporary. 👍

nikke234:
> This is fine. Architecturally, I see no fundamental problem with this solution (except the actual "meaning" of Points class should be re-thought) but I understand we don't ultimately want to make it teacher's responsibility to list submittable exercises in course config file.

I see putting the burden on the maintainer of the course config file as a fundamental problem with this approach, especially since the information is already available to us from the API. Architecturally I guess it's fine, but then we should refactor `exerciseModules` into a more general exercise metadata container, instead of just mapping exercises to modules.

Issue #273, opened 2020-07-30
Replace "Nothing to show" with a more informative text in the modules and assignments list


superseacat:
how about replacing the default "Nothing to show" with custom "Hey, you know what, this is empty for a reason, how about you turn the project into A+ Project (hyperlink to the corresponding action) to display some data?"

![turn_project_into_A_plus](https://user-images.githubusercontent.com/9991098/88907841-c8c16680-d261-11ea-86de-e48b3e165390.PNG)

superseacat:
alternatives are not showing the whole tab or a part of it until turned into an A+ project

superseacat:
`Label` instead of a `Tree` and that says whatever you want!

PR #490, opened 2021-01-22
Replace text

stellatsv:
# Description of the PR
The draft PR relates to #273. It contains functionality for triggering CourseProjectAction from Modules/Assignments view when the project is not an A+ project yet. 

There is some unstable behaviour which occurs on startup when the project that opens is already an A+ project. While the project is loading the default informative text is visible and the user can then click and trigger the CourseProjectAction. If the user closes/cancels that window then the Assignment tree is not always loaded and the default text persists (other times though everything works fine after closing/cancelling the window). In order to display the Assignments the user needs to restart IntelliJ which is not ideal. Any input on the issue is much appreciated!

There is always the fallback option of displaying the informative text on startup without the extra hyperlink functionality (it will still be an improvement).

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [x] Yes
- [ ] Not relevant


nikke234:
I'm not actually able to reproduce the unstable behavior :thinking: However, while starting the IDE, I got the following exception:

```
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'project' of com/intellij/openapi/project/impl/ProjectManagerImpl.addProjectManagerListener must not be null
        at com.intellij.openapi.project.impl.ProjectManagerImpl.$$$reportNull$$$0(ProjectManagerImpl.java)
        at com.intellij.openapi.project.impl.ProjectManagerImpl.addProjectManagerListener(ProjectManagerImpl.java)
        at fi.aalto.cs.apluscourses.intellij.services.PluginSettings.lambda$getMainViewModel$0(PluginSettings.java:164)
        at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1705)
        at fi.aalto.cs.apluscourses.intellij.services.PluginSettings.getMainViewModel(PluginSettings.java:161)
        at fi.aalto.cs.apluscourses.intellij.actions.InstallModuleAction.update(InstallModuleAction.java:55)
        at com.intellij.openapi.actionSystem.ex.ActionUtil.performDumbAwareUpdate(ActionUtil.java:178)
[rest omitted]
```

It kinda makes sense: `InstallModuleAction.update` gets the project from `AnActionEvent.getProject` and calls `PluginSettings.getMainViewModel` with that project. `PluginSettings` then passes that project parameter to `ProjectManager.addProjectManagerListener`, which marks project with `@NotNull`. However, `AnActionEvent.getProject` is marked as `@Nullable`, so it can return `null`.

Seems to be clearly a bug in our existing code. Interesting that it hasn't shown up before and that something here now makes it show up. The uncaught exception could well explain some of the unstable behavior that you are seeing.

I recommend reading the following introduction to actions (not everything there is relevant so skimming through it should do):
https://plugins.jetbrains.com/docs/intellij/basic-action-system.html

My alarm bells started ringing a bit when I was able to start the action from the link even during indexing, which shouldn't be possible for `CourseProjectAction`. With the link it's possible however, because the code manually creates a `CourseProjectAction`, which we shouldn't do, since we are now circumventing IntelliJ's action system, which manages actions and determines which are enabled, disabled, etc.

Later when I'm less busy I'll add an idea for how to implement it so that these issues are fixed. Btw as you might have noticed, working with the platform can be quite frustrating at times, but I hope that this hasn't been too discouraging :smile:

OlliKiljunen (commented):
Unfortunately, I've so far been unable to reproduce the exact problem you described.

At least, I can say, this error was thrown: ([full stacktrace](https://pastebin.com/BBqRfDQr))
```
update failed for AnAction(fi.aalto.cs.apluscourses.intellij.actions.InstallModuleAction) with ID=fi.aalto.cs.apluscourses.intellij.actions.InstallModuleAction

java.lang.IllegalArgumentException: Argument for @NotNull parameter 'project' of com/intellij/openapi/project/impl/ProjectManagerImpl.addProjectManagerListener must not be null
```
 It's barely related to the issue in question but it doesn't occur in master branch. 🤔

What I see suspicious, is adding and removing UI widgets from panes. It makes me wonder that the data context IntelliJ associates with them has wrong (null) project. Have you tried just making the widgets visible / invisible?*

I can take a closer look at this on Monday.

*) That shouldn't require re-validating the panes either. Ideally, the logic of what should be shown (list or text) would be located in the view model but let's look at that later if it feels necessary.

OlliKiljunen:
Ok I posted my comment without reading Nikolas' comment first. I agree on what he says about the exception thrown.

nikke234:
To run the action without creating a new instance of `CourseProjectAction` you could use our `ActionUtil.launch` method.

stellatsv:
The functionality is now stable. I used a CardLayout and displayed a different component with each card instead of adding/removing components. 

Something new that wasn't present in the previous commits: If a user turned the project into an A+ project but was not authenticated then the message about the A+ project was still displayed in the Assignments view. Now, an appropriate text is shown informing the user to set the A+ token. A possible improvement here would be to make this a hyperlink towards the action as well (I can create a new issue about that).

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -32,12 +58,26 @@ public JPanel getBasePanel() {
...
 
+  private void updateComponents() {
+    emptyText.setText(getText("ui.module.ModuleListView.turnIntoAPlusProject"));
+    emptyText.setHorizontalAlignment(SwingConstants.CENTER);
+    emptyText.setVerticalAlignment(SwingConstants.CENTER);
+    if (hasCourse) {
+      cl.show(cardPanel,"TreeCard");
+    } else {
+      cl.show(cardPanel, "LabelCard");
+    }
```
OlliKiljunen, line 78: This logic could be moved to view model (and unit tested) if there was a method in the view model which returns the name of the card to be shown. Then here it would simply be:
```java
cl.show(cardPanel, mainViewModel.getCardToBeShown());
```

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModulesView.java
```diff
@@ -47,9 +62,23 @@ public JPanel getBasePanel() {
...
+
+  private void updateComponents() {
+    emptyText.setText(getText("ui.module.ModuleListView.turnIntoAPlusProject"));
+    emptyText.setHorizontalAlignment(SwingConstants.CENTER);
+    emptyText.setVerticalAlignment(SwingConstants.CENTER);
+    if (moduleListView.getModel().getSize() == 0) {
+      cl.show(cardPanel, "LabelCard");
+    } else {
+      cl.show(cardPanel, "TreeCard");
+    }
```
OlliKiljunen, line 82: [Same](https://github.com/Aalto-LeTech/intellij-plugin/pull/490#discussion_r564764183) here.

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModulesView.java
```diff
@@ -47,9 +62,23 @@ public JPanel getBasePanel() {
...
+    },
         ModalityState.any()
     );
+
+  }
+
+  private void updateComponents() {
+    emptyText.setText(getText("ui.module.ModuleListView.turnIntoAPlusProject"));
+    emptyText.setHorizontalAlignment(SwingConstants.CENTER);
+    emptyText.setVerticalAlignment(SwingConstants.CENTER);
```
OlliKiljunen, line 77: Could this be moved to the constructor? (also in exercises)

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -1,26 +1,52 @@
...
+  private JLabel emptyText;
   private JPanel basePanel;
-
+  private JScrollPane pane;
   @GuiObject
   public JPanel toolbarContainer;
+  private JPanel cardPanel;
+  private CardLayout cl;
+  private MainViewModel mainViewModel;
+  private boolean hasCourse = false;
```
OlliKiljunen, line 35: If you follow my suggestions, you shouldn't need this private field. In any case, I think passing information via mutable private fields (heap) should be avoided, if it can be passed in arguments of method calls (stack).

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -1,26 +1,52 @@
...
 
-  public ExercisesView() {
-    // See ModulesView.java
+  /**
+   * Creates an ExerciseView that uses mainViewModel to dynamically adjust its UI components.
+   */
+  public ExercisesView(MainViewModel mainViewModel) {
+    this.mainViewModel = mainViewModel;
+    mainViewModel.exercisesViewModel
+            .addValueObserver(this, ExercisesView::viewModelChanged);
```
OlliKiljunen, line 43: Please, move this to the end of the ctor. `ObservableProperty::addValueObserver` calls immediately `this::viewModelChanged`, so it is possible that value for `cl` is not yet set when it is referenced in `updateComponents` (called by `viewModelChanged`).

I must apologize that I made you move that observer registration here; I now see it's a bit controversal to have it here because it "leaks" not-yet-constructed `this` (which is a commonly known source source of anomalies). It's however sage to have it here, as long as it's the last statement of the ctor (btw, you don't have to call `updateComponents` separately here, you can remove that line). Other possibility is that it's called from the caller of this ctor, as it was previously (sorry for confusion).

OlliKiljunen (commented):
Looks already really good! :+1: Very well solved previous issues. Some parts I would still change.

Looking forward to have this merge-ready!

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
OlliKiljunen: The following is by no means required, just an idea:

Should you like, you could write `CardPanel` class that inherits `JPanel` and has `getCurrentCardName` and `setCurrentCardName` methods. Then you could add `currentCardNameBindable` field to that class and bind it to the corresponding observable property in the view model. And tada: whenever that property changes, the correct card is automatically shown without any "update view" method calls.

If you want to try this, I'd be glad to give further help.

OlliKiljunen (commented):
Added one comment.

src/main/java/fi/aalto/cs/apluscourses/intellij/toolwindows/APlusToolWindowFactory.java
```diff
@@ -1,23 +1,32 @@
 package fi.aalto.cs.apluscourses.intellij.toolwindows;
 
+import com.intellij.ide.DataManager;
 import com.intellij.openapi.actionSystem.ActionGroup;
 import com.intellij.openapi.actionSystem.ActionManager;
 import com.intellij.openapi.actionSystem.ActionPlaces;
 import com.intellij.openapi.actionSystem.ActionPopupMenu;
 import com.intellij.openapi.actionSystem.ActionToolbar;
+import com.intellij.openapi.actionSystem.AnActionEvent;
```
nikke234, line 9: This import is unused now

src/main/java/fi/aalto/cs/apluscourses/intellij/toolwindows/APlusToolWindowFactory.java
```diff
@@ -1,23 +1,32 @@
...
+import com.intellij.openapi.actionSystem.DataContext;
+import com.intellij.openapi.actionSystem.Presentation;
```
nikke234, line 11: This import is unused now

nikke234:
Just noting, that eventually it would be nice to also have the hyperlinks styled like links, but I think that can be a separate issue. `LinkLabel` could be considered for that:

https://upsource.jetbrains.com/idea-ce/file/idea-ce-4b94ba01122752d7576eb9d69638b6e89d1671b7/platform/platform-api/src/com/intellij/ui/components/labels/LinkLabel.java

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -118,4 +121,17 @@ public ExercisesTreeViewModel getExercises() {
...
+   * and is not null.</p>
+   */
+  public String getCardToBeShown() {
+    if (Optional.ofNullable(courseViewModel.get())
+            .map(BaseViewModel::getModel)
+            .orElse(null) != null) {
+      return "TreeCard";
+    }
+    return "LabelCard";
+  }
```
nikke234, line 136: Nit: the cards are sort of an implementation detail of the UI, so it might make more sense to have methods like `isEmptyTextVisible` or `isListVisible`, but this isn't very important though

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
```diff
@@ -1,5 +1,7 @@
 package fi.aalto.cs.apluscourses.ui.base;
 
+import static fi.aalto.cs.apluscourses.utils.PluginResourceBundle.getText;
```
nikke234, line 3: Unused import

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModulesView.java
```diff
@@ -47,9 +64,12 @@ public JPanel getBasePanel() {
    * Update this modules view with the given view model (which may be null).
    */
   public void viewModelChanged(@Nullable CourseViewModel course) {
-    ApplicationManager.getApplication().invokeLater(
-        () -> moduleListView.setModel(course == null ? null : course.getModules()),
+    ApplicationManager.getApplication().invokeLater(() -> {
+      moduleListView.setModel(course == null ? null : course.getModules());
+      cl.show(cardPanel, (course != null) ? "TreeCard" : "LabelCard");
```
nikke234, line 69: This should probably also use `MainViewModel.getCard...` (unless the method is changed, as I mentioned in another comment)

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -1,26 +1,53 @@
...
   private TreeView exerciseGroupsTree;
+  private JLabel emptyText;
   private JPanel basePanel;
-
+  private JScrollPane pane;
   @GuiObject
   public JPanel toolbarContainer;
+  private JPanel cardPanel;
+  private CardLayout cl;
+  private MainViewModel mainViewModel;
```
nikke234, line 34: The fact that `ExercisesView` has a reference to the `MainViewModel` is a bit questionable IMO, which now makes me think that the "`ExercisesTreeViewModel` is never null" might have been a cleaner approach. Of course, I have the benefit of hindsight here, so I'm not saying that I knew this beforehand :smile: 

Anyways, I don't this needs to be changed at the moment. We can consider later if this is something that we want to modify.

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -1,26 +1,53 @@
...
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.ModalityState;
 import fi.aalto.cs.apluscourses.intellij.actions.ActionUtil;
 import fi.aalto.cs.apluscourses.intellij.actions.OpenSubmissionAction;
+import fi.aalto.cs.apluscourses.presentation.MainViewModel;
 import fi.aalto.cs.apluscourses.presentation.exercise.ExercisesTreeViewModel;
 import fi.aalto.cs.apluscourses.ui.GuiObject;
 import fi.aalto.cs.apluscourses.ui.base.TreeView;
+
+import java.awt.CardLayout;
```
nikke234, line 14: Cool! I've worked so rarely with pure swing, that I had no idea what this was, but always happy to learn new things during reviews :smile:

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -61,7 +62,9 @@ private void updateExercises() {
         .orElse(null);
     Authentication auth = authentication.get();
     if (course == null || auth == null) {
-      exercisesViewModel.set(null);
+      ExercisesTreeViewModel emptyModel = new ExercisesTreeViewModel(new ArrayList<>(),
+              new Options());
+      exercisesViewModel.set(emptyModel);
```
nikke234, line 67: Hmm, in another comment I noted that this could be an alternative to the current approach of passing `MainViewModel` to the ctor, since `viewModelChanged` would then always receive a non-null parameter.

But with the current approach, I think this change isn't needed anymore, since I think there's no problem if `exercisesViewModel` is null instead, as it used to be

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -1,26 +1,53 @@
...
     basePanel.putClientProperty(ExercisesView.class.getName(), this);
+    cl = (CardLayout) cardPanel.getLayout();
+    exerciseGroupsTree.getEmptyText().appendLine(getText("ui.exercise.ExercisesView.setToken"));
+    exerciseGroupsTree.getEmptyText().appendLine(
+            getText("ui.exercise.ExercisesView.setTokenDirections"));
+    emptyText.setText(getText("ui.module.ModuleListView.turnIntoAPlusProject"));
+    emptyText.setHorizontalAlignment(SwingConstants.CENTER);
+    emptyText.setVerticalAlignment(SwingConstants.CENTER);
+    mainViewModel.exercisesViewModel
+            .addValueObserver(this, ExercisesView::viewModelChanged);
```
OlliKiljunen, line 50: Sorry for once more changing my mind on this, but perhaps, now when I think of it, this would suit best to the caller method. 😞 Totally my bad. But it's that not-yet-constructed  `this` that concerns me even though it's okay here (but thinking for the future).

I only have cheek to ask this (after first saying it's fine to have it here) because I would otherwise ask you to change `viewModelChanged` private, so in any case, there would be an urge for some change. ;)

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -32,8 +59,10 @@ public JPanel getBasePanel() {
    * Sets the view model of this view, or does nothing if the given view model is null.
    */
   public void viewModelChanged(@Nullable ExercisesTreeViewModel viewModel) {
-    ApplicationManager.getApplication().invokeLater(
-        () -> exerciseGroupsTree.setViewModel(viewModel),
+    ApplicationManager.getApplication().invokeLater(() -> {
+      exerciseGroupsTree.setViewModel(viewModel);
+      cl.show(cardPanel, mainViewModel.getCardToBeShown());
+    },
         ModalityState.any()
```
OlliKiljunen, line 66: Nit: indention

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModulesView.java
OlliKiljunen: I agree that this should use the same method as ExercisesView. It would require passing MainViewModel to this class also but that's nice symmetry.

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
OlliKiljunen: I agree it's a bit questionable and I agree it's an issue that doesn't have to be solved now.

Basically, how this could be solved, is putting `getCardToBeShown` in an interface (implemented by MainViewModel) and use that interface here.

Another option would be doing something like I suggested [here](https://github.com/Aalto-LeTech/intellij-plugin/pull/490#discussion_r564794544).

nikke234 (commented):
Very nice! This feature is a major usability improvement, so I'm glad to see it done :+1: 

Two things that I would like to change before approving:
  + Remove the change that makes `exercisesViewModel` be empty instead of `null`. The Slack discussion was pretty confusing, but the suggestions that Olli and I made were alternatives, so only one of them should be implemented.
  + Replace `getCardToBeShown` with a method that isn't directly linked to the implementation details of the UI (view).

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -118,4 +121,17 @@ public ExercisesTreeViewModel getExercises() {
...
   }
+
+  /**
+   * <p>Returns the name of the Card to be shown based on whether a Course object exists
+   * and is not null.</p>
+   */
+  public String getCardToBeShown() {
+    if (Optional.ofNullable(courseViewModel.get())
+            .map(BaseViewModel::getModel)
+            .orElse(null) != null) {
```
OlliKiljunen, line 132: It should be enough to test that `courseViewModel.get() != null`. `BaseViewModel::getModel` is `@NotNull`.

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModulesView.java
nikke234: Sorry nevermind, the modules view of course doesn't have a reference to the main view model

OlliKiljunen (changes requested):
Really good code! Also, thanks for getting our previous ideas into consideration.

I left a few comments, minor issues mostly, though I'd like to see changes at least [here](https://github.com/Aalto-LeTech/intellij-plugin/pull/490/files#r566025688).

Other thing that I'd like to see: a unit test for `getCardToBeShown`.

> Replace getCardToBeShown with methods that aren't directly linked to the implementation details of the UI (view).

@nikke234 Can we still discuss this? As I mentioned in a comment, I disagree this should be changed to boolean value methods. That would only add logic in the view. Strings as keys for what to show are abstract enough!

@stellatsv If you end up following @nikke234's suggestion on this, I'd like to have unit tests for those methods. Then, there would be a method in the view that solves the name of the card from the values returned by those boolean methods. That method, could also be unit tested but it doesn't have to as we don't usually test view classes (as there shouldn't be any logic there!)

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
nikke234: Unless I misunderstood something, I still think it belongs in `ExercisesTreeViewModel` instead.

An interface that main view model implements would fix the broken encapsulation here, but I think the more correct solution would be to not pass main view model to the ctor at all.

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModulesView.java
OlliKiljunen: > Sorry never mind

No! Please mind this. 😂

@nikke234 Okay it seems we have a bit of a disagreement. 😘

@stellatsv I'd like to hear what do you think.

nikke234: @OlliKiljunen Look at the comment times! :smile: I posted mine before your review, but confusingly it's shown after your comment.

In other words, I changed my mind on my own comment before seeing your comment at all.

That having been said, I would not introduce main view model to `ModulesView` (even though it is symmetric), since I think it also doesn't belong to `ExercisesView` :smile:

nikke234:
> @nikke234 Can we still discuss this? As I mentioned in a comment, I disagree this should be changed to boolean value methods. That would only add logic in the view. Strings as keys for what to show are abstract enough!

@OlliKiljunen I don't think it would actually add any logic to the view, since the logic is already in the view. It just happens to be hidden in a card layout, but if you try to write the view without using card layout, then you will find yourself comparing strings in the view

nikke234:
...and if the view code would look like
```
if (cardToBeShown.equals("...")) {
  // show something
} else if (cardToBeShown.equals("...")) {
  // show something else
}
```

then surely you would prefer booleans as well :smile:

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
OlliKiljunen: Yeah, you're right. Or better yet, there should probably be a class between `MainViewModel` and `ExercisesTreeViewModel` so the composition would be `[MainViewModel]<>1---1[ThatClass]<>0..1---1[ExercisesTreeViewModel]`.

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
nikke234: A good point in that the strings don't have to refer to a card layout, but can instead be seen as just some descriptions of what is shown.

Strings are quite brittle for this though, since you rely on magic constants that both the view model and view have to understand, and the method has to return a specific subset of strings, even though there are infinitely many strings.

In this case it works quite neatly, but that is only because the view uses a card layout (implementation detail)

OlliKiljunen: It's just a Java way to use constants, either integers or strings, to pass "enum" values (actual enums being not that nice to work with). Using either string or int valued constants would make sense in this class's interface, strings perhaps a bit more because there's no order between them.

Similarly, I think that the designers of Swing made a good choice when having CardLayout use string keys (instead of numeric indexing). I guess they had this kind of usage in mind.

About those boolean methods: when thinking those two methods as one `MainViewModel -> Boolean * Boolean` valued function, we see that only values `(true, false)` and `(false, true)` make sense. `(true, true)` (i.e. show both) and `(false, false)` (show neither) are "nonsensical". ;)

nikke234: Yeah good point that half of the boolean pairs are nonsensical as well :+1: (I also considered enums because of that, but as you note, they aren't that nice to work with)

CardLayout using string vs indexing however is orthogonal here, because indices have the same issue as strings here, and as such, I would not suggest using them here

nikke234: In this sense a single `isEmpty` in `ExercisesTreeViewModel` would be the most correct here, since there are exactly two possibilities. That would sort of add some logic to the view, but as I mentioned below, it could be argued that the logic is already there

nikke234: Both modules and exercises view actually seem to show the same empty text. Exercises view has the addition that if the tree is shown, but the shown tree is empty, then it has its own empty text, which is done using intellij's `Tree`.

So from that perspective, actually the most correct would be a single boolean function like `showEmptyText`, which would then always return a sensical value. This would add some logic to the view though (could be argued that the logic is there already)

OlliKiljunen: I like your reasoning and I'm ready to give up on this issue if you are certain of your viewpoint. 😄 It's not a big issue after all. 

I'd like to remind, though, that methods can have pre- and post-conditions stronger than just type-correctness. "Enum-like" string and int consts are a typical example of this. Other examples include `String::charAt` that never accepts negative integers and `String::indexOf` that never returns value lower than -1.  Types were never meant to specify functions' _domain_ and _range_ to the full extent. What a function can take and return, must be specified in its documentation.

nikke234: I for sure very much like using the static type system to make violations of preconditions impossible (or more accurately, compile-time errors). Similarly, in many languages indexing operations return some sort of optional type, some languages take it even further and define integer division to return an optional type due to division by zero. But anyways, not to derail this too much...

@stellatsv The options are:
  + Keep `getCardToBeShown` as is.
  + Replace it with a boolean function, named for an example `hasCourse`, `isCourseProject`, or whatever.

Choose whichever you like best.

More important than this issue, in my opinion, is whether we want to further break encapsulation by adding main view model to modules view, so that is something I would like to discuss before it is done

superseacat:
47 comments -> this thread starts to finally look like a "regular" PR discussion we used to have 🤣

OlliKiljunen (approved):
Excellent! 🤩 Thank you for implementing requested changed. I like very much the way you solved those.

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -1,26 +1,49 @@
...
+  private CardLayout cl;
+  private MainViewModel mainViewModel;
 
-  public ExercisesView() {
-    // See ModulesView.java
+  /**
+   * Creates an ExerciseView that uses mainViewModel to dynamically adjust its UI components.
+   */
+  public ExercisesView(MainViewModel mainViewModel) {
+    this.mainViewModel = mainViewModel;
```
nikke234, line 38: `MainViewModel` isn't used here anymore, so let's remove the reference and the ctor param completely

nikke234 (approved):
Great, well worth the effort!

Please remove `MainViewModel` from `ExercisesView`, but other than that, feel free to merge :+1:

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
stellatsv: Oh, missed that

PR #494, opened 2021-01-26
Open all assignment tree items in browser

### LINKED ISSUE #277: open "assignment" in browser, the same way as "submission" (for now)?
and also weeks (with a "open in browser" button)

as long as we don't have a headless browser in the IDE shall we implement the same way of showing the assignments as we have for submissions?

https://github.com/Aalto-LeTech/intellij-plugin/issues/245

jaakkonarhi:
# Description of the PR

Closes #277 

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant


src/main/java/fi/aalto/cs/apluscourses/intellij/actions/OpenItemAction.java
```diff
@@ -58,14 +61,24 @@ public void actionPerformed(@NotNull AnActionEvent e) {
...
-        (SubmissionResultViewModel) exercisesTree.findSelected().getLevel(3);
-    if (submission == null) {
+    SelectableNodeViewModel<?> nodeViewModel = exercisesTree.getSelectedItem();
+    if (nodeViewModel == null) {
+      return;
+    }
+
+    String url;
+    if (nodeViewModel instanceof SubmissionResultViewModel) {
+      url = ((SubmissionResultViewModel) nodeViewModel).getModel().getUrl();
```
nikke234, line 71: I think we should change the name of `getUrl` in `SubmissionResult` to `getHtmlUrl`. Not only is it consistent with the other two, but it's also more correct, since it is in fact a URL that is intended to be opened in a browser and not an API end point.

After that we have three different classes with the method `getHtmlUrl`, which raises the question if they should all implement an interface with a method `getHtmlUrl`.

nikke234:
It's not directly related to this PR, but since the changes are closely related, I suggest that we also change `UrlRenderer` to use `BrowserUtil.browse` instead of the current `Desktop.browse` (due to #403)

src/test/java/fi/aalto/cs/apluscourses/intellij/actions/OpenItemActionTest.java
```diff
@@ -15,80 +15,127 @@
...
+  private UrlRenderer urlRenderer;
 
   /**
    * Called before each test.
+   * @param viewModelType The type of SelectableNodeViewModel to be tested, as a string
    */
-  @Before
-  public void setUp() {
-    Exercise exercise = new Exercise(223, "TestEx", "http://example.com", 0, 1, 10, true);
+  public void setUp(String viewModelType) {
```
nikke234, line 49: I recommend passing an instance of `SelectableNodeViewModel` directly instead of the current approach which, as you noted, isn't that nice. So one test would create an exercise group view model and pass that, another would create a submission result view model, and so on

jaakkonarhi: oh yeah why didn't I do that

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/OpenItemAction.java
```diff
@@ -58,14 +61,24 @@ public void actionPerformed(@NotNull AnActionEvent e) {
...
+    }
+
+    String url;
+    if (nodeViewModel instanceof SubmissionResultViewModel) {
+      url = ((SubmissionResultViewModel) nodeViewModel).getModel().getHtmlUrl();
+    } else if (nodeViewModel instanceof ExerciseViewModel) {
+      url = ((ExerciseViewModel) nodeViewModel).getModel().getHtmlUrl();
+    } else if (nodeViewModel instanceof ExerciseGroupViewModel) {
+      url = ((ExerciseGroupViewModel) nodeViewModel).getModel().getHtmlUrl();
+    } else {
```
nikke234, line 76: As much as I'm allergic to inheritance, a chain like this basically screams for 
```
interface Browsable {
  @NotNull
  String getHtmlUrl();
}
```
which would then be implemented by `ExerciseGroup`, `Exercise`, and `SubmissionResult` (and add `@Override` to the `getHtmlUrl` methods)

nikke234:
I'll remove the draft status since reviews have been requested. `SubmissionRenderingErrorNotification` should also be renamed to match the changes here

nikke234 (approved):
Great! :+1: Please do rename `SubmissionRenderingErrorNotification` though :wink:

jaakkonarhi:
Now it matches `UrlRenderer`

Taikelenn (approved):
Really nice! I actually see a slight performance improvement (when the browser opens) on Windows with this change 😁

Issue #267, opened 2020-07-30
searchability (filtering) the Modules/Exercises/etc. inside the A+ Courses subtabs


superseacat:
research how complex a feature alike would be to implement for Modulus' and Exercises' contents:
![search_example](https://user-images.githubusercontent.com/9991098/88890479-5a24de80-d24a-11ea-92bc-bdfe0c491a1e.png)

superseacat:
`SearchPopUp`, `PopUpFactory`, etc.

PR #502, opened 2021-01-29
Search improvements

Taikelenn:
# Description of the PR

Related to issue #267.

This pull request:
+ rewrites `ModuleListView` so that it uses a custom renderer instead of Swing forms for each entry
+ improves on the existing `ExercisesTreeRenderer` by explicitly specifying the main label text (the text that is to be subject to searching)
+ adds searching capabilities for the modules and exercises view using the IntelliJ's List- and TreeSpeedSearch

To be fixed/discussed:
+ `ModuleListView` depends on `ModuleListElementView`, which is now redundant because the drawing logic has been moved to the custom renderer. However, `BaseListView` which is the base class of `ModuleListView` requires *some* Swing form being present, so we can't just simply delete that `ModuleListElementView`.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [x] Yes
- [ ] Not relevant


nikke234:
> However, BaseListView which is the base class of ModuleListView requires some Swing form being present, so we can't just simply delete that ModuleListElementView.

Unless you also delete it from `BaseListView` :wink: But I'll have to take a closer look at this

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModuleListElementView.java
```diff
@@ -1,28 +1,49 @@
...
+      SpeedSearchUtil.applySpeedSearchHighlighting(list, this, true, selected);
+    }
   }
 
+  public static final ColoredModuleListRenderer CellRenderer = new ColoredModuleListRenderer();
+
   @Override
   public JComponent getRenderer() {
-    return basePanel;
+    return CellRenderer;
```
nikke234, line 47: Should be either `CELL_RENDERER` or `cellRenderer` for consistency with the existing capitalization style

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModuleListElementView.java
```diff
@@ -1,28 +1,49 @@
...
-
-  public ModuleListElementView() {
-    nameLabel.setIcon(PluginIcons.A_PLUS_MODULE);
+  private static class ColoredModuleListRenderer
+          extends ColoredListCellRenderer<ModuleListElementViewModel> {
+
+    private static final SimpleTextAttributes STATUS_TEXT_STYLE = new SimpleTextAttributes(
+            SimpleTextAttributes.STYLE_ITALIC | SimpleTextAttributes.STYLE_SMALLER, null);
+    private static final SimpleTextAttributes UPDATE_TEXT_STYLE = new SimpleTextAttributes(
+            SimpleTextAttributes.STYLE_BOLD, new Color(254, 127, 156));
```
nikke234, line 22: Does the color look good in both light and dark themes? [`JBColor`](https://upsource.jetbrains.com/idea-ce/file/idea-ce-4b94ba01122752d7576eb9d69638b6e89d1671b7/platform/util/ui/src/com/intellij/ui/JBColor.java) provides an alternative that contains two colors

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -44,6 +48,14 @@ private void createUIComponents() {
     exerciseGroupsTree.setCellRenderer(new ExercisesTreeRenderer());
     exerciseGroupsTree.addNodeAppliedListener(
         ActionUtil.createOnEventLauncher(OpenItemAction.ACTION_ID, exerciseGroupsTree));
+
+    new TreeSpeedSearch(exerciseGroupsTree, x -> {
+      Object treeObject = ((DefaultMutableTreeNode) x.getLastPathComponent()).getUserObject();
+      if (treeObject instanceof ExerciseViewModel) {
+        return ((ExerciseViewModel) treeObject).getPresentableName();
+      }
+      return ""; // disable searching on entries other than assignments
```
nikke234, line 57: I guess searching for submissions is pretty pointless, since they are all named "Submission x", but could searching for rounds/weeks be useful?

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModuleListElementView.java
Taikelenn: Will fix

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
Taikelenn: Aren't all weeks called "week/round X" as well? There's no problem in adding that of course, but exercises are special in the way that they have custom names rather than just numbers

Taikelenn: And all weeks are instantly visible & sorted when launching IntelliJ, so it's arguably faster to just locate the desired week with your eyes

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -44,6 +48,14 @@ private void createUIComponents() {
     exerciseGroupsTree.setCellRenderer(new ExercisesTreeRenderer());
     exerciseGroupsTree.addNodeAppliedListener(
         ActionUtil.createOnEventLauncher(OpenItemAction.ACTION_ID, exerciseGroupsTree));
+
+    new TreeSpeedSearch(exerciseGroupsTree, x -> {
+      Object treeObject = ((DefaultMutableTreeNode) x.getLastPathComponent()).getUserObject();
```
nikke234, line 53: It's a bit unfortunate that the code here has to deal with `DefaultMutableTreeNode`, since `TreeView` doesn't really "officially promise" to use `DefaultMutableTreeNode`.

One option would be that `TreeSpeedSearch` is instead added in `TreeView`, and `SelectableNodeViewModel` would then have some sort of `getSearchableString` method, which gets overridden appropriately by `ExerciseGroupViewModel`, `ExerciseViewModel`, and `SubmissionResultViewModel`.

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModuleListElementView.java
Taikelenn: Will check, if it doesn't look good I'll replace it with the standard text color that's currently used

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
nikke234: Well, to be specific, weeks/rounds are called whatever the course calls them. Submissions however are named by us. The difference with rounds and submissions is that submissions will have duplicate names ("Submission 2") might have 10 matches, but ("Week 2") should usually only have one match.

But I think you're right, it's probably not very useful in practice to search for them.

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
nikke234: Hmm, apparently `TreeView` has a static `getViewModel` method (probably specifically because of this reason), could that be used here?

Taikelenn: It throws IllegalArgumentException

nikke234: `TreeView.getViewModel(x.getLastPathComponent())` shouldn't throw

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModuleListElementView.java
```diff
@@ -1,28 +1,49 @@
...
+                                         ModuleListElementViewModel element,
+                                         int index,
+                                         boolean selected,
+                                         boolean hasFocus) {
+      append(element.getName(), element.getTextAttribute(), true);
+      append("  [" + element.getStatus() + "]", STATUS_TEXT_STYLE);
+      setToolTipText(element.getTooltip());
+      setIcon(PluginIcons.A_PLUS_MODULE);
+      setIconTextGap(4);
+      if (!element.isUpdateAvailable()) {
```
nikke234, line 35: This should surely not be negated :smile:

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModuleListElementView.java
```diff
@@ -1,28 +1,49 @@
...
+        append("  UPDATE AVAILABLE!", UPDATE_TEXT_STYLE);
+      }
```
nikke234, line 37: The condition should surely not be negated :smile:

nikke234:
I wonder if it would be possible to match exercises even if the parent node (week/round) is closed? The current search is already an improvement, so it can be kept as is if IntelliJ doesn't provide a simple way to achieve that

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModuleListElementView.java
nikke234: The color certainly makes the update stand out, which is good, but it also almost gives the impression that the module has some kind of error
image

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
Taikelenn: Indeed you're right. I will switch to this because it makes the code look better, however it does not really increase the type safety anyway because the implementation of this function is essentially this:

Taikelenn: Screenshot_127

nikke234: You are correct that it doesn't improve type safety, but that's also not the concern here.

The concern is that with the current implementation, `ExercisesView` assumes that `TreeView` uses `DefaultMutableTreeNode`, but that is just an implementation detail in `TreeView` and not part of its public interface. In other words, it should be possible to change `DefaultMutableTreeNode` to something else without breaking any code using `TreeView`.

That is where `getViewModel` comes in. It encapsulates the implementation detail and now if the implementation of `TreeView` is modified, it won't break `ExercisesView`.

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModuleListElementView.java
Taikelenn: To be fair I copied the color code from the current ModuleListElementView Swing form, but for some reason the color code is not shown in the actual UI

nikke234: Ah true, weird that it was never visible then :smile:

src/main/java/fi/aalto/cs/apluscourses/presentation/base/SearchableNode.java
```diff
@@ -0,0 +1,10 @@
+package fi.aalto.cs.apluscourses.presentation.base;
+
+import org.jetbrains.annotations.NotNull;
+
+public interface SearchableNode {
+  @NotNull
+  default String getSearchableString() {
+    return "";
+  }
+}
```
nikke234, line 10: Sorry I was a bit unclear, but I mainly suggested this as a workaround before I discovered the `Treeview.getViewModel` method, so I would say this isn't really needed anymore

nikke234:
Regarding this:
> I wonder if it would be possible to match exercises even if the parent node (week/round) is closed? The current search is already an improvement, so it can be kept as is if IntelliJ doesn't provide a simple way to achieve that

Some of the [`TreeSpeedSearch` constructors](https://upsource.jetbrains.com/idea-ce/file/idea-ce-4b94ba01122752d7576eb9d69638b6e89d1671b7/platform/platform-impl/src/com/intellij/ui/TreeSpeedSearch.java) take a parameter named `canExpand`, which sounds like it should be possible

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -66,6 +68,11 @@ private void createUIComponents() {
     exerciseGroupsTree.setCellRenderer(new ExercisesTreeRenderer());
     exerciseGroupsTree.addNodeAppliedListener(
         ActionUtil.createOnEventLauncher(OpenItemAction.ACTION_ID, exerciseGroupsTree));
+
+    new TreeSpeedSearch(exerciseGroupsTree, x -> {
```
nikke234, line 72: Nitpick: can we rename `x` to `path` (or `treePath`)

Taikelenn: Yes

src/main/java/fi/aalto/cs/apluscourses/presentation/base/SearchableNode.java
Taikelenn: I like this in the way that it keeps the lambda in TreeSpeedSearch's constructor short, and moves the "get searchable string" logic from View to the ViewModel

Taikelenn:
> 
> 
> Regarding this:
> 
> > I wonder if it would be possible to match exercises even if the parent node (week/round) is closed? The current search is already an improvement, so it can be kept as is if IntelliJ doesn't provide a simple way to achieve that
> 
> Some of the [`TreeSpeedSearch` constructors](https://upsource.jetbrains.com/idea-ce/file/idea-ce-4b94ba01122752d7576eb9d69638b6e89d1671b7/platform/platform-impl/src/com/intellij/ui/TreeSpeedSearch.java) take a parameter named `canExpand`, which sounds like it should be possible

Yes, setting this parameter to true automatically expands the week in which the found entry is located

src/main/java/fi/aalto/cs/apluscourses/presentation/base/SearchableNode.java
nikke234: Alright :+1: Just one nit, can we then remove `Node` from the name, since this isn't limited to tree nodes

OlliKiljunen (changes requested):
Very nice code!

I left a few comments. In particular, I hope that you would address the second of them, one related to `ModuleListElementView` class.

Futhermore, I would like you to consider moving this search functionality into the base classes, namely `BaseListView` and `TreeView`. `ListElementViewModel` and `SelectableNodeViewModel` (that could be renamed to something more generic!) would then implement (abstractly) `Searchable`.

Why I'd prefer this? Aren't those base classes already "too big"? Well, generic UI widgets are (in my opinion) exceptions of single-responsibility principle of classes: it's okay to equip them with all kind of "bells and whistles" for the convenience of the subclasses (even if not all subclasses necessarily need them). And they usually are big.

The main advantage this would give us, is that the search functionality would be then separated from our domain-specific classes, making them easier to understand and maintain.

All in all, really good solution! :+1: I only request changes related to  the above-mentioned need for some refactoring.

And yeah, `getSearchableString` could be unit-tested for the implementing classes... :P

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModuleListView.java
```diff
@@ -1,29 +1,26 @@
...
-    extends BaseListView<ModuleListElementViewModel, ModuleListElementView> {
-
-
+        extends BaseListView<ModuleListElementViewModel, ModuleListElementView> {
+  /**
+   * Constructs a view for the module list.
+   */
   public ModuleListView() {
     super(new ModuleListElementView());
+    setFixedCellHeight(26);
```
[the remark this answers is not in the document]
Taikelenn: Screenshot_128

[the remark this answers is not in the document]
Taikelenn: This is how it looks without that `setFixedCellHeight`, and it seems like changing the font size does not influence the spacing (i.e. the spacing is not preserved, which is not good)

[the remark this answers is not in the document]
OlliKiljunen: It doesn't have to look "good" with a non-default font size but ideally (for better accessibility), the text should remain visible even if font size were increased.

I wonder if it's possible to use fontsize- bound length units like in HTML/CSS. If it was, that could perhaps be a better  choice there.

It doesn't have to be fixed here, though. This code is ok, as-is.

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -66,6 +68,11 @@ private void createUIComponents() {
     exerciseGroupsTree.setCellRenderer(new ExercisesTreeRenderer());
     exerciseGroupsTree.addNodeAppliedListener(
         ActionUtil.createOnEventLauncher(OpenItemAction.ACTION_ID, exerciseGroupsTree));
+
+    new TreeSpeedSearch(exerciseGroupsTree, treePath -> {
```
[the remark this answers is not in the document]
Taikelenn: Yes, the constructor registers the search functionality. It's how JetBrains documentation recommends this: https://plugins.jetbrains.com/docs/intellij/lists-and-trees.html#listspeedsearch-and-treespeedsearch

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModuleListView.java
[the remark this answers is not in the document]
Taikelenn: This might be possible by calling `setFixedCellHeight` with a parameter dependant on `list.getFont().getSize()` - however, you'd have to somehow catch the event of the font being changed (unlikely that something like this exists, however I haven't checked), set the cell height AND refresh the list

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModuleListElementView.java
```diff
@@ -1,28 +1,49 @@
...
 import icons.PluginIcons;
+import java.awt.Color;
 import javax.swing.JComponent;
-import javax.swing.JLabel;
-import javax.swing.JPanel;
+import javax.swing.JList;
+
+import org.jetbrains.annotations.NotNull;
 
 public class ModuleListElementView implements ListElementView {
```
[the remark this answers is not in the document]
Taikelenn: Removed

superseacat (changes requested):
how about an entry into TESTING.md? 😉

src/main/java/fi/aalto/cs/apluscourses/ui/module/ModuleListView.java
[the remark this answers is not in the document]
OlliKiljunen: There's no reason to go that far for an issue this small, so let's then just keep it as it is in this PR. :+1:

OlliKiljunen (approved):
Excellent, approved. :+1: Thanks for taking my review comments into account.

I just would like to know if you considered my previous suggestion of moving search functionality into the base classes and what's your reasoning on that.

Anyway, great code!

Taikelenn:
> 
> 
> how about an entry into TESTING.md? 😉

Added! :]

Taikelenn:
> 
> 
> Excellent, approved. 👍 Thanks for taking my review comments into account.
> 
> I just would like to know if you considered my previous suggestion of moving search functionality into the base classes and what's your reasoning on that.
> 
> Anyway, great code!

Moving the search functionality to the base class is possible, but I can think of some potential drawbacks:
1. If we ever add a list/tree which we do not want to be searchable, it won't be possible because the base class already added the search module
2. The mere adding of List- or TreeSpeedSearch only makes the search partially working - you have to use a custom renderer in the derived class by calling `setCellRenderer` and potentially other UI customization methods (such as `setFixedCellHeight`). So even if we move the `new ListSpeedSearch` to the base class, we still have work to do in the derived class. IMO it looks nicer when all the searching-related code is in the same place.

nikke234 (approved):
Great stuff! & thanks for fixing the review comments :+1: 

You could add a test for `ExerciseViewModel` and `ModuleListElementViewModel` which checks that `getSearchableString` returns their names instead of the default empty string. But this is also good to merge as is

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExerciseViewModel.java
```diff
@@ -20,6 +21,11 @@ public String getPresentableName() {
     return APlusLocalizationUtil.getEnglishName(getModel().getName());
   }
 
+  @Override
+  public @NotNull String getSearchableString() {
```
superseacat, line 25: do you think this could have a javadoc?

src/main/java/fi/aalto/cs/apluscourses/presentation/module/ModuleListElementViewModel.java
```diff
@@ -87,13 +95,11 @@ public String getStatus() {
   }
 
   /**
-   * Returns a font weight in which the module is shown on a list.
-   * @return A {@link Float} that can be set to font weight.
+   * Indicates whether the module shown on a list should be displayed bold-faced.
    */
-  public float getFontWeight() {
+  @NotNull
+  public Boolean isBoldface() {
```
superseacat, line 101: a `Boolean` that is `@NotNull` makes me think there's no need for a 3-valued **boolean**, but why not the primitive then?

src/main/java/fi/aalto/cs/apluscourses/presentation/base/Searchable.java
```diff
@@ -0,0 +1,10 @@
+package fi.aalto.cs.apluscourses.presentation.base;
+
+import org.jetbrains.annotations.NotNull;
+
+public interface Searchable {
+  @NotNull
+  default String getSearchableString() {
+    return "";
```
superseacat, line 8: Nit: `StringUtils.EMPTY` or externalizing the `""` into props, what do you think :question:

PR #503, opened 2021-02-01
Fix assignment tree double click

jaakkonarhi:
# Description of the PR

Because of #494 items that should get expanded when double clicking in the assignment tree also get opened in the browser. Now only items with no children get opened

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes
- [x] Not relevant

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
@@ -12,6 +12,8 @@ import java.awt.event.ActionListener;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


OlliKiljunen (approved):
This is good. I approve.

In Swing/AWT (like in the most of the event-driven GUI frameworks), there's a thing called "event consumption". That is, when event (like "mouse clicked" here) gets "consumed" by some handler, it is no longer passed to the subsequent handlers.

I see this PR's problem setting as a textbook example of such a situation: if double-clicking causes node expansion, the event would be already consumed by the base class' handler and never get to this handler. That way, we wouldn't need such a logic here that "guesses" whether this is a case where the node expansion happens.

We can actually take a look at the base class' handler method (IntelliJ's [Tree](https://github.com/JetBrains/intellij-community/blob/91d851a2f7673d35fd1daf4c3fcf1142f961b356/platform/platform-api/src/com/intellij/ui/treeStructure/Tree.java#L660-L668) class):
```java
    @Override
    public void mouseReleased(MouseEvent event) {
      setPressed(event, false);
      if (event.getButton() == MouseEvent.BUTTON1 &&
          event.getClickCount() == 2 &&
          TreeUtil.isLocationInExpandControl(Tree.this, event.getX(), event.getY())) {
        event.consume();
      }
    }
```

We see that if the expansion happens, the event is consumed, as expected. Why does it still gets to the handler of this PR?

Apparently, Java calls mouse listeners in reverse order in which they were registered (which actually makes sense). Thus, the listener of the base class (that was registered in the base class' constructor) is called only after our listener (that was registered this class' constructor).* I'm afraid there's no straightforward enough way to change that order to make it possible to rely on event consumption instead of "forward-looking guessing". So let's do it this way. 👍 There's nothing wrong in this way, anyway.

I hope this makes sense.

*) The former also consideres `mouseReleased` event and the latter `mouseClicked` but that could be changed if needed.

stellatsv (approved):
I approve of the code as well. Nice work spotting the issue!

Issue #517, opened 2021-02-09
UI tests in CI


superseacat:
- [x] into CI: https://github.com/JetBrains/intellij-ui-test-robot/blob/master/ui-test-example/FAQ.md#can-i-run-the-tests-on-github-actions
- [x] run async: https://github.com/JetBrains/intellij-ui-test-robot#quick-start
- [x] caching in background

PR #518, opened 2021-02-09
Automate UI testing: into the CI world!

superseacat:
UI tests in CI #517

# Description of the PR

-  added caching for the build with "unit" and integration tests
-  adds a new job for running of UI tests in headless environments (linux)
-  will separate windows and mac struggle into a separate branch `automate-ui-testing-win-mac`

I'll squash the commits, apparently 😉


OlliKiljunen:
If adding sleep time between Gradle tasks is problematic, UI tests can be modified to have more time before timeouting finding welcome frame.

superseacat:
> If adding sleep time between Gradle tasks is problematic, UI tests can be modified to have more time before timeouting finding welcome frame.

naaah, it's ugly but works... I'll see why `sleep` does not work properly afterward (I still prefer raw Linux commands to those hipster-re-invented actions)

.github/workflows/build.yml
```diff
@@ -14,8 +14,41 @@ jobs:
...
+          java-version: 11
+      - name: Setup UI testing env
+        run: |
+          export DISPLAY=:99.0
+          Xvfb -ac :99 -screen 0 1920x1080x16 &
+          gradle runIdeForUiTests &
+      - name: Sleep for 180 seconds
+        uses: jakejarvis/wait-action@master
+        with:
+          time: '180s'
```
nikke234, line 52: Just to make sure,
```
- name: Sleep for 180 seconds
  run: |
    sleep 180
```
doesn't work?

superseacat: actually, surprisingly, it does now! thanks 🤩

nikke234: Great! :+1: Then the commented previous implementation could be removed as well

.github/workflows/build.yml
```diff
@@ -14,8 +14,41 @@ jobs:
...
+        with:
+          path: ~/.sonar/cache
+          key: ${{ runner.os }}-sonar
+          restore-keys: ${{ runner.os }}-sonar
+      - name: Cache Gradle packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
```
nikke234, line 28: Just curious, are these needed or just an improvement in how long the action takes to run?

nikke234 (approved):
Great! :+1: I can see from the commit history that getting this to work wasn't easy :joy: So thanks for the perseverance :+1: I suggest squashing the commits when merging :wink:

.github/workflows/build.yml
superseacat: with cache: https://github.com/Aalto-LeTech/intellij-plugin/actions/runs/568557365
without cache: https://github.com/Aalto-LeTech/intellij-plugin/actions/runs/568555658

so caching is build is shorter (see the `build` jobs)

OlliKiljunen (approved):
Wonderful work! 👍 

I think these configs are even understandable enough for us to maintain. :)

PR #521, opened 2021-02-10
Late assignments

### LINKED ISSUE #405: shadow points (late assignments)
One small detail: if exercises are submitted late, they get zero points and a red X is shown. The exercise can however be correctly solved. I’d suggest that in case of a zero score after the weekly deadline, the symbol could be a clock and its color something more neutral than red. (gray/orange/yellow/blue) or :ghost: 

https://apluslms.slack.com/archives/CQ92SAHNY/p1601459651020900

jaakkonarhi:
# Description of the PR

With this PR each assignment gets checked for their best submission, and if that submission has a late penalty, then the assignment is marked late and gets a new icon and status text

image

The submission results should be cached and `MainViewModel::UpdateExercises` should be run in a new thread

Closes #405 

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.


src/main/java/fi/aalto/cs/apluscourses/model/Exercise.java
```diff
@@ -28,30 +28,34 @@
...
    * Construct an exercise instance with the given parameters.
-   *
-   * @param id                The ID of the exercise.
+   *  @param id                The ID of the exercise.
    * @param name              The name of the exercise.
    * @param htmlUrl           A URL to the HTML page of the exercise.
    * @param userPoints        The best points that the user has gotten from this exercise.
    * @param maxPoints         The maximum points available from this exercise.
    * @param maxSubmissions    The maximum number of submissions allowed for this exercise.
+   * @param latePenalty       The penalty applied if the best submission is late.
```
superseacat, line 41: %, right?

jaakkonarhi: it's from 0.0 to 1.0, so is it called a multiplier?

src/main/java/fi/aalto/cs/apluscourses/model/Exercise.java
```diff
@@ -132,6 +137,14 @@ public int getMaxSubmissions() {
     return maxSubmissions;
   }
 
+  public double getLatePenalty() {
+    return latePenalty;
+  }
+
+  public boolean isLate() {
+    return this.latePenalty != 0.0;
+  }
```
nikke234, line 146: It's a bit weird that an exercise would have a late penalty, since it's not exercises that are late or penalized but submissions. I then saw that submission results also have a late penalty, I wonder why that information is duplicated?

jaakkonarhi: Would a better way be to store the best submission and check the penalty of that

nikke234:
I see now that you removed the late penalty from exercises, which is great :+1: 

IMO this is what a very nice solution to this issue would roughly look like: each submission result knows its own late penalty. Each exercise knows which of its submissions was the best one. The current implementation is already very close to this, but there is one major issue left.

The `SubmissionResult` which corresponds to the best submission gets special treatment compared to other submission results. In practice this means that the best `SubmissionResult` has correct data, but the other `SubmissionResult` instances have incorrect data. For an example, if the best submission of an exercise has a late penalty, then the following bizarre scenario happens:

- The list of submission results for that exercise includes a submission result with the same ID as `getBestSubmission`, but with a late penalty of 0!
- `getBestSubmission` has the same ID as one of the submissions in the list, but this time with a non-zero late penalty!

The quick-fix would be that `Exercise.fromJsonObject` also adds the correct late penalty in the for-loop, if `submissionId == exercise.bestSubmission.getId()`. But even after that fix, all other submission results would still have a potentially incorrect late penalty.

@stellatsv has faced this same issue with the way the statuses are handled: `Exercise.fromJsonObject` just assumes that the status is always either UNOFFICIAL or GRADED, which is sometimes incorrect. 

At the moment `SubmissionResult` works in the following horrifying way:

- The status of a `SubmissionResult` is potentially incorrect, unless you are looking at a `SubmissionResult` created by `SubmissionStatusUpdater` which actually looks at the individual submission API.
- The late penalty of a `SubmissionResult` is potentially incorrect, unless you are looking at a `SubmissionResult` which is specifically the best submission of an exercise (and even weirder - it has to be not from the list but from the `getBestSubmission` method).

I think every `SubmissionResult` should be created by looking at the individual submission API similarly to how you do it with the best submission at the moment. That is the only way to have the correct status and late penalty for all submissions. It would make the rest of this very easy, and also solve the issue that @stellatsv faced with incorrect statuses.

It will need some design and caching but I think doing it right would be worth it in the long run. Let's discuss the details very soon, so that the PR doesn't get stalled indefinitely. Caching will be needed either way, so I see no reason not to fetch every `SubmissionResult` properly instead of adding more potentially incorrect data to `SubmissionResult`.

OlliKiljunen (commented):
I think the main problem with this PR stems from the non-object oriented design of `Points` class. We should made some refactoring so that Exercises and SubmissionResults (with User?) would form a coherent object model (and get rid of ID-to-value maps etc.).

I agree with Nikolas's idea. What I would add to that is that objects such as Exercises and SubmissionResults should be created as "stub" objects as soon as their ID is known (that is, when their "parent" object is loaded) and the objects themselves would then populate their data when needed (by making an API request).

The code footprint of this PR is reasonably small that it doesn't hurt to merge this. Even though it would cause those anomalies to object model that Nikolas described, that shouldn't block the merge, as long as everything works for the user. Those anomalies seems to me more like symptoms of existing design problems.

PR #528, opened 2021-02-11
Fixes IllegalStateException related to exercises tree

OlliKiljunen:
# Description of the PR

+ A quick fix to eliminate `java.lang.IllegalStateException: @NotNull method fi/aalto/cs/apluscourses/ui/utils/TreeModelBuilder.getUserObject must not return null`

+ Now, before `ExercisesTreeRenderer::customizeCellRenderer` handles a node, it checks if it happens to be an "irrelevant" placeholder node which should just be ignored.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesTreeRenderer.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesTreeRenderer.java
@@ -12,6 +12,7 @@ import fi.aalto.cs.apluscourses.ui.base.TreeView;
 import icons.PluginIcons;
 import javax.swing.Icon;
 import javax.swing.JTree;
+import javax.swing.tree.DefaultMutableTreeNode;
 import org.jetbrains.annotations.NotNull;
 
 public class ExercisesTreeRenderer extends ColoredTreeCellRenderer {
@@ -45,6 +46,9 @@ public class ExercisesTreeRenderer extends ColoredTreeCellRenderer {
                                     boolean isLeaf,
                                     int row,
                                     boolean hasFocus) {
+    if (isIrrelevantNode(value)) {
+      return;
+    }
     SelectableNodeViewModel<?> viewModel = TreeView.getViewModel(value);
     if (viewModel instanceof ExerciseViewModel) {
       ExerciseViewModel exerciseViewModel = (ExerciseViewModel) viewModel;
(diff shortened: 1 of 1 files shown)


src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesTreeRenderer.java
```diff
@@ -77,4 +81,20 @@ public void customizeCellRenderer(@NotNull JTree tree,
...
+   * @param value A parameter passed to {@code customizeCellRenderer}
+   * @return True, if node is irrelevant and should be ignored, otherwise false.
+   */
+  private boolean isIrrelevantNode(Object value) {
+    // That irrelevant root has no user object
+    // We assume that all relevant nodes that implement DefaultMutableTreeNode
+    // do have user objects.
+    return value instanceof DefaultMutableTreeNode
+        && ((DefaultMutableTreeNode) value).getUserObject() == null;
+  }
```
nikke234, line 99: I have two questions related to this:

1. Is it always an instance of `DefaultMutableTreeNode`, i.e. can we remove the `instanceof`?
2. Our `TreeView` hides the root node, so I guess ideally it shouldn't be passed to the custom renderer at all, but I guess IntelliJ's `Tree` doesn't do that. Maybe a less weird way would be `isRoot`, which checks for `((DefaultMutableTreeNode) value).getParent() == null`

OlliKiljunen: 1. Actual, relevant nodes could, in theory, be something else so it is good that the check is there
2. Hmm... so is it so, that for any relevant node it holds that their parent is never null? In that case, we should follow your suggestion. 👍

nikke234: 1. Oops I missed that this is in `ExercisesTreeRenderer`, not `TreeView`. But in that case then this doesn't actually work if the tree model used is no longer `DefaultMutableTreeNode`, since the type of the root node would change as well, and it would no longer be classified as irrelevant. I guess we wouldn't rely on `DefaultMutableTreeNode` anywhere except in `TreeView` 
2. I'm pretty sure they should all have a parent, but I'll double-check to make sure.

OlliKiljunen: 1. Yeah, the check is a bit fuzzy. It wasn't easy to find rules that could be assumed to always match to those "irrelevant" nodes, so I came up with that. If it's enough to check it ain't a root node then that's much better rule.

nikke234: (But I realize that this is just a fix for how `Tree` uses the custom renderer, so it doesn't have to nice or perfect)

Taikelenn (approved):
Good change. I was also wondering if the same issue can occur in the searching functionality, since it also (indirectly) operates on DefaultMutableTreeNodes. However it seems like that won't happen because TreePaths are used there

jaakkonarhi (approved):
I think I understand what's going on, so I approve

PR #539, opened 2021-02-15
Session-based auth [DO-NOT-MERGE]

OlliKiljunen:
**WARNING:** THIS CODE HANDLES USER CREDENTIALS IN AN INSECURE WAY. THIS CODE IS INTENDED FOR TEST/PROTOTYPING USE ONLY. **NEVER USE THIS CODE IN PRODUCTION!**

# Description of the PR

+ Session based authentication, proof of concept 
+ HTTP POST does not work (would require the correct CSRF token)


OlliKiljunen:
Still doesn't work :(

OlliKiljunen:
Apparently, the API doesn't accept session based authentication in POST requests.

OlliKiljunen:
I was able to construct a post request to HTML URL and get it authenticated but apart from being ugly that would cause separate handling of the response. Tell me, if you want to see my latest code on this.

src/main/java/fi/aalto/cs/apluscourses/dal/SessionAuthentication.java
```diff
@@ -0,0 +1,86 @@
...
+
+  }
+
+  public void setCookie(Cookie cookie) {
+    LOGGER.debug("setCookie: {}", cookie);
+    synchronized (cookies) {
+      cookies.remove(cookie);
+      cookies.add(cookie);
+    }
+  }
```
OlliKiljunen, line 50: I re-implemented cookie storage only to realise that what I should have done is to re-use HttpClient object from request to request. :facepalm: :(

OlliKiljunen:
A possibility (= another bad idea): use session-based auth to read the token and then use that.

nikke234:
Very nice findings! :star_struck:  Is the following a correct summary of the available options?

- Ask A+ to add support for session based authentication in POST requests to the API
- Make POST requests to the HTML URLs used by the web interface (hacky)
- Use session based authentication to read the user token from the HTML (hacky)

If that's correct, I suggest we could then close this PR and decide on one of those approaches.

Issue #531, opened 2021-02-12
Refactoring wish list


OlliKiljunen:
How do you wanna see the code base refactored. Add your ideas here.

superseacat:
- [ ] replace concurrent installation of components (like learning modules) with the sequential approach, simplify
- [ ] replace raw thread implementation with some `Executors` approach, in other words, delegate Thread handling to some ready-made JVM mechanism
- [ ] choose-and-use some alternative existing ForkJoin implementation

OlliKiljunen:
Rename the project (intellij-plugin) to something more relevant.

stellatsv:
No longer relevant.

PR #544, opened 2021-02-16
Migrate to Java 11

nikke234:
Yes, it is this easy

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ plugins {
 group 'fi.aalto.cs'
 version '2.0'
 
-sourceCompatibility = 1.8
+sourceCompatibility = 11
 
 def remoteRobotVersion = '0.10.3'
 def fixturesVersion = "1.1.18"
@@ -55,7 +55,7 @@ sourceSets {
 
 compileE2eKotlin {
     kotlinOptions {
-        jvmTarget = "1.8"
+        jvmTarget = "11"
     }
 }
 


OlliKiljunen:
Is this in line with our `since-build` compatibility clause in `plugin.xml`?

nikke234:
Our `since-build` is set to 203. Quoting from [here](https://blog.jetbrains.com/platform/2020/09/intellij-project-migrates-to-java-11/):
> What does it mean for you? If you develop a plugin, you can use new language features and APIs from Java 11, but only if the since-build attribute in plugin.xml is set to 203. If your plugin distribution needs to run on multiple versions of the IDE, for example 202 and 203, then you should remain on Java 8.

TL;DR: yes

OlliKiljunen (approved):
👍

jaakkonarhi (approved):
great

superseacat:
and what about the modularization from Java 9 @nikke234 https://apluslms.slack.com/archives/CSDV9CA31/p1613045743037700 ?

superseacat:
```kotlin
compileE2eKotlin {
    kotlinOptions {
        jvmTarget = "1.8"
    }
}
```
should we change this in any way? does it compile still with the existing option?

OlliKiljunen:
@superseacat said:
> 
> and what about the modularization from Java 9 @nikke234 https://apluslms.slack.com/archives/CSDV9CA31/p1613045743037700 ?

AFAIK, defining modules is optional; you don't have to modularize your code only because your using Java 9+. But we should do that in the future as a part of #531, I'd say.

superseacat (approved):
let's try, I'm leaving soon anyway 😀

nikke234:
> should we change this in any way? does it compile still with the existing option?

Nice catch! They still compile and work as usual, but I guess there is no harm in using 11 there as well (which also works), so I'll change it

Issue #527, opened 2021-02-11
Plugin doesn't handle submission status "waiting" in the A+ API


nikke234:
The plugin only considers the following statuses: "unknown", "unofficial", and "graded". The plugin should also account for "waiting", which is the status of a submission that is currently in grading, as @jaakkonarhi discovered. `getStatusText` in `SubmissionResultViewModel` should return "in grading" instead of "0/n" when the status is "waiting".

stellatsv:
Some remarks about the issue (from working on #364) that may help:
- In the `Exercise.java` method `fromJsonObject`(called when updating `mainViewModel`) the only available information is the number of submissions. The extra submissions are marked as Unofficial and the rest as Graded. This is not accurate because the last submission may be still In Grading. 

- Additionally, it is because the status Unknown is not set to any of the loaded SubmissionResults that the text 'In Grading' is not displayed.

PR #545, opened 2021-02-16
Pending status

### LINKED ISSUE #364: Make exercise status pending when a submission is made
Currently the exercise status remains as is after a submission. This can lead to confusion, since the exercise may show the no-points icon even when the student has full points until the exercise list is refreshed. Adding a pending icon after a submission is made makes it clear that a refresh is needed to get the most up-to-date status.

stellatsv:
# Description of the PR
Closes #364
I approached 'In Grading' as an attribute of the Exercises, given that the change in the UI is happening on the Exercise level. Also, there were some issues related to the `SubmissionResult` status (see #527), but even if these were fixed, checking on the level of the `Exercise` and not iterating through the `SubmissionResults` seems more efficient.         
  
# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.


jaakkonarhi (approved):
It works great, good job

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -125,4 +134,15 @@ public Options getExerciseFilterOptions() {
     return exerciseFilterOptions;
   }
 
+  private void setInGrading(List<ExerciseGroup> exerciseGroups) {
+    Map<Long, Exercise> exercises = new LinkedHashMap<>();
+    exerciseGroups.stream().forEach(group -> exercises.putAll(group.getExercises()));
+    if (exercises.containsKey(submittedForGrading.getId())) {
+      exercises.get(submittedForGrading.getId()).setInGrading(true);
+    }
```
superseacat, line 142: perhaps an improvement to the functional approach:
```java
final long id = submittedForGrading.getId();

    exerciseGroups
            .stream()
            .map(ExerciseGroup::getExercises)
            .filter(exercise -> exercise.containsKey(id))
            .forEach(exercise -> exercise.get(id).setInGrading(true));
```
which is equivalent to
```java
final long id = submittedForGrading.getId();

    for (ExerciseGroup exerciseGroup : exerciseGroups) {
      Map<Long, Exercise> exercise = exerciseGroup.getExercises();
      if (exercise.containsKey(id)) {
        exercise.get(id).setInGrading(true);
      }
    }
```
?

src/main/java/fi/aalto/cs/apluscourses/model/Exercise.java
```diff
@@ -28,6 +28,8 @@
 
   private final boolean submittable;
 
+  private boolean inGrading = false;
```
superseacat, line 31: `false` is the default, so can be omitted: https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -47,6 +51,8 @@
 
   private AtomicBoolean hasTriedToReadAuthenticationFromStorage = new AtomicBoolean(false);
 
+  private Exercise submittedForGrading = null;
```
superseacat, line 54: `null` is the default and can be omitted: https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html

superseacat (approved):
good design and nothing that could prevent the PR "as is" 👍🏻

nikke234:
> I approached 'In Grading' as an attribute of the Exercises, given that the change in the UI is happening on the Exercise level. Also, there were some issues related to the SubmissionResult status (see #527), but even if these were fixed, checking on the level of the Exercise and not iterating through the SubmissionResults seems more efficient.

I like your reasoning and approach here. At the same time though, I wonder if 'in grading' would kinda belong in `SubmissionResult` design-wise, since exercises aren't graded but submissions are. Similarly, the "waiting" status comes from the submission end point instead of the exercise end point. And not to get too philosophical or anything, but if UI classes are decoupled from our model classes, then we shouldn't add something to `Exercise` just because the UI shows it there instead of next to the submission.

After reading your comment on #527 I now realize that it's not possible to do in a nice way at the moment, because `SubmissionResult` statuses are messed up. Your comment also made me realize that Jaakko's draft PR has a similar issue, so I'm starting to think it might be time to bite the bullet and fix `SubmissionResult` properly.

That will definitely take some time though, so I think going with this for now might be a good option - it's definitely an improvement UI-wise :+1: In that case I would still like to review the code itself, if you don't mind waiting despite the two approvals already. :slightly_smiling_face:

stellatsv:
> > I approached 'In Grading' as an attribute of the Exercises, given that the change in the UI is happening on the Exercise level. Also, there were some issues related to the SubmissionResult status (see #527), but even if these were fixed, checking on the level of the Exercise and not iterating through the SubmissionResults seems more efficient.
> 
> I like your reasoning and approach here. At the same time though, I wonder if 'in grading' would kinda belong in `SubmissionResult` design-wise, since exercises aren't graded but submissions are. Similarly, the "waiting" status comes from the submission end point instead of the exercise end point. And not to get too philosophical or anything, but if UI classes are decoupled from our model classes, then we shouldn't add something to `Exercise` just because the UI shows it there instead of next to the submission.
> 
> After reading your comment on #527 I now realize that it's not possible to do in a nice way at the moment, because `SubmissionResult` statuses are messed up. Your comment also made me realize that Jaakko's draft PR has a similar issue, so I'm starting to think it might be time to bite the bullet and fix `SubmissionResult` properly.
> 
> That will definitely take some time though, so I think going with this for now might be a good option - it's definitely an improvement UI-wise +1 In that case I would still like to review the code itself, if you don't mind waiting despite the two approvals already. slightly_smiling_face

Glad the comment helped clear things. Feel free to review the code as well if you want, there's no rush to merge!

src/main/java/fi/aalto/cs/apluscourses/model/SubmissionStatusUpdater.java
```diff
@@ -109,9 +109,13 @@ private void run() {
         try {
           submissionResult =
               dataSource.getSubmissionResult(submissionUrl, exercise, authentication);
-          if (submissionResult.getStatus() != SubmissionResult.Status.UNKNOWN) {
+          if (SubmissionResult.Status.WAITING.equals(submissionResult.getStatus())) {
+            PluginSettings.getInstance().getMainViewModel(project).setSubmittedForGrading(exercise);
+            PluginSettings.getInstance().updateMainViewModel(project);
```
nikke234, line 114: An update includes the following:

- Fetch the course configuration JSON and parse it again.
- Fetch a list of all exercises from the A+ API
- Fetch a list of points for the student.

In this case that's a bit too much, because this gets run repeatedly in increments (0 s, 10 s, 25 s, 45 s, etc). I assume this is here so that the icon gets updated? If that's the case, I have an idea for how to avoid the update, I'll open a PR against this branch

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -47,6 +51,8 @@
 
   private AtomicBoolean hasTriedToReadAuthenticationFromStorage = new AtomicBoolean(false);
 
+  private Exercise submittedForGrading;
```
nikke234, line 54: Since this gets set in one thread (the one in `SubmissionStatusUpdater`) and read in another (when an update occurs and `updateExercises` is called), it should at least be marked as `volatile` to make sure that the modification made by one thread becomes visible in other threads. However in this case, instead of volatile, some synchronization should be used, because the following race condition exists:

- User starts an update using the refresh button, which calls `updateExercises`.
- `SubmissionStatusUpdater` simultaneously calls `setSubmittedFromGrading` from another thread.
- Now there is a race condition in the `submittedForGrading != null` check in `updateExercises`.

I'll add the synchronization in my PR since it's a small change :+1:

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -16,9 +17,12 @@
 import fi.aalto.cs.apluscourses.utils.observable.ObservableReadWriteProperty;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
```
nikke234, line 20: Unused import :smile:

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
nikke234: I now realized that only storing one exercise also has some issues in the very unlikely scenario of submitting two different exercises one after another. So I think there is a way to kill two birds with one stone here: using a `ConcurrentHashMap` allows storing multiple exercises in grading and it also includes synchronization.

nikke234:
Very nice stuff, I especially like the new icon :sunglasses: I opened a PR #557 against this branch that fixes a few of the issues I mentioned above. Please take a look at it and test it. If you think it's fine, then merge it to this branch (no need to review it or wait for two approvals, since the PR isn't going to master). After that I'll go through this again and approve, as it looks solid apart from those minor issues

src/main/java/fi/aalto/cs/apluscourses/model/Exercise.java
```diff
@@ -28,6 +28,8 @@
 
   private final boolean submittable;
 
+  private boolean inGrading;
```
nikke234, line 31: I realized that if you merge my PR, then this should be either `volatile` or `AtomicBoolean`, because in my PR the boolean gets modified in the thread of `SubmissionStatusUpdater` and then read in other threads when the tree is updated. Although atomicity isn't strictly necessary, I suggest going with `AtomicBoolean`, since the `submissionResults` variable in `Exercise` also uses a synchronized list (unnecessarily).

nikke234:
Just make `inGrading` `AtomicBoolean`, remove the unused import, and this is ready to go as far as I'm concerned :rocket:

nikke234 (approved):
:ship: :it:

Issue #366, opened 2020-09-01
code style refactoring


superseacat:
- [x] sonar: reconfigure (gate & rules)
- [x] sonar: fix majority of the code smells
- [x] sonar: agree on the common approach
- [x] IDEA's own **partial** styling (static analysis)

superseacat:
* sonar **quality gate**: these are quilcky checked in PRs (and cause most annoyance) https://sonarcloud.io/organizations/aalto-letech-intellij-plugin/quality_gates/show/41060
* sonar **quality profiles (set of rules)**: language specific, consider changing these if you need to do `//nosonar` in the code
* you can use this plugin in IDEA and get to know about all the coming issues BEFORE the `push`: https://plugins.jetbrains.com/plugin/7973-sonarlint

superseacat:
sonar **quality gate** for test coverage is now set to the current main's coverage **57.4%** @Aalto-LeTech/apluscourses-plugin-team

superseacat:
let approach this bit by bit:

** sonar: fix all the stuff like code smells**, iterating through the "issues" and:

1. silence them down one by one or somehow bulk
2. silence them down **AND** change the corresponding rules, so not happening again
3. fix the issue (rules remain)
4. search for some alternative existing set of rules, reapply

see: https://sonarcloud.io/project/issues?id=Aalto-LeTech_intellij-plugin&resolved=false for example

Issue #522, opened 2021-02-11
plugin configuration file is fetched from remote


superseacat:
- [ ] decide where to host the plugin conf file
- [ ] create a structured file
- [ ] get it from plugin

stellatsv:
The list of courses should be in the configuration file.

Taikelenn:
https://apluscourses-reports.cs.aalto.fi/ we can use our error reporting server

PR #548, opened 2021-02-17
Fixing Sonar (part 1): Fixing the simple issues

superseacat:
code style refactoring #366

doable in an hour they said...

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/dal/TokenAuthentication.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/dal/TokenAuthentication.java
@@ -8,7 +8,6 @@ import org.jetbrains.annotations.Nullable;
 public abstract class TokenAuthentication implements Authentication {
   @NotNull
   protected final String user;
-  @NotNull
   protected final char[] token;
   @Nullable
   protected final PasswordStorage passwordStorage;
@@ -20,8 +19,8 @@ public abstract class TokenAuthentication implements Authentication {
    * @param token           Token.  The array is cloned and the argument is not cleared.
    * @param passwordStorage Password storage.  If null, the authentication won't be persisted.
    */
-  public TokenAuthentication(@NotNull String user,
-                             @NotNull char[] token,
+  protected TokenAuthentication(@NotNull String user,
+                                char[] token,
                              @Nullable PasswordStorage passwordStorage) {
     this.user = user;
     this.token = token.clone();
@@ -50,6 +49,6 @@ public abstract class TokenAuthentication implements Authentication {
 
   public interface Factory {
     @NotNull
-    TokenAuthentication create(@NotNull char[] token);
+    TokenAuthentication create(char[] token);
   }
 }
--- a/src/main/java/fi/aalto/cs/apluscourses/dal/TokenAuthentication.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/dal/TokenAuthentication.java
@@ -8,6 +8,7 @@ import org.jetbrains.annotations.Nullable;
 public abstract class TokenAuthentication implements Authentication {
   @NotNull
   protected final String user;
+  @NotNull
   protected final char[] token;
   @Nullable
   protected final PasswordStorage passwordStorage;
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/APlusAuthenticationAction.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/APlusAuthenticationAction.java
@@ -17,8 +17,6 @@ import fi.aalto.cs.apluscourses.model.Course;
 import fi.aalto.cs.apluscourses.presentation.AuthenticationViewModel;
 import fi.aalto.cs.apluscourses.presentation.CourseViewModel;
 import fi.aalto.cs.apluscourses.presentation.MainViewModel;
-import fi.aalto.cs.apluscourses.presentation.base.BaseViewModel;
-import java.util.Optional;
 import org.jetbrains.annotations.NotNull;
 
 public class APlusAuthenticationAction extends DumbAwareAction {
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/APlusAuthenticationAction.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/APlusAuthenticationAction.java
@@ -69,11 +69,11 @@ public class APlusAuthenticationAction extends DumbAwareAction {
   public void actionPerformed(@NotNull AnActionEvent e) {
     Project project = e.getProject();
     MainViewModel mainViewModel = mainViewModelProvider.getMainViewModel(project);
-    Course course = Optional.ofNullable(mainViewModel.courseViewModel.get())
-        .map(BaseViewModel::getModel)
-        .orElse(null);
-
-    if (course == null) {
+    Course course;
+    var courseViewModel = mainViewModel.courseViewModel.get();
+    if (courseViewModel != null) {
+      course = courseViewModel.getModel();
+    } else {
       return;
     }
 
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
@@ -31,7 +31,6 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.ExecutionException;
@@ -73,7 +72,7 @@ public class CourseProjectAction extends AnAction {
 
   private final ExecutorService executor;
 
-  private static final List<CourseItemViewModel> AVAILABLE_COURSES = Arrays.asList(
+  private static final List<CourseItemViewModel> AVAILABLE_COURSES = List.of(
       new CourseItemViewModel("O1", "Fall 2020",
           "https://grader.cs.aalto.fi/static/O1_2020/projects/o1_course_config.json"),
       new CourseItemViewModel("Ohjelmointistudio 2 / Programming Studio A", "Spring 2021",
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
@@ -73,7 +73,6 @@ public class CourseProjectAction extends AnAction {
 
   private final ExecutorService executor;
 
-  // TODO: don't hard code this list
   private static final List<CourseItemViewModel> AVAILABLE_COURSES = Arrays.asList(
       new CourseItemViewModel("O1", "Fall 2020",
           "https://grader.cs.aalto.fi/static/O1_2020/projects/o1_course_config.json"),
@@ -194,7 +193,10 @@ public class CourseProjectAction extends AnAction {
     executor.execute(() -> {
       try {
         autoInstallDone.get();
-        if (projectSettingsImported.get() && importIdeSettings && ideSettingsImported.get()) {
+        var settingsPresent = projectSettingsImported.get() &&
+                importIdeSettings &&
+                ideSettingsImported.get();
+        if (Boolean.TRUE.equals(settingsPresent)) {
           ideRestarter.run();
         }
       } catch (InterruptedException ex) {
@@ -229,7 +231,7 @@ public class CourseProjectAction extends AnAction {
 
       CourseSelectionViewModel viewModel = new CourseSelectionViewModel(AVAILABLE_COURSES);
       boolean cancelled = !dialogs.showCourseSelectionDialog(project, viewModel);
-      return cancelled ? null : new URL(viewModel.selectedCourseUrl.get());
+      return cancelled ? null : new URL(Objects.requireNonNull(viewModel.selectedCourseUrl.get()));
     } catch (MalformedURLException e) {
       // User entered an invalid URL (or the default list has an invalid URL, which would be a bug)
       logger.error("Malformed course configuration file URL", e);
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
@@ -193,9 +193,9 @@ public class CourseProjectAction extends AnAction {
     executor.execute(() -> {
       try {
         autoInstallDone.get();
-        var settingsPresent = projectSettingsImported.get() &&
-                importIdeSettings &&
-                ideSettingsImported.get();
+        var settingsPresent = projectSettingsImported.get()
+                && importIdeSettings
+                && ideSettingsImported.get();
         if (Boolean.TRUE.equals(settingsPresent)) {
           ideRestarter.run();
         }
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporter.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporter.java
@@ -15,7 +15,6 @@ import java.io.IOException;
 import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 import net.lingala.zip4j.ZipFile;
@@ -65,7 +64,7 @@ public class SettingsImporter {
     CoursesClient.fetch(ideSettingsUrl, file);
     String configPath = FileUtilRt.toSystemIndependentName(PathManager.getConfigPath());
     StartupActionScriptManager.addActionCommands(
-        Arrays.asList(
+        List.of(
             new StartupActionScriptManager.UnzipCommand(file.toPath(), Path.of(configPath)),
             new StartupActionScriptManager.DeleteCommand(file.toPath())
         )
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporter.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporter.java
@@ -66,8 +66,8 @@ public class SettingsImporter {
     String configPath = FileUtilRt.toSystemIndependentName(PathManager.getConfigPath());
     StartupActionScriptManager.addActionCommands(
         Arrays.asList(
-            new StartupActionScriptManager.UnzipCommand(file, new File(configPath)),
-            new StartupActionScriptManager.DeleteCommand(file)
+            new StartupActionScriptManager.UnzipCommand(file.toPath(), Path.of(configPath)),
+            new StartupActionScriptManager.DeleteCommand(file.toPath())
         )
     );
 
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/utils/CourseFileManager.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/utils/CourseFileManager.java
@@ -75,7 +75,7 @@ public class CourseFileManager {
    * the course file doesn't exist, {@code true} otherwise.
    *
    * @return {@code true} if the course file was successfully loaded, {@code false} if the course
-   *         file doesn't exist.
+   * file doesn't exist.
    * @throws IOException   If an IO error occurs while reading the course file.
    * @throws JSONException If the course file contains malformed JSON.
    */
@@ -129,6 +129,7 @@ public class CourseFileManager {
   /**
    * Returns the language chosen by the user for the course corresponding to the currently loaded
    * course file. This should only be called after a course file has been successfully loaded.
+   *
    * @return
    */
   @NotNull
@@ -210,7 +211,8 @@ public class CourseFileManager {
       try {
         downloadedAt = ZonedDateTime.parse(moduleObject.getString(MODULE_DOWNLOADED_AT_KEY));
       } catch (JSONException e) {
-        logger.error("Module " + moduleName + " missing 'downloadedAt' in course file", e);
+        logger
+            .error(String.format("Module %s missing 'downloadedAt' in course file", moduleName), e);
         downloadedAt = Instant.EPOCH.atZone(ZoneId.systemDefault());
       }
 
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/utils/CourseFileManager.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/utils/CourseFileManager.java
@@ -75,7 +75,7 @@ public class CourseFileManager {
    * the course file doesn't exist, {@code true} otherwise.
    *
    * @return {@code true} if the course file was successfully loaded, {@code false} if the course
-   * file doesn't exist.
+  file doesn't exist.
    * @throws IOException   If an IO error occurs while reading the course file.
    * @throws JSONException If the course file contains malformed JSON.
    */
@@ -157,7 +157,7 @@ public class CourseFileManager {
     FileUtils.writeStringToFile(courseFile, jsonObject.toString(), StandardCharsets.UTF_8);
   }
 
-  /*
+  /**
    * Returns the course file corresponding to the given project.
    */
   @NotNull
@@ -169,7 +169,7 @@ public class CourseFileManager {
         .toFile();
   }
 
-  /*
+  /**
    * Returns a JSONObject corresponding to the contents of the modulesMetadata map.
    */
   @NotNull
@@ -182,7 +182,7 @@ public class CourseFileManager {
     return modulesObject;
   }
 
-  /*
+  /**
    * Initializes local variables from the given JSON object.
    */
   private void loadFromJsonObject(@NotNull JSONObject jsonObject) throws IOException {
--- a/src/main/java/fi/aalto/cs/apluscourses/model/FileDoesNotExistException.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/model/FileDoesNotExistException.java
@@ -1,13 +1,17 @@
 package fi.aalto.cs.apluscourses.model;
 
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
 import java.nio.file.Path;
 import org.jetbrains.annotations.NotNull;
 
 public class FileDoesNotExistException extends Exception {
 
-  private static final long serialVersionUID = -8726453899914862488L;
+  private static final long serialVersionUID = 526305689216094388L;
+
   @NotNull
-  private final Path path;
+  private transient Path path;
   @NotNull
   private final String name;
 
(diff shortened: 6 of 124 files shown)


src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
```diff
@@ -194,7 +193,10 @@ public void actionPerformed(@NotNull AnActionEvent e) {
     executor.execute(() -> {
       try {
         autoInstallDone.get();
-        if (projectSettingsImported.get() && importIdeSettings && ideSettingsImported.get()) {
+        var settingsPresent = projectSettingsImported.get()
```
superseacat, line 196: `var` 😝

src/main/java/fi/aalto/cs/apluscourses/model/SubmissionStatusUpdater.java
```diff
@@ -128,4 +118,21 @@ private void run() {
...
+      submissionResult =
+          dataSource.getSubmissionResult(submissionUrl, exercise, authentication);
+      if (submissionResult.getStatus() != SubmissionResult.Status.UNKNOWN) {
+        notifier.notifyAndHide(
+            new FeedbackAvailableNotification(submissionResult, exercise), project);
+        PluginSettings.getInstance().updateMainViewModel(project);
+        return true;
+      }
+    } catch (IOException e) {
+      // Fail silently
```
superseacat, line 133: @nikke234 , seriously? 😝

src/main/resources/META-INF/plugin.xml
```diff
@@ -4,7 +4,7 @@
     <vendor email="intellij-plugin-letech@aalto.fi"
             url="https://research.cs.aalto.fi/LeTech/">LeTech Group at Aalto University</vendor>
     <!--  this is the range of supported IntelliJ IDEA version the plugin can be used with  -->
-    <idea-version since-build="203"/>
+    <idea-version since-build="203.3645.34"/>
```
superseacat, line 7: oh, I'm so unsure about this ❗❗❗

nikke234: Is there a need to do this? I think it gets patched by `gradle` in its `patchPluginXml` task, so I think we should keep it at 203.

src/main/java/fi/aalto/cs/apluscourses/model/SubmissionStatusUpdater.java
nikke234: It's my favorite thing to do when catching exceptions ;) My thinking here was that since it polls the API repeatedly, there would be no reason to notify the user if one attempt fails, but I also wouldn't be opposed to a notification

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
nikke234: Good reason to migrate to Java 11 :tada:

nikke234: I assume sonar complains that `Boolean` can be null? In that case the new version `projectSettingsImported.get() && importIdeSettings` would still throw unfortunately

superseacat: yeah, and I just bluntly coded their suggestion, please, feel free to change properly

src/main/resources/META-INF/plugin.xml
superseacat: this change is required to be able to use these two 
```java
import com.intellij.util.concurrency.annotations.RequiresReadLock;
import com.intellij.util.concurrency.annotations.RequiresWriteLock;
```

instead of deprecated `@CalledWithWriteLock` & `@CalledWithReadLock`

src/test/java/fi/aalto/cs/apluscourses/utils/DateDifferenceFormatterTest.java
```diff
@@ -15,13 +15,17 @@ public void testFormatWithLargestTimeUnit() {
...
+    assertEquals("Formatter returns the correct value for the '51 seconds' case.",
+        "51 seconds ago",
+        DateDifferenceFormatter.formatWithLargestTimeUnit(sampleTime, currentClock1));
+    assertEquals("Formatter returns the correct value for the '1 year ago' case.",
+        "1 year ago",
+        DateDifferenceFormatter.formatWithLargestTimeUnit(sampleTime, currentClock2));
+    assertEquals("Formatter returns the correct value for the '11 months ago' case.",
+        "11 months ago",
+        DateDifferenceFormatter.formatWithLargestTimeUnit(sampleTime, currentClock3));
+    assertEquals("Formatter returns the correct value for the 'just now' second case.",
```
superseacat, line 27: the other way around @Taikelenn

src/main/java/fi/aalto/cs/apluscourses/model/UrlRenderer.java
```diff
@@ -2,11 +2,13 @@
...
 import com.intellij.ide.BrowserUtil;
 import java.net.URI;
+import java.net.URISyntaxException;
+
 import org.jetbrains.annotations.NotNull;
 
 public class UrlRenderer {
 
-  public void show(@NotNull String url) throws Exception {
+  public void show(@NotNull String url) throws URISyntaxException {
```
superseacat, line 11: usually the most narrow checked `exception` is being thrown (@jaakkonarhi  check [this](https://stackoverflow.com/questions/25780619/why-do-we-follow-narrow-to-broad-exception-catching-mecahnism-in-java) as a starting point)

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/APlusAuthenticationAction.java
```diff
@@ -69,11 +69,11 @@ public void update(@NotNull AnActionEvent e) {
   public void actionPerformed(@NotNull AnActionEvent e) {
     Project project = e.getProject();
     MainViewModel mainViewModel = mainViewModelProvider.getMainViewModel(project);
-    Course course = Optional.ofNullable(mainViewModel.courseViewModel.get())
-        .map(BaseViewModel::getModel)
-        .orElse(null);
-
-    if (course == null) {
+    Course course;
```
superseacat, line 72: perhaps this could work too? IMO improves the readability a bit... @OlliKiljunen

OlliKiljunen: Yes, this is better 👍

src/main/java/fi/aalto/cs/apluscourses/model/UrlRenderer.java
OlliKiljunen: Right, a thing to remember, though: an overriding method in a subclass cannot "broaden" the exception type in the `throws` declaration (the type of the exception is said to be _covariant_, i.e., for subtypes it can be more specific but not more generic).

Anyway, plain `throws Exception` should be generally avoided.

src/main/java/fi/aalto/cs/apluscourses/dal/TokenAuthentication.java
```diff
@@ -20,8 +19,8 @@
    * @param token           Token.  The array is cloned and the argument is not cleared.
    * @param passwordStorage Password storage.  If null, the authentication won't be persisted.
    */
-  public TokenAuthentication(@NotNull String user,
-                             @NotNull char[] token,
+  protected TokenAuthentication(@NotNull String user,
+                                char[] token,
```
OlliKiljunen, line 23: Just curious: What's wrong with `@NotNull` arrays?

src/main/java/fi/aalto/cs/apluscourses/ui/courseproject/CourseSelectionView.java
```diff
@@ -82,6 +87,11 @@ public void focusGained(FocusEvent focusEvent) {
 
       @Override
       public void focusLost(FocusEvent focusEvent) {
+        try {
+          throw new OperationNotSupportedException();
+        } catch (OperationNotSupportedException e) {
+          logger.info("Operation not supported", e);
+        }
       }
```
OlliKiljunen, line 95: Do you throw and catch to get the stacktrace to the log?

Still, this makes very little sense to me.

Why we have to log events we don't care about? I guess Sonar complained of an empty method body, right? That can be solved by adding a comment there. This method should do nothing, IMO.

src/main/java/fi/aalto/cs/apluscourses/ui/module/ColoredModuleListRenderer.java
```diff
@@ -30,7 +30,8 @@ protected void customizeCellRenderer(@NotNull JList<? extends ModuleListElementV
     setToolTipText(element.getTooltip());
     setIcon(PluginIcons.A_PLUS_MODULE);
     setIconTextGap(4);
-    if (element.isUpdateAvailable()) {
+    var updateAvailable = element.isUpdateAvailable();
+    if (Boolean.TRUE.equals(updateAvailable)) {
```
OlliKiljunen, line 34: `updateAvailable` variable could perhaps be inlined.

src/main/resources/META-INF/plugin.xml
OlliKiljunen: Then I guess it's ok.

OlliKiljunen (changes requested):
Great work!

I only request a change to [this](https://github.com/Aalto-LeTech/intellij-plugin/pull/548#discussion_r578234477).

Otherwise, all good.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
```diff
@@ -73,7 +73,6 @@
 
   private final ExecutorService executor;
 
-  // TODO: don't hard code this list
```
nikke234, line 76: This shouldn't be removed since the TODO hasn't actually been fixed :smile:

src/main/java/fi/aalto/cs/apluscourses/ui/base/TextField.java
```diff
@@ -30,6 +36,11 @@ public void removeUpdate(DocumentEvent documentEvent) {
 
       @Override
       public void changedUpdate(DocumentEvent documentEvent) {
+        try {
+          throw new OperationNotSupportedException();
+        } catch (OperationNotSupportedException e) {
+          logger.info("Operation not supported", e);
+        }
```
nikke234, line 43: I wouldn't throw something just to catch it immediately, so I would either keep just the throw itself or the info log

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporter.java
```diff
@@ -66,8 +66,8 @@ public void importIdeSettings(@NotNull Course course) throws IOException {
     String configPath = FileUtilRt.toSystemIndependentName(PathManager.getConfigPath());
     StartupActionScriptManager.addActionCommands(
         Arrays.asList(
-            new StartupActionScriptManager.UnzipCommand(file, new File(configPath)),
-            new StartupActionScriptManager.DeleteCommand(file)
+            new StartupActionScriptManager.UnzipCommand(file.toPath(), Path.of(configPath)),
```
superseacat, line 69: take a look @OlliKiljunen

src/main/java/fi/aalto/cs/apluscourses/ui/base/TextField.java
OlliKiljunen: Oh, this is a similar issue I commented on elsewhere. It cannot throw because this method is called when that event triggers. It should do nothing and return normally and quietly.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporter.java
OlliKiljunen: You changed `File` to `Path` (makes sense) but also reversed the order of commands? Why?

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
superseacat: I'd remove as the issue for this exists: https://github.com/Aalto-LeTech/intellij-plugin/issues/522 , overall, I suggest to create **ISSUEs** instead of such `TODOs`, on the other hand, the dead man's word ☠️

src/main/java/fi/aalto/cs/apluscourses/intellij/model/SettingsImporter.java
OlliKiljunen: Sorry, I looked wrong. nevermind. all good.

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/CourseFileManager.java
```diff
@@ -75,7 +75,7 @@ public synchronized void createAndLoad(@NotNull URL courseUrl,
    * the course file doesn't exist, {@code true} otherwise.
    *
    * @return {@code true} if the course file was successfully loaded, {@code false} if the course
-   *         file doesn't exist.
+  file doesn't exist.
```
nikke234, line 78: This is missing an asterisk at the beginning of the line

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
nikke234: Yeah it's good to have an issue, but TODO's are also easy to search for in the code and removing gives the impression that the issue has been fixed, but it hasn't :smile: So I would rather keep it (it's almost like cheating just to silence sonar)

OlliKiljunen: I'd say both TODOs and issues have their uses (but as you reasoned here, no both of them should be applied to the same problem).

TODOs can be used to keep track of minor refactoring needs (that shouldn't have a ticket of their own).

src/main/java/fi/aalto/cs/apluscourses/dal/TokenAuthentication.java
superseacat: `Primitive type members cannot be annotated`

```java
org.jetbrains.annotations @Documented 
@Retention(RetentionPolicy.CLASS) 
@Target({ElementType.METHOD,ElementType.FIELD,ElementType.PARAMETER,ElementType.LOCAL_VARIABLE,ElementType.TYPE_USE}) 
public interface NotNull
extends annotation.Annotation
```

`An element annotated with NotNull claims null value is forbidden to return (for methods), pass to (parameters) and hold (local variables and fields). Apart from documentation purposes this annotation is intended to be used by static analysis tools to validate against probable runtime errors and element contract violations.
`

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/CourseFileManager.java
superseacat: go try adding it! 😈

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
superseacat: create an issue, put it into the comments everywhere as a pointer???

OlliKiljunen (approved):
:+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
superseacat: there'll be more PRs on this topic soonish, so you can plug in and fix this at any time later too @nikke234

src/main/java/fi/aalto/cs/apluscourses/dal/TokenAuthentication.java
Taikelenn: But is `char[]` a primitive?

nikke234: char would be, but arrays not

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/CourseProjectAction.java
nikke234: Yeah will do at some point :+1:

Taikelenn (approved):
Very good changes, especially if they make our friend Sonar happy :>
A few remarks:
1. Inside ColoredModuleListRenderer.java the change to `Boolean.TRUE.equals` shouldn't be necessary, because the `Module` class returns a normal `boolean` - it's the view model that overzealously "upgrades" the boolean to its nullable variant
2. Perhaps arrays of primitives could be annotated `char @NotNull[]`, which means that the array instance cannot be null. `@NotNull char[]` would mean that no _element_ of the array could be null, which is indeed impossible and a valid warning.
3. We seem to have preliminarily agreed not to care about transient/serializable, but on the other hand adding `serialVersionUID` suggests that we do _somewhat_ care about serialization 😄

nikke234:
>We seem to have preliminarily agreed not to care about transient/serializable, but on the other hand adding serialVersionUID suggests that we do somewhat care about serialization

I think we agreed to either turn off the warning completely or then carefully consider serialization instead of just adding `transient` to silence the warning, so this takes the approach of careful consideration :slightly_smiling_face:

nikke234 (approved):

src/main/java/fi/aalto/cs/apluscourses/dal/TokenAuthentication.java
superseacat: `char @NotNull [ ] array` 😉 
thanks to @nikke234

superseacat:
> > We seem to have preliminarily agreed not to care about transient/serializable, but on the other hand adding serialVersionUID suggests that we do somewhat care about serialization
> 
> I think we agreed to either turn off the warning completely or then carefully consider serialization instead of just adding `transient` to silence the warning, so this takes the approach of careful consideration

This is what Sonar says:
_"Fields in a "Serializable" class should either be transient or serializable_
_Code smell_
_Critical
java:S1948_
 _Fields in a Serializable class must themselves be either Serializable or transient even if the class is never explicitly serialized or deserialized. For instance, under load, most J2EE application frameworks flush objects to disk, and an allegedly Serializable object with non-transient, non-serializable data members could cause program crashes, and open the door to attackers. In general a Serializable class is expected to fulfil its contract and not have an unexpected behaviour when an instance is serialized.
This rule raises an issue on non-Serializable fields, and on collection fields when they are not private (because they could be assigned non-Serializable values externally), and when they are assigned non-Serializable types within the class."_

this is a bit different POV on serialization. Also, @nikke234 @Taikelenn adding did not satisfy the **Sonar God**, so `transient` or silencing the rule is still on the agenda!?

nikke234:
>>> We seem to have preliminarily agreed not to care about transient/serializable, but on the other hand adding serialVersionUID suggests that we do somewhat care about serialization

>>I think we agreed to either turn off the warning completely or then carefully consider serialization instead of just adding transient to silence the warning, so this takes the approach of careful consideration

>This is what Sonar says:
"Fields in a "Serializable" class should either be transient or serializable
Code smell
Critical
java:S1948
Fields in a Serializable class must themselves be either Serializable or transient even if the class is never explicitly serialized or deserialized. For instance, under load, most J2EE application frameworks flush objects to disk, and an allegedly Serializable object with non-transient, non-serializable data members could cause program crashes, and open the door to attackers. In general a Serializable class is expected to fulfil its contract and not have an unexpected behaviour when an instance is serialized.
This rule raises an issue on non-Serializable fields, and on collection fields when they are not private (because they could be assigned non-Serializable values externally), and when they are assigned non-Serializable types within the class."

>this is a bit different POV on serialization. Also, @nikke234 @Taikelenn adding did not satisfy the Sonar God, so transient or silencing the rule is still on the agenda!?

Ah right, of course. So we have some class that is serializable (most likely because it extends another class that is serializable), but then our class has some fields in it that are NOT serializable and sonar complains.

I don't know which classes sonar is complaining about, but if we don't control the classes of the class fields, then we can't make them serializable. The only option then is creating our own class that is serializable or marking the field as transient.

Given that we don't use serialization at all, I propose the following:
- Turn off the warning in sonar.
- Remove ALL serial version  UIDs and ALL uses of `transient` from our code.

nikke234:
Well, on closer look, it seems that `ReplConfigurationFormModel` (which we control) is the only class that sonar complains about, so I actually would suggest we make that serializable instead and don't turn off the warning. I would also double-check that all of the classes that have serial version UIDs also actually serializable.

OlliKiljunen:
I found in total of 8 fields that Sonar warns of and 10 fields marked as transient. Here they are:

Class | Serializable supertype | Type of the field | Field | Status
--- | --- | --- | --- | ---
BaseListView | JComponent | Object | popupMenuLock | code smell
BaseListViewModel | AbstractListModel | List | elements | code smell
FileDoesNotExistException | Throwable | Path | path | code smell
IconListCellRenderer | JComponent | Icon | icon | code smell
IconListCellRenderer | JComponent | Function | getText | code smell
ReplConfigurationForm | JComponent | ReplConfigurationFormModel | model | code smell
OurComboBox | JComponent | TwoWayBindable| selectedItemBindable | code smell
UnexpectedResponseException | Throwable | HttpResponse | response | code smell 
TextField | JComponent | TwoWayBindable | textBindable | transient
BaseListViewModel | AbstractListModel | List | selectionModel | transient
BaseTreeViewModel | - | SelectableNodeViewModel | selectedItem | transient, NOSONAR
CheckBox | JComponent | TwoWayBindable |  isCheckedBindable | transient
CheckBox | JComponent | Bindable |  isEnabledBindable | transient
TemplateLabel | JComponent | Bindable | templateArgumentBindable | transient, unused
TreeView | JComponent | Set | nodeAppliedListeners | transient
TreeView | JComponent | Object | popupMenuLock | transient
TreeView | JComponent | BaseTreeViewModel | viewModel | transient
TreeView | JComponent | Object | viewModelLock | transient

I'll come back later with my suggestions of what should be done to them...

superseacat:
> I would also double-check that all of the classes that have serial version UIDs also actually serializable.

added the serial version ID based on Hector's advice (via IDEA "suggest action" and activate additional setting), but re-checking never hurts

stellatsv:
In `ModuleTest.java` there is a proposal to group the last 3 tests into one `ParameterizedTest`. This requires JUnit 5.4 (currently we are at 4.12). I am not sure if there would be any consequences from upgrading but we could consider it.

OlliKiljunen:
I've updated [my earlier comment](https://github.com/Aalto-LeTech/intellij-plugin/pull/548#issuecomment-781570681) with some ideas of how some of the "problems" should be fixed.

But here's my most important message: **We shouldn't care of serialization of classes extending Swing classes.** The support of serialization in Swing itself is so poor no framework should rely on that. Furthermore, IntelliJ's own Swing extensions (like `JBList`) are "broken" in that sense.

So, I suggest (somewhat similarly as @nikke234 earlier did) that **we remove `transient` keywords and `serialVersionUID`s from classes that extend Swing classes.**

After that, we have two cases where the class is a subclass of `Exception`. Those we could handle properly, i.e. use serializable types in them.

IMO, **we can change Sonar rules to silence this type of code smells**, as it is most expected to occur in Swing related classes. Let's, however, try to remember to take that into account in other cases, like in exceptions, where it could, at least in theory, matter.

If we don't want to silence the rule, one possibility is to make all "problematic" Swing-extending classes implement methods `private void readObject(ObjectInputStream)` and `private void writeObject(ObjectOutputStream)` with one-liners that just throw, for example, `NotSerializableException`. That would both satisfy Sonar and make serialization (if it ever happened) fail fast and explicitly, which is desirable.

superseacat:
> In `ModuleTest.java` there is a proposal to group the last 3 tests into one `ParameterizedTest`. This requires JUnit 5.4 (currently we are at 4.12). I am not sure if there would be any consequences from upgrading but we could consider it.

for some silly reason, `JUnit 4.x` (perhaps @superseacat 's prior experience) this particular version was chosen. However, I see little reason why not to migrate to latest JUnit. It actually has some nice features I've bumped into earlier and that would make the tests more lightweight, so I vote **"for"** it. 👍🏻

superseacat:
> > In `ModuleTest.java` there is a proposal to group the last 3 tests into one `ParameterizedTest`. This requires JUnit 5.4 (currently we are at 4.12). I am not sure if there would be any consequences from upgrading but we could consider it.
> 
> for some silly reason, `JUnit 4.x` (perhaps @superseacat 's prior experience) this particular version was chosen. However, I see little reason why not to migrate to latest JUnit. It actually has some nice features I've bumped into earlier and that would make the tests more lightweight, so I vote **"for"** it. 👍🏻

does not look terribly complicated: https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4
Thus, in order not to overwhelm this particular PR I suggest making a separate issue/PR for this. What do you think @stellatsv?

superseacat:
> IMO, **we can change Sonar rules to silence this type of code smells**, as it is most expected to occur in Swing related classes. Let's, however, try to remember to take that into account in other cases, like in exceptions, where it could, at least in theory, matter.

an addition to this^: we can silence only the currently highlighted issues and continue with the rule active. silencing the particular issue is a matter of 2-click and a comment (and then the comment trace remains, like **[here](https://sonarcloud.io/project/issues?id=Aalto-LeTech_intellij-plugin&open=AXLB9d56Sv_mbHSxVbKx&resolutions=WONTFIX)**)

src/main/java/fi/aalto/cs/apluscourses/model/FileDoesNotExistException.java
```diff
@@ -24,4 +30,15 @@ public Path getPath() {
   public String getName() {
     return name;
   }
+
+  private void writeObject(ObjectOutputStream oos) throws IOException {
```
superseacat, line 34: kinda like this @Aalto-LeTech/apluscourses-plugin-team ?

superseacat:
> implement methods `private void readObject(ObjectInputStream)` and `private void writeObject(ObjectOutputStream)` with one-liners that just throw, for example, `NotSerializableException`

as half of the application will crash-and-burn in case of serialization anyway, I see little reason to add boilerplate, perhaps just silence the cases in sonar (thus the rule remains active)

src/main/java/fi/aalto/cs/apluscourses/model/FileDoesNotExistException.java
OlliKiljunen: This looks like a nice use of `readObject`/`writeObject` but...

- The string representation returned by `Path::toString` is platform specific (`/` vs `\`). Generating a path from that string may not be possible on another platform. Thus, this serialization from one machine to another may not work.
- This adds a considerable amount of must-be-tested code while providing very little value.

So...


Would it just be safer and easier to store the path as a string?

nikke234: I would in general suggest that we don't store paths as a string just to support serialization which we do not use, this is maybe a special case since it's just an exception but still

superseacat: I'll just make `Path` and `HttpResponse` `transient` 😝

stellatsv:
> > > In `ModuleTest.java` there is a proposal to group the last 3 tests into one `ParameterizedTest`. This requires JUnit 5.4 (currently we are at 4.12). I am not sure if there would be any consequences from upgrading but we could consider it.
> > 
> > 
> > for some silly reason, `JUnit 4.x` (perhaps @superseacat 's prior experience) this particular version was chosen. However, I see little reason why not to migrate to latest JUnit. It actually has some nice features I've bumped into earlier and that would make the tests more lightweight, so I vote **"for"** it. 👍🏻
> 
> does not look terribly complicated: https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4
> Thus, in order not to overwhelm this particular PR I suggest making a separate issue/PR for this. What do you think @stellatsv?

Sounds good, I will make a separate issue.

Issue #244, opened 2020-07-23
Auto-installed modules sometimes show "Not installed"


OlliKiljunen:
This bug wasn't completely fixed with #341. So it's reopened.

---

From [a Slack discussion](https://apluslms.slack.com/archives/CSDV9CA31/p1612296643039100) from 2021-02-03:
@OlliKiljunen:
>[...] I think I found the root cause of the "auto-installed show not installed" issue. It's probably because `CourseProjectAction` uses one `Course` object but `MainViewModelUpdater` creates another. The module that see itself installed is not the one that is rendered on the list.

>I tried to "quickly" fix this in two different ways:
>1. make `StateMonitor`s be reused by modules when the `Course` object changes
>2. make `startAutoInstalls` use `Course` from the newly-updated main view model
>Neither of these worked nicely. If you wanna see my code, send a message. But it's probably better try something else (that is, fix it properly).

@nikke234:
>Oh yeah, that's a good find!
>Maybe it's time for an update that just mutates the existing course instance in a thread-safe manner instead of creating new ones. For an example, `CourseProjectAction` could create the main view model, and the scheduled updater would be just responsible for updating the existing view model instead of creating it as well. It would also remove the constant adding and removing of listeners related to each new created course instance. (And it should also make it easier to make the current hacky update preventing mechanism non-static.)

---

## Original description of the issue:

Turn project into A+ project. O1Library is installed but in the list of modules, it still shows "Not installed".

---

This bug actually reflects a more major issue underneath, an issue that could cause more harmful errors (like local changes overwritten without user acceptance when updating modules). Luckily, these errors are unlikely to happen in normal use.

The issue is that when main view model is updated (periodically), a new model object instances are created too. Because we use `stateMonitor`s as concurrency control for module installation, and each model object instance has their own `stateMonitor`, we could end up in a situation where two instances of, say, O1Library module are downloading/installing themselves (into the same directory, that is) at the same time. The if-clause that skips main view model update if some of the components are active, is likely to prevent this from happening but it's not guaranteed.

---

**Easy fix** for this particular case (that auto-installed modules show initially "Not installed"): when starting main view model auto-updater, do it so that it's not run immediately but only after the first cycle 15 minutes.

**Proper fix**: Re-desing. A lot.

[...]

PR #560, opened 2021-02-23
Improved course updating mechanism

nikke234:
# Description of the PR

This is a refactor that rewrites the course updating mechanism. Among other improvements, this fixes the infamous bug about O1Library showing as not installed (#244). It also removes all the problems associated with the current updating approach of throwing objects away and creating new ones.

It uses a simple approach of updating the existing course instance and triggering an event after each update. All "interested parties" can then listen to said event for course updates.

Added or removed modules aren't detected by the update, but that could be implemented as a separate PR, since it's very rare anyways.


OlliKiljunen (commented):
In overall, this looks like a great improvement. I didn't find any flaws in this design (of course, it's difficult to judge when it's still WIP).

I didn't focus on threading at this stage ( if you want, I can do it. I definitely will, when it is ready. ).

Thanks for working on this. All I can say now: Please go on!

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseProject.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.intellij.openapi.project.Project;
+import fi.aalto.cs.apluscourses.intellij.notifications.DefaultNotifier;
+import fi.aalto.cs.apluscourses.intellij.services.PluginSettings;
+import fi.aalto.cs.apluscourses.model.Course;
+import fi.aalto.cs.apluscourses.utils.Event;
+import java.net.URL;
+import org.jetbrains.annotations.NotNull;
+
+public class CourseProject {
```
superseacat, line 11: I haven't worked with the code for a while, so got a bit confused here... Initially I was like "wow, finally a `CourseProject` class", but then I started to think what is its interrelation with others? 😉 

(continues after the pics)
I've made some old-fashioned pen-and-paper sketches, to support the discussion:
IMG_20210225_091220
IMG_20210225_091427

As `IntelliJCourse` is a `Course`, can I claim that `CourseProject` has a `IntelliJCourse` 🤣 ? In other words, is there a way to perhaps simplify and flatten the structure or maybe introduce more distinctive names and `JavaDoc`s for classes? For example this is `AplusProject` class, it is a representation of ..., use it for .., well, actually, this is a good example, why should it be used (and what does it name tells in the context of the plugin)?

superseacat: additionally, what is the relation between `CourseProject` and a `Project`, should there be some?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseProject.java
```diff
@@ -0,0 +1,44 @@
...
+  private final CourseUpdater courseUpdater;
+
+  @NotNull
+  public final Event courseUpdated;
+
+  /**
+   * Construct a course project from the given course, course configuration URL (used for updating),
+   * and project.
+   */
+  public CourseProject(@NotNull Course course, @NotNull URL courseUrl, @NotNull Project project) {
```
superseacat, line 26: additionally, seems that the `CourseProject` still needs a `Project`, but only to pass it over to the `CourseUpdater`, this is not exactly straightforward...

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
```diff
@@ -176,35 +177,32 @@ public MainViewModel getMainViewModel(@Nullable Project project) {
...
-      return new MainViewModel(exerciseFilterOptions);
-    });
-
-    mainViewModelUpdaters.computeIfAbsent(key, projectKey -> {
-      MainViewModelUpdater mainViewModelUpdater = new MainViewModelUpdater(
-          mainViewModel, project, MAIN_VIEW_MODEL_UPDATE_INTERVAL,
-          new DefaultNotifier(), IntelliJPasswordStorage::new);
-      mainViewModelUpdater.start();
-      return mainViewModelUpdater;
+  public void registerCourseProject(@NotNull Project project,
```
superseacat, line 190: perhaps this could be divided into smaller chunks (like two) and I would love to see a unit (or any other type) of test for this particular method ❤️

nikke234: Good idea!

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseProject.java
nikke234: Nice drawing :smile: And the understanding from the drawing is completely correct.

I now realize that the name `CourseProject` is pretty confusing. As you mentioned, it doesn't even contain a reference to a project, but just passes it over to course updater.

What it really is, is a combination of a course and an updater that regularly updates the course. I'm not sure yet what a good name would be :thinking: 

I'll at least add JavaDocs for the classes, because as you mention, it's getting a bit confusing :smile:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
```diff
@@ -0,0 +1,147 @@
...
+   * interrupt and exited. Note, that making the 'run' method 'synchronized' would cause a deadlock,
+   * because 'restart' is synchronized as well.
+   */
+  private final Object runLock = new Object();
+
+  private void run() {
+    synchronized (runLock) {
+      while (true) { //  NOSONAR
+        try {
+          var newCourse = fetchCourse();
```
OlliKiljunen, line 88: Nit:  it's a bit misleading that we create a `Course` object only to copy data from it to our original `Course` object, isn't it? Would it be possible that we wouldn't construct a new object but only the fetched modules as a list or something?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
```diff
@@ -0,0 +1,147 @@
...
+    this.updateInterval = updateInterval;
+  }
+
+  /**
+   * Starts or restarts this updater. This method is thread safe and can be called from multiple
+   * threads.
+   */
+  public synchronized void restart() {
+    if (thread != null) {
+      thread.interrupt();
```
OlliKiljunen, line 70: Would it be safest to join the old thread before continuing? Surely, the new thread cannot do anything before the old one releases the lock but still, it could help reasoning things out. Like, then it would be guaranteed that after `restart` returns, the old thread cannot trigger events, and so on.

On the other hand, it would make this method more prone to unnoticed concurrency bugs causing deadlock.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
OlliKiljunen: Constructing an `IntelliJCourse` object is also only reason that this class must be in `intellij` package, right?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
```diff
@@ -0,0 +1,147 @@
...
+  private static URI urlToUri(@NotNull URL url) {
+    try {
+      return url.toURI();
+    } catch (URISyntaxException e) {
+      return null;
+    }
+  }
+
+  private static Map<URI, Module> createUriToModuleMapping(@NotNull Course course) {
+    // hashCode and equals of URLs can cause DNS lookups, so URIs are preferred in maps
```
OlliKiljunen, line 140: That's... interesting. o.O Nice catch, anyways!

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
OlliKiljunen: Btw, would it be reasonable to check if the thread is interrupted after the fetching is done?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+      }
+    }
+  }
+
+  @Nullable
+  private Course fetchCourse() {
+    try {
+      // Read-lock needed, since we access the project
+      return ReadAction.compute(() -> Course.fromUrl(courseUrl, new IntelliJModelFactory(project)));
```
OlliKiljunen, line 114: Do you know whether `ReadAction::compute` raises `InterruptedException`?

OlliKiljunen (approved):
A very welcome improvement well implemented! I left a few questions that you can consider but if you are confident this works, there's nothing I'd require to be changed, so approved. :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/services/PluginSettings.java
```diff
@@ -176,35 +177,32 @@ public MainViewModel getMainViewModel(@Nullable Project project) {
...
+      mainViewModel.courseViewModel.set(new CourseViewModel(courseProject.getCourse()));
+      courseProject.courseUpdated.addListener(
+          mainViewModel.courseViewModel, ObservableProperty::valueChanged);
+      courseProject.getCourseUpdater().restart();
+      return courseProject;
     });
+    var passwordStorage = new IntelliJPasswordStorage(courseProject.getCourse().getApiUrl());
+    TokenAuthentication.Factory authenticationFactory =
+        APlusTokenAuthentication.getFactoryFor(passwordStorage);
+    mainViewModel.readAuthenticationFromStorage(passwordStorage, authenticationFactory);
```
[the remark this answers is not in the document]
nikke234: Happy to hear that, as that was also a longer-term idea here :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
nikke234: > Btw, would it be reasonable to check if the thread is interrupted after the fetching is done?
Yes, very good idea, the quicker it responds to interrupts the better

nikke234: > Nit: it's a bit misleading that we create a Course object only to copy data from it to our original Course object, isn't it? Would it be possible that we wouldn't construct a new object but only the fetched modules as a list or something?

Yeah true, this is a pretty lazy way of reusing the existing parsing mechanism :smile: I'll either make the relevant module JSON parsing method from `Course` public or just write some JSON parsing here, and only deal with a list of modules instead of constructing a course

nikke234: > Constructing an IntelliJCourse object is also only reason that this class must be in intellij package, right?

Also the notifications make it part of `intellij` (although I'm sure there are some who would want to add a layer of abstraction over notifications to separate more code from the platform :wink:)

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
nikke234: Interesting idea! That would indeed give the same guarantees as the run lock mechanism, and it would also make it very clear that the previous thread is completely done with everything. It would actually render the current run lock unnecessary I think.

The main difference with the run lock would be that with a `join`, the `restart` method (and by extension its caller) now blocks until the running thread has reacted to the interruption. With the current locking mechanism, only the newly created thread is the one that blocks until the old thread is done.

`restart` blocking until the thread has reacted to the interruption makes me a bit nervous, because we fetch network resources in the update loop, which could cause unpredictable delays. Apache's HTTP library surely has some reasonable default timeouts, but I haven't looked into it

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
nikke234: This WTF-fact was actually noted by sonar :smile: One of the rare moments where I feel grateful for its existence

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
nikke234: Also now realized that not constructing an `IntelliJCourse` would also conveniently avoid the read lock, which is always nice

PR #569, opened 2021-02-26
Automate UI testing for Mac

superseacat:
# Description of the PR

**(will squash on merge)**

**it's OK for `build` task to fail, as it's being executed from my fork and thus does not have a key required for sonar to analyze**

This is a solution to our `macos-latest` struggle by [**@nizienko**](https://github.com/nizienko). The discussion occurs on JB's [forum](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360010463479-using-runIdeForUiTests-in-GitHub-Actions-on-Windows-MacOs).

I strongly advise taking a look a the implementation:
* added logging
* added taking storing screenshots (as artifacts). see the:
  * [job summary page](https://github.com/nizienko/intellij-plugin/actions/runs/600345658) (bottom of the page, to download artifacts)
  * [corresponding build.yml file](https://github.com/nizienko/intellij-plugin/actions/runs/600345658/workflow)
* multiple attempts to trigger an action (on `HeavyWeightWindow`)


nikke234:
Thanks for the help @nizienko!

src/e2e/kotlin/fi/aalto/cs/apluscourses/e2e/steps/CommonSteps.kt
```diff
@@ -2,78 +2,47 @@ package fi.aalto.cs.apluscourses.e2e.steps
...
+                        button("Finish").click()
+                    }
+                }
+            }
+            step("Add OpenJDK 11") {
+                with(dialog("Project Structure", Duration.ofSeconds(20))) {
+                    sidePanel().findText("Project").click()
+                    customComboBox("\u001BProject SDK:").dropdown()
+                    with(heavyWeightWindow()) {
+                        attempt(2) {
```
OlliKiljunen, line 31: If this is a common problem with comboboxes/heavyweightwindows, `attempt(2)` would be better to be put "under the hood" so it doesn't clutter test procedures

src/e2e/resources/log4j2.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="WARN">
```
OlliKiljunen, line 2: I feel like something like this file could be useful in `main` and `test` too...?

OlliKiljunen (approved):

src/e2e/kotlin/fi/aalto/cs/apluscourses/e2e/steps/CommonSteps.kt
jaakkonarhi: is there a possibility it does the first click successfully but not the second one, and then on the second attempt it closes the thing it opened? maybe it should be 3 attempts in that case

src/e2e/kotlin/fi/aalto/cs/apluscourses/e2e/steps/CommonSteps.kt
```diff
@@ -2,78 +2,47 @@ package fi.aalto.cs.apluscourses.e2e.steps
...
-      }
-      step("Add OpenJDK 11") {
-        with(dialog("Project Structure", Duration.ofSeconds(20))) {
-          sidePanel().findText("Project").click()
-          customComboBox("\u001BProject SDK:").dropdown()
-          with(heavyWeightWindow()) {
-            findText("Add SDK").click()
-            HierarchyDownloader.catchHierarchy {
-              heavyWeightWindow().findText("Download JDK...").click()
+    fun createProject() {
```
jaakkonarhi, line 14: Do I have to reindent my code?

jaakkonarhi (approved):

Taikelenn (approved):

src/e2e/resources/log4j2.xml
superseacat: sure, I'll add, and let's see if any benefit! 👍🏻

src/e2e/kotlin/fi/aalto/cs/apluscourses/e2e/steps/CommonSteps.kt
superseacat: this is something I'm not sure about actually... my indentation happens to be identical to the author's. perhaps this could be solved with a common checkstyle https://github.com/Aalto-LeTech/intellij-plugin/issues/570?

src/e2e/kotlin/fi/aalto/cs/apluscourses/e2e/steps/CommonSteps.kt
superseacat: I don't think it executes twice if successful on the first try. After all, it's the same code executed across all the environments (Linux, Mac), so if the assumption would hold, then for Linux it would open the window, close the window, fail.

Issue #524, opened 2021-02-11
Replace text when no assignments to show


jaakkonarhi:
![image](https://user-images.githubusercontent.com/34682147/107655542-4bed5b80-6c8c-11eb-861f-0083b3c352f9.png)

If all the assignments are filtered out this text shows. It could be replaced with something like "You've completed everything" or "No assignments to show" or a loading bar when loading them for the first time (#519)

OlliKiljunen:
Similarly, when opening an existing A+ project, the modules list's message says the project should be turned to A+ project until indexing is done.

stellatsv:
When the user is not authenticated the filtering options look like this:
![image](https://user-images.githubusercontent.com/51020697/110694176-f9528100-81f0-11eb-899b-8a3c2ce40418.png)

It would be better if the icons were disabled in that case since there are no available assignments to interact with.

PR #572, opened 2021-02-28
Assignments text

stellatsv:
# Description of the PR
This PR relates to #524. It contains a fix for the bug. 

In this issue, there is also mention of the existing problem that when IntelliJ is loading, the `Modules` and `Assignments` views display the wrong text. This is corrected by the time the application is ready for use and the wrongly suggested activity is safe to perform. 
### Attempts
I tried setting `InitializationActivity` as a `postStartupActity` which delayed the activity's execution a bit but it was not enough on its own. I also tried using a  [ProjectManagerListener](https://github.com/JetBrains/intellij-community/blob/master/platform/projectModel-api/src/com/intellij/openapi/project/ProjectManagerListener.java), its `projectOpened` method in particular, and from there set a boolean indicating that the project is ready, but the method got called earlier than what is needed. 

Using boolean values similar to what was used to display the correct empty text is what has helped the most so far. There is however one issue remaining with this approach when it comes to Modules view:
- When the project is an A+ project, the wrong text is visible for a little bit before displaying the tree. This is because a null `CourseViewModel` is passed to `ModulesView` during startup and there is currently no way to know if the project is an A+ project in order to display the correct text. 

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.


src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -115,12 +119,22 @@ public void setAuthentication(Authentication auth) {
...
       return;
     }
     Optional.ofNullable(passwordStorage)
         .map(PasswordStorage::restorePassword)
         .map(factory::create)
         .ifPresent(this::setAuthentication);
+    exercisesViewModel.get().setProjectReady(true);
+    //courseViewModel.get().setProjectReady(true);
+    if (!exercisesViewModel.get().isAuthenticated()) {
+      this.updateExercises(); // When authentication is null mainViewModel is not updated?
```
OlliKiljunen, line 133: > When authentication is null mainViewModel is not updated?

Correct. Calling `setAuthentication` a few lines above causes `updateExercise` to be executed*. However, it's only called when the authentication read from the password storage is not null (note `ifPresent`).

---

*) Because of this line in the constructor:
```java
authentication.addValueObserver(this, MainViewModel::updateExercises);
```

OlliKiljunen (changes requested):
I see this turned out to be a tricky one... Thanks for putting effort into this!

Your solution seems to work. Good! :+1: However, I'm afraid it adds quite too much complexity to solve such a "minor" thing ("minor" as its effect to the UI). And what I mean by complexity here, is, for example, that it's very difficult to reason whether those Boolean state variables in view model classes are correctly set in all cases (mutable states in general are difficult). Maintainer of the code would probably find it very hard to know if they have to do something for those manually when adding/removing some code from exercise updating logic. Concurrency brings an additional layer of uncertainty.

The reason why this turned out to be so tricky may be that the procedure that handles creation of main view model etc. was/is poorly designed and even flawed. @nikke234's open PR #560 improves and, especially, _simplifies_ that a lot, which may help it solve this issue too.

So what I suggest is that you make those minor changes I requested in my inline comments (also to make this PR easier to understand for further reviewers), then we wait for #560 to be merged (should be soon) and see together how things look after that (let's add @nikke234 as a reviewer then, too).

You've put a great effort in this task and that is, in any case, highly valuable for the project. Hopefully, you're not already totally exhausted of fighting this bug (and if you are, it's ok if you want to work with other tickets before returning to this).

By the way, I couldn't help noticing a not-too-big refactoring need closely related to this PR. It's when that `emptyModel` is created in `MainViewModel::updateExercises`. It would be more OO design, to have `EmptyExercisesTreeViewModel` as a separate subclass of `ExercisesTreeViewModel`. Then `isEmptyTextVisible` field wouldn't be required because the base class would always return true for `isEmptyTextVisible()` and the subclass would override it to return always false. Would you like to make that change (as a part of this PR, or separately)? Not necessary, just an idea.

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/ExercisesTreeViewModel.java
```diff
@@ -11,7 +11,11 @@
 public class ExercisesTreeViewModel extends BaseTreeViewModel<List<ExerciseGroup>>
         implements Searchable {
 
-  private boolean isEmptyTextVisible = false;
+  private boolean isEmptyTextVisible;
+
+  private boolean isAuthenticated;
+
+  private  boolean isProjectReady;
```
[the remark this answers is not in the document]
OlliKiljunen: Hence, all calls to `ExercisesTreeViewModel::setProjectReady` should be removed as they have no effect anywhere. Please correct me if I'm wrong.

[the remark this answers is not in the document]
stellatsv: The getter is used in `ExercisesView`.

src/main/java/fi/aalto/cs/apluscourses/presentation/exercise/EmptyExercisesTreeViewModel.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class EmptyExercisesTreeViewModel extends ExercisesTreeViewModel {
+
+  /**
+   * Construct an exercises tree view model from the given exercise groups.
+   */
+  public EmptyExercisesTreeViewModel(@NotNull List<ExerciseGroup> exerciseGroups,
+                                     @NotNull Options filterOptions) {
+    super(exerciseGroups, filterOptions);
+  }
```
nikke234, line 17: Since it's supposed to be empty, I would suggest that this constructor doesn't take a `List` parameter at all, and instead it always passes an empty list to the superclass constructor

src/main/resources/META-INF/plugin.xml
```diff
@@ -21,7 +21,7 @@
     <depends>org.intellij.scala</depends>
 
     <extensions defaultExtensionNs="com.intellij">
-        <backgroundPostStartupActivity implementation="fi.aalto.cs.apluscourses.intellij.activities.InitializationActivity"/>
+        <postStartupActivity implementation="fi.aalto.cs.apluscourses.intellij.activities.InitializationActivity"/>
```
nikke234, line 24: Do you have a link or something where I could read about the differences of these two? My googling isn't bringing results :smile:

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -69,12 +70,14 @@ private void updateExercises() {
...
     Authentication auth = authentication.get();
     if (course == null || auth == null) {
-      ExercisesTreeViewModel emptyModel =
-          new ExercisesTreeViewModel(new ArrayList<>(), new Options());
+      ExercisesTreeViewModel emptyModel = new ExercisesTreeViewModel(new ArrayList<>(),
+              new Options());
       if (course == null) {
-        emptyModel.setEmptyTextVisible(true);
+        emptyModel = new EmptyExercisesTreeViewModel(new ArrayList<>(), new Options());
       }
```
nikke234, line 77: I think the inner `if (course == null)` could be removed and just directly assign `new EmptyExercise...` to `emptyModel`

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -84,6 +87,8 @@ private void updateExercises() {
       List<ExerciseGroup> exerciseGroups = dataSource.getExerciseGroups(course, points, auth);
       inGrading.forEach((id, exercise) -> setInGrading(exerciseGroups, id));
       exercisesViewModel.set(new ExercisesTreeViewModel(exerciseGroups, exerciseFilterOptions));
+      exercisesViewModel.get().setAuthenticated(auth != null);
```
nikke234, line 90: `auth` is never `null` at this point, because the earlier `if (course == null || auth == null)` ends with a return.

nikke234 (commented):
Nice! I left a few comments about code clarity. It's a bit tricky to reason and be confident that all the booleans get set correctly, so I'll double-check that and see if the newest merges could simplify anything here

src/main/resources/META-INF/plugin.xml
stellatsv: https://plugins.jetbrains.com/docs/intellij/plugin-components.html#project-open

nikke234: Thanks!

Running that initialization activity on the EDT (more familiarly the "UI thread") should be avoided, because the UI freezes until the activity is done. This is especially important in this case, because the initialization activity fetches network resources, which can have long and unpredictable delays. While running the code I occasionally got log entries like this:
```
2021-03-14 23:04:32,865 [  20349]   WARN - .diagnostic.PerformanceWatcher - UI was frozen for 12581ms, details saved to /l/drosden1/intellij-plugin/build/idea-sandbox/system/log/threadDumps-freeze-20210314-230425-IC-203.7148.57-ObservableReadWriteProperty.get-12sec
```
Then again running it in the background makes the rest more difficult. The more I work on this issue, the more I realize how complicated it is to solve it :smile: I'm personally at least tempted to reconsider whether this should be "release-blocking", let's discuss in tomorrow's daily.

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -69,11 +70,13 @@ private void updateExercises() {
     }
     Authentication auth = authentication.get();
     if (course == null || auth == null) {
-      ExercisesTreeViewModel emptyModel =
-          new ExercisesTreeViewModel(new ArrayList<>(), new Options());
+      ExercisesTreeViewModel emptyModel = new ExercisesTreeViewModel(new ArrayList<>(),
+          new Options());
       if (course == null) {
-        emptyModel.setEmptyTextVisible(true);
+        emptyModel = new EmptyExercisesTreeViewModel();
```
nikke234, line 76: This doesn't have to be fixed now, but it's quite confusing that `new ExercisesTreeViewModel(new ArrayList<>(), new Options)` is actually different from `new EmptyExercisesTreeViewModel()`, because the first uses the empty text of the exercises tree, whereas the second one hides the tree and instead shows the separate empty text label. One option would be renaming `isEmptyTextVisible` to something else, but then `EmptyExercisesTreeViewModel` should also be renamed. Anyways, doesn't need to be fixed now

src/main/resources/resources.properties
```diff
@@ -144,6 +144,8 @@ ui.aboutDialog.GithubWebsite=A+ Courses plugin GitHub
 ui.aboutDialog.attributes=<html><div style='font-size:8px'>The plugin uses the following 3<sup>rd</sup> party libraries:<ul><li><a href="https://commons.apache.org/proper/commons-io/">Apache Commons IO</a></li><li><a href="https://plugins.jetbrains.com/plugin/1347-scala">IntelliJ Scala Plugin</a></li><li><a href="https://www.json.org/json-en.html">json.org</a></li><li><a href="https://www.scala-lang.org/api/current/">Scala Standard Library 2.13.4</a></li><li><a href="https://github.com/srikanth-lingala/zip4j">zip4j</a></li></ul></div></html>
 
 ### Assignments Tree
+ui.exercise.ExercisesView.allAssignmentsFiltered=All Assignments have been filtered out.
+ui.exercise.ExercisesView.loading=Loading..
```
nikke234, line 148: Nit: I would use three dots :smile:

nikke234 (approved):
Looks good :+1: Definitely not the easiest issue to solve :grimacing:

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/ProjectKey.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.intellij.openapi.project.Project;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A class that makes it easy to use projects as keys in maps. The equals and hashCode methods are
+ * based on the project base path. No reference to the project is stored, only a string containing
+ * the path.
+ */
```
OlliKiljunen, line 11: ❤️

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
```diff
@@ -121,6 +127,9 @@ public void readAuthenticationFromStorage(@Nullable PasswordStorage passwordStor
         .map(PasswordStorage::restorePassword)
         .map(factory::create)
         .ifPresent(this::setAuthentication);
+    if (!exercisesViewModel.get().isAuthenticated()) {
+      this.updateExercises();
+    }
```
OlliKiljunen, line 132: I'm not sure if it makes sense to ensure `updateExercises()` is called here. That should perhaps be done by the caller.

src/test/java/fi/aalto/cs/apluscourses/intellij/utils/ProjectKeyTest.java
```diff
@@ -0,0 +1,43 @@
+package fi.aalto.cs.apluscourses.intellij.utils;
+
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+
+import com.intellij.openapi.project.Project;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ProjectKeyTest {
```
OlliKiljunen, line 10: 😍:pray:

OlliKiljunen (approved):
Great work! And sorry for a late review.

Thanks for fighting this issue and congrats for beating it for now. 😄

src/main/java/fi/aalto/cs/apluscourses/presentation/MainViewModel.java
nikke234: BTW what's the reason that the call is needed here in the first place?

stellatsv: There is an earlier comment in this PR that sheds some light. I am not 100% sure about this but I think the view was not updated in a case where there was no token set and I needed to manually call updateExercises. I tested that everything works without it now so I can remove it.

Issue #147, opened 2020-04-29
The plugin checks that its own version is compatible with the course version when creating or opening a project


superseacat:

OlliKiljunen:
Also do the same check when trying to turn a project into A+ project (when first reading the course config file)

PR #588, opened 2021-03-15
Add a warning/error message if the course requires a later plugin version than the user uses

Taikelenn:
# Description of the PR
Related to #147.

This pull request adds an option for course administrators to recommend or enforce a minimum plugin version.

This can be used if a breaking change would be added to some course feature and older plugin version would not be functional anymore - in that case we can cleanly display an error message rather than waiting for some exception to be thrown. Another usage for this feature is that when a new useful functionality is added to the plugin, users would be warned that they might be missing out on this new feature.

**Improvements to the wording of error/warning messages are welcome!**

TODO:
- ~~unit tests~~
- ~~perhaps a "do not show again for this project"~~ _will be converted to notifications_
- ~~choose wording for the message boxes~~
- ~~add a warning label in the A+ project creation dialog~~
- ~~possibly change the message boxes to notifications~~
- ~~consider: how would it be possible that the user is trying to open a project using an outdated plugin?~~

Gallery:
example
example2
example3

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.


OlliKiljunen (commented):
First of all, this PR is fabulous in its rigour. :+1: Great work and many thanks! 

I'll take a closer look later but vefore that, I want to point out that there is an existing `Version` class available, and you can get the current version with something like `BuildInfo.INSTANCE.getVersion()`. Did you consider using that?

Taikelenn:
> 
> 
>There is an existing `Version` class available, and you can get the current version with something like `BuildInfo.INSTANCE.getVersion()`. Did you consider using that?

I did consider already existing "version classes", but they have two (well, now just one) disadvantages:
1. The plugin version is not necessarily the "course plugin version". The latter will be changed only if there is some significant change to the plugin that harms backwards compatibility. So, for example, plugin versions 2.0 and 2.1 should have the same course plugin version.
2. Initially the course plugin version contained more fields than just `major` and `minor`, so that's why a new class was created.

I guess 2. doesn't apply anymore but 1. still stands.

src/main/java/fi/aalto/cs/apluscourses/model/CoursePluginVersion.java
```diff
@@ -0,0 +1,51 @@
...
+   */
+  public Status checkVersion() {
+    if (MAJOR_VERSION < major) {
+      return Status.UPDATE_REQUIRED;
+    }
+    if (MAJOR_VERSION > major) {
+      return Status.VALID;
+    }
+    return MINOR_VERSION >= minor ? Status.VALID : Status.UPDATE_OPTIONAL;
+  }
```
nikke234, line 50: Correct me if I'm wrong: this seems to assume that newer plugin versions are compatible with older courses, even if there has been a change in the major version number? I think it would be safer to enable breaking changes both ways, i.e. the plugin is considered compatible with the course only if the major versions match.

Unrelated: it might be nice to have this method take another version parameter instead of using the static one, something like `Status isCompatibleWith(CoursePluginVersion other)`, then it would be more generic and unit testable

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -255,6 +262,11 @@ public String getName() {
     return Collections.unmodifiableMap(resourceUrls);
   }
 
+  @NotNull
+  public CoursePluginVersion getRequiredPluginVersion() {
+    return minimumPluginVersion;
+  }
```
nikke234, line 268: The name of the variable and the getter are pretty confusing. With this name, I would assume the version to correspond to a plugin version (e.g. 2.1), but it's actually a separate version for the course itself. Perhaps something like `courseVersion`?

src/main/java/fi/aalto/cs/apluscourses/intellij/activities/InitializationActivity.java
```diff
@@ -60,6 +62,16 @@ public void runActivity(@NotNull Project project) {
       notifier.notify(new NetworkErrorNotification(e), project);
       return;
     }
+
+    CoursePluginVersion minVersion = course.getRequiredPluginVersion();
+    CoursePluginVersion.Status versionCheckResult = minVersion.checkVersion();
```
nikke234, line 67: Nit: just noting that we have `var` now :wink:

src/main/java/fi/aalto/cs/apluscourses/model/CoursePluginVersion.java
```diff
@@ -0,0 +1,51 @@
+package fi.aalto.cs.apluscourses.model;
+
+import org.jetbrains.annotations.NotNull;
+import org.json.JSONObject;
+
+public class CoursePluginVersion {
```
nikke234, line 6: This is also potentially confusing in the same way as `getRequiredPluginVersion`. Perhaps this could be renamed to `CourseVersion` and `CURRENT_VERSION` to `DEFAULT_VERSION`?

src/main/java/fi/aalto/cs/apluscourses/model/CoursePluginVersion.java
nikke234: Sorry, the name `isCompatibleWith` is of course a bad idea, because it doesn't return a boolean. But taking the compared-to version as a parameter would still be a good idea

Taikelenn: Correct, new plugin versions are compatible with all previous minor releases of the same major AND all previous major releases, period.

What should the user do if the course requires an older major version? Do they even have an option for downgrading the A+ plugin? I haven't really thought about it but it sounds like a flaw on our side if we decide to drop support for old but currently running courses. (The finished courses will eventually be removed from the course selection anyway, like O1 2020 will stop showing up once 2021 is out).

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
Taikelenn: `courseVersion` is nice, but I still opt for including the word "plugin" in the JSON property name, since in the end it's truly a "course version" only in the context of our plugin

src/main/java/fi/aalto/cs/apluscourses/intellij/activities/InitializationActivity.java
Taikelenn: ~~Finally, Java becomes almost as good as C#~~

src/main/java/fi/aalto/cs/apluscourses/model/CoursePluginVersion.java
nikke234: Downgrading the plugin is possible, but it's not trivial: uninstall the plugin, download a ZIP of the correct version from the JetBrains plugin website, and install it using "install plugin from disk".

> I haven't really thought about it but it sounds like a flaw on our side if we decide to drop support for old but currently running courses

If the old course is still supported, then why would it have a different major version?

The way I see it, with the current approach we either support old courses until eternity, or the old courses break in some way but the user doesn't get a coherent incompatibility error notification.

I also haven't carefully considered this through so I might be missing something

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
nikke234: I could see someone creating the course config seeing `"minPluginVersion"` and going "oh right, I want my students to use at least version 2.2 of the plugin" :thinking:

src/main/java/fi/aalto/cs/apluscourses/model/CoursePluginVersion.java
Taikelenn: > The way I see it, with the current approach we either support old courses until eternity

Well... that's what I thought ;v or at least not until eternity, but as long as the course is present in the course selection list. To be honest, I was actually under the impression that old courses disappear from A+ entirely - maybe I was biased by this.

If someone manually enters the O1 2020 link (even though the course is gone from the course selection list), they're already assumed to be an "advanced user" - normal people don't even know what a course URL is!

But let's ignore what I wrote above and consider this: when we release an update with the IDE activities, we'll surely bump the major version from 1 to 2 (version 1 won't be able to run the IDE activities). But this does not mean that a version 2 plugin cannot open the O1 2020 course (which runs on version 1, without IDE activities). So in reality, the course would need to feature a "minimum compatible version" and a "maximum compatible version", with the latter manually updated when we release a truly breaking change (like a change to the course JSON structure). That's unreasonable both for us and the course administrators.

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
Taikelenn: Will change to `version`

src/main/java/fi/aalto/cs/apluscourses/model/CoursePluginVersion.java
nikke234: For future reference: after a duel in the daily, we decided to proceed with Pawel's plan.

OlliKiljunen: > when we release an update with the IDE activities, we'll surely bump the major version from 1 to 2 (version 1 won't be able to run the IDE activities)

I'm confused.

Are we talking about something else than the plugin's version (that is currently 2.0)?

I see the schema of course config file as an _interface_ that the plug-in _realizes_ and courses _use_. Thus, it makes sense to refer to the _version of the plugin_ in the course config file to solve the maintainability question.

Adding support to the assessable IDE activities isn't a breaking change: the plugin is still compatible with the older courses. Thus, the major version of the plugin shouldn't be increased because of that. Let's assume it's gonna be v2.2 that first introduces IDE-acts. Then a course that requires them refers to v2.2 in its course config file. Plugin versions 2.2, 2.3, 2.4, ... should all consider themselves compatible with that course. Then again, v2.1 and v3.0 should not, the former because it doesn't have the required functionality (IDE-acts, in this case), the latter because it contains some changes that breaks the contract the author of that course relied on.

Taikelenn: > Are we talking about something else than the plugin's version (that is currently 2.0)?

Yes - we're talking about the version specified in `CoursePluginVersion` class as MAJOR_VERSION and MINOR_VERSION constants. This is completely unrelated to the plugin build version (currently "v2.0").

> Adding support to the assessable IDE activities isn't a breaking change

Indeed, it does not break backwards compatibility. However the CoursePluginVersion should have its major version increased, because that's how the presence of the new required feature (IDE activities) will be distinguished. Otherwise the new O1 course (with IDE activities) won't have any way of actually requiring a new plugin version.

That's how this would work in practice:
1. Let's say we release v2.3, v2.4, v2.5 without IDE activities. These have MAJOR_VERSION = 1.
2. In v2.6 IDE activities are fully working and released. MAJOR_VERSION = 2.
3. The O1 2021 course now specifies that major version 2 is required. This means that all users running v2.6 get accepted, and all users running v2.5 or older get prompted to update.
4. If we ever release some new important feature for other courses, we'll bump MAJOR_VERSION to 3. O1 2021 will still work though because its required major version is lower than 3.

So the `CoursePluginVersion` is more like a "feature set version" rather than a "backwards compatibility version".

If we ever release an update that destroyes backwards compatibility with existing courses (courses with lower required major version no longer work) then **this ticket doesn't solve this problem**. Running courses should not be modified like that anyway, and we've (I and Nikolas) somewhat agreed on the fact that after a course is removed from the plugin's course selection list, it will most likely be never instantiated anymore. This would also mark the end of support for these courses.

OlliKiljunen: I appreciate your investment in the issue! I see you have thought this trough very well and have a strong understanding on the issue. By keeping challenging your ideas, I don't want to shoot them down, but make sure we are talking about the same things and are solving the same problems. I hope you have energy to answer my questions below.

> we're talking about the version specified in CoursePluginVersion class as MAJOR_VERSION and MINOR_VERSION constants. This is completely unrelated to the plugin build version (currently "v2.0").

Does this then leave "the plugin build version" out of any semantic meaning? Initially, it was our idea that the plugin version itself is what is used to resolve the compatibility with courses. Now, this sounds to me that we are introducing another version (version of the course-config language/fileformat?). Is this what you intended?

> the CoursePluginVersion should have its major version increased, because that's how the presence of the new required feature (IDE activities) will be distinguished. Otherwise the new O1 course (with IDE activities) won't have any way of actually requiring a new plugin version.

Why wouldn't we use minor version for additions that aren't breaking? Courses would then require plugin version _A.B_ with some major version _A_ and minor version _B_. Plugin versions _a.b_ where _a=A_ and _b>=B_ would then be compatible. Usually, that's what major/minor versions are for (such as in [SemVer standard](https://semver.org/)).

>That's how this would work in practice:
>
> 1. Let's say we release v2.3, v2.4, v2.5 without IDE activities. These have MAJOR_VERSION = 1.
> 2. In v2.6 IDE activities are fully working and released. MAJOR_VERSION = 2.
> 3. The O1 2021 course now specifies that major version 2 is required. This means that all users running v2.6 get accepted, and all users running v2.5 or older get prompted to update.
> 4. If we ever release some new important feature for other courses, we'll bump MAJOR_VERSION to 3. O1 2021 will still work though because its required major version is lower than 3.

Does MINOR_VERSION carry any meaning here?

> So the CoursePluginVersion is more like a "feature set version" rather than a "backwards compatibility version".

Could it be both if minor version is the "feature set version" and major version is the "backwards compatibility version"?

> This would also mark the end of support for these courses, when changes that break backwards compatibility can be released.

Course config files can keep circling around even if their courses have ended. A teacher can start a new instance of an old course by copying the old course config file. I'd see it ideal that they could just try loading the old file with the newest plugin version and, by seeing no warnings, make sure it is still supported by the plugin (meaning there's no breaking changes in-between). If they want to include new features they can just upgrade the required version to the newest one (or, if they are super pedantic, they can find the actual version that first introduced that feature and use that). What do you think?

Taikelenn: > Does this then leave "the plugin build version" out of any semantic meaning?

To be honest, it seems so. Perhaps I really haven't thought this thoroughly enough, but for example when v2.0 was released, there really weren't any breaking changes introduced despite major version increasing. Right now v1.8 and v2.0 are simultenously in active use in the same course. That's probably what influenced my decision of introducing a parallel version class.

> Why wouldn't we use minor version for additions that aren't breaking? Could it be both if minor version is the "feature set version" and major version is the "backwards compatibility version"?

Yes, this is possible. The thing is that when there's a mismatch between minor versions, a warning is shown, not an error. So the user might proceed anyway using a version that does not have IDE activities and be confused (perhaps exceptions due to unknown assignment types/unknown JSON fields or perhaps the assignments simply won't show up - either way, it's unspecified behaviour - something that we're trying to say goodbye to with this ticket).

Perhaps minor version should actually cause an error? But then it's no longer a truly "minor" version... the more I think about it, the more confused I become ;<

OlliKiljunen: No wonder we are confused here -- these are by no means easy things. 🤷‍♂️

>when v2.0 was released, there really weren't any breaking changes introduced despite major version increasing. Right now v1.8 and v2.0 are simultenously in active use in the same course. That's probably what influenced my decision of introducing a parallel version class.

You're absolutely correct here. We jumped to v2.0, I'd say, more for "marketing" reasons than technical ones. I'll add my review to this PR later today and try to consider from the perspective I got from your answers. So I can give feedback of the code. Let's consider then later (in a meeting?) which approach we want to take with the versioning. 👍

nikke234: I think 2.0 does count as breaking, because it introduces a *mandatory* `languages` property to the course configuration

src/main/java/fi/aalto/cs/apluscourses/intellij/notifications/CourseVersionError.java
```diff
@@ -0,0 +1,21 @@
...
+import com.intellij.notification.NotificationType;
+import fi.aalto.cs.apluscourses.intellij.services.PluginSettings;
+
+public class CourseVersionError extends Notification {
+  /**
+   * Notification to be shown when the plugin is outdated with respect to the current course.
+   *
+   * @param isError If true, the notification is an error. Otherwise, it is a warning.
+   */
+  public CourseVersionError(boolean isError) {
```
OlliKiljunen, line 15: Nit: perhaps polymorphism, that is, separate classes for warning and error, would make more sense than differentiating between them by a boolean variable. After all, they have very little in common.

src/main/java/fi/aalto/cs/apluscourses/model/CourseVersion.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public enum Status {
+    VALID,
+    UPDATE_OPTIONAL,
+    UPDATE_REQUIRED
+  }
+
+  // the major and minor version of the current plugin build
+  private static final int MAJOR_VERSION = 1;
+  private static final int MINOR_VERSION = 0;
```
OlliKiljunen, line 16: Remark: So, if I understood correctly, whenever the course config file-format is extended/changed, these constants should be increased, too. Not sure, if this is the most obvious place to maintain these version numbers.

OlliKiljunen (commented):
The code itself is very good, great job! 👍 Let's still discuss the overall approach of versioning, before approval.

src/main/java/fi/aalto/cs/apluscourses/intellij/notifications/CourseVersionTooNewError.java
```diff
@@ -0,0 +1,17 @@
+package fi.aalto.cs.apluscourses.intellij.notifications;
+
+import static fi.aalto.cs.apluscourses.utils.PluginResourceBundle.getText;
+
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationType;
+import fi.aalto.cs.apluscourses.intellij.services.PluginSettings;
+
+public class CourseVersionTooNewError extends Notification {
```
nikke234, line 9: I was thinking of a better name and came up with "indated" as the opposite of outdated, but apparently that's not an english word :frowning_face:

src/test/java/fi/aalto/cs/apluscourses/model/CourseTest.java
```diff
@@ -163,12 +171,13 @@ public void testGetModuleWithMissingModule() throws NoSuchComponentException {
...
   private static String replInitialCommands = "\"repl\": {\"initialCommands\": {\"GoodStuff\": ["
       + "\"import o1._\",\"import o1.goodstuff._\"]}}";
+  private static String courseVersion = "\"version\": \"5.8\"";
 
   @Test
   public void testFromConfigurationFile() throws MalformedCourseConfigurationException {
     StringReader stringReader = new StringReader("{" + idJson + "," + nameJson + "," + urlJson
         + "," + languagesJson + "," + modulesJson + "," + exerciseModulesJson + "," + resourcesJson
-        + "," + autoInstallJson + "," + replInitialCommands + "}");
+        + "," + autoInstallJson + "," + replInitialCommands + "," + courseVersion + "}");
```
nikke234, line 180: This is perhaps more of an issue with how the tests are implemented, but now all of the malformed JSON tests that are later in this file don't actually test what they claim to test; all of them would fail in any case because the missing version causes an exception.

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -255,6 +264,11 @@ public String getName() {
     return Collections.unmodifiableMap(resourceUrls);
   }
 
+  @NotNull
+  public Version getCourseRequiredVersion() {
```
nikke234, line 268: Not hugely important, but just `getCourseVersion` would be nicer :smile:

src/main/resources/resources.properties
```diff
@@ -30,6 +30,10 @@ notification.FeedbackAvailableNotification.content=Feedback for {0} [{1}] is now
 notification.OpenSubmissionNotificationAction.content=Click here to open the feedback.
 notification.CourseFileError.title=Could not access course settings file
 notification.CourseFileError.content=The local file for course settings could not be accessed. Error message: {0}.
+notification.CourseVersionError.title=Plugin incompatible with course
+notification.CourseVersionOutdatedError.content=This course project cannot be loaded because your A+ Courses plugin version is outdated. Please update the A+ Courses plugin to open this course.
+notification.CourseVersionTooNewError.content=This course is no longer supported by your A+ Courses plugin version. Please contact the course staff for assistance.
+notification.CourseVersionOutdatedWarning.content=This course recommends using a newer version of the A+ Courses plugin. Update so as not to miss out on any new useful features!
```
nikke234, line 36: Perhaps we could quickly consult on these in the weekly. I'm not too good at making these up, but "update so as not to miss out on any new useful features!" is very markety :smile:

src/main/java/fi/aalto/cs/apluscourses/utils/BuildInfo.java
```diff
@@ -20,6 +20,9 @@
   @NotNull
   public final Version version;
 
+  @NotNull
+  public final Version courseVersion;
```
nikke234, line 24: Once again not hugely important, but renaming the other one to `pluginVersion` would help disambiguate the two

nikke234 (approved):
Looks good! Some minor things here and there but approved as-is. I appreciate that you stuck through with the issue despite the numerous confusions and delays :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/activities/InitializationActivity.java
```diff
@@ -75,6 +80,19 @@ public void runActivity(@NotNull Project project) {
...
+
+    if (versionComparison == Version.ComparisonStatus.MAJOR_TOO_OLD
+        || versionComparison == Version.ComparisonStatus.MAJOR_TOO_NEW) {
+      notifier.notify(
+          versionComparison == Version.ComparisonStatus.MAJOR_TOO_OLD
+          ? new CourseVersionOutdatedError() : new CourseVersionTooNewError(), project);
+      return;
+    } else if (versionComparison == Version.ComparisonStatus.MINOR_TOO_OLD) {
+      notifier.notify(new CourseVersionOutdatedWarning(), project);
+    }
```
OlliKiljunen, line 94: This logic is duplicated. My suggestion: create a factory method that takes a comparison status and returns an appropriate notification.

Then something like:

```java
if (!comparisonStatus.isFullyCompliant())
  notifier.notify(notificationFactory.create(comparisonStatus));

if (!comparisonStatus.isBreaking())
  return;
```

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -492,6 +506,23 @@ private static String getCourseAPlusUrl(@NotNull JSONObject jsonObject, @NotNull
     return replInitialCommands;
   }
 
+  @NotNull
+  private static Version getCourseVersion(@NotNull JSONObject jsonObject,
+                                                @NotNull String source)
+      throws MalformedCourseConfigurationException {
+    String versionJson = jsonObject.optString("version", null);
+    if (versionJson == null) {
+      return BuildInfo.INSTANCE.courseVersion;
```
OlliKiljunen, line 515: Does that effectively mean that if version is missing from course config file, it is expected to be fully compliant by default.

OlliKiljunen (commented):
I'm still a bit confused when it should be an error and when a warning. I think this solution of yours (which I think is fine in its own logic) is contrary to what was discussed the last time.

I remember our consensus was that error is shown when there is an actual error: when, for example, a property is missing from a file that shoukd be there by the current version of specs. On the other hand, all anomalies listed in ComparisonStatus enum (too new/old major, too old minor) should show a warning (unless error is already shown because of an actual error as described above).

So can you once again remind me what did we decide on the issue together, and if you, in the end, chose not to follow that design (which could have been for some very valid reason), what was the reasoning behind that.

nikke234:
> I'm still a bit confused when it should be an error and when a warning. I think this solution of yours (which I think is fine in its own logic) is contrary to what was discussed the last time.
> 
> I remember our consensus was that error is shown when there is an actual error: when, for example, a property is missing from a file that shoukd be there by the current version of specs. On the other hand, all anomalies listed in ComparisonStatus enum (too new/old major, too old minor) should show a warning (unless error is already shown because of an actual error as described above).
> 
> So can you once again remind me what did we decide on the issue together, and if you, in the end, chose not to follow that design (which could have been for some very valid reason), what was the reasoning behind that.

I think this implements it correctly as far I remember the discussion and decision:

- Major mismatch = ERROR (which often corresponds to a missing property, but can in principle be anything, maybe some existing property gets new values in incompatible ways etc.)
- Minor outdated = WARNING

OlliKiljunen:
Okay, it is possible that I remember our discussion wrong. If that's the case, this is fine (although I wish you'd still address my other comment about code duplication; but approvable also as-is).

src/main/java/fi/aalto/cs/apluscourses/intellij/notifications/CourseVersionTooNewError.java
Taikelenn: `CourseVersionBackwardsIncompatibleError` specific but too verbose for my liking

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
Taikelenn: Yes, if there's no information about version in the course configuration file, then no version check should be performed. This is achieved by making the course version always the same as the user's version (= always fully compliant, as you said)

OlliKiljunen (approved):

PR #607, opened 2021-04-10
Improved exercise updates

nikke234:
# Description of the PR

This PR moves exercises updates and authentication out of `MainViewModel`. Along with the code improvements it also fixes two user-facing issues:

- Refreshing assignments no longer causes the UI to freeze (the update no longer happens in the UI thread)
- Module and assignment refreshes actually do different things.

The exercise update still has the previous approach of replacing the entire `List<ExerciseGroup>` with a new list, instead of reusing the existing objects. It's possible to instead just update the existing model instances (e.g. update the points of exercises, add any missing submissions, etc.), although I'm not sure how that would interact with filtering.

TODO: Unit tests

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
        <li>- [ ] new <b>unit</b> tests created</li>
        <li>- [x] all <b>unit</b> tests pass</li>
      </ul>
    </td>
    <td>
      <ul>
        <li>- [ ] new <b>integration</b> tests created</li>
        <li>- [x] all <b>integration</b> tests pass</li>
      </ul>
    </td>
    <td>
      <ul>
        <li>- [ ] new <b>e2e</b> tests created</li>
        <li>- [x] all <b>e2e</b> tests pass</li>
      </ul>
    </td>
    <td>
      <ul>
        <li>- [ ] <b>manual</b> testing went well</li>
      </ul>
    </td>
  </tr>
</table>  
  
# Have you updated the [TESTING.md](https://github.com/Aalto-LeTech/intellij-plugin/blob/master/TESTING.md) or other relevant documentation on _your_ branch?

- [ ] Yes.
- [ ] Not yet. I will do it next.
- [x] Not relevant.

PR #609, opened 2021-04-13
Loading bar

### LINKED ISSUE #519: Loading bar
It would be nice to have a loading bar for things like fetching modules and exercises, and downloading modules (#466)

Here's mockups

image
This text could be replaced with a loading bar, since it isn't relevant after the A+ token is already set

image
A progress bar could also go down here and be used for everything that needs loading

jaakkonarhi:
# Description of the PR

+ Loading bar for module and assignment refreshes.
+ The bar can either set as indeterminate or determinate.
+ If there are two separate refreshes at the same time, the progresses get combined and the label says Loading....
+ Should be merged after #607 
+ Sorry I couldn't make it pink
+ Closes #519 

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.


OlliKiljunen (commented):
Thank you for this PR: it's very nice code and written in a clearly manner. 👍

I understand that there is a single instance of `ProgressViewModel` (per project) used for various purposes. It is a bit unclear what happens if multiple progresses are going on simultaneously; apparently the text is set just "Loading" and the max value is the sum of all the progresses. What if one of these progresses ends in an error? Another thing I find confusing is `isIndeterminate` status of the progress. When is it client code's responsibility to set that value and when does the ProgressVM do it itself?

Have you considered an approach where each of the "progresses" had a progressVM (or just "progress") instance of their own and either just one of them (by some logic) or some "aggregate-progress" is shown to the user? I would preferably see different progresses handled separately to prevent them from confusing each other.

I neither approve nor request changes because what I want is to hear you thoughts of what I wrote. The code itself is fine, if we want to go with this approach. I left one comment you can consider, but it's a minor issue.

stellatsv (changes requested):
The code is very good! The case that the project is not yet turned into A+ should also be handled, but this looks very nice so far.

OlliKiljunen (approved):
Really nice work! Thanks for addressing issues of earlier reviews. 👍 

In this implementation, `ProgressViewModel` only knows to update itself if changes to the `Progress` objects happen through it (by someone calling `ProgressViewModel::increment`). If `Progress`es are changes otherwise, `ProgressViewModel` doesn't know it should update its state. This is just a remark (and something we can think of later, if it becomes topical).

Anyway, a challenging task well done! Approved.

stellatsv (approved):
The functionality is all good now. Approved!

PR #612, opened 2021-04-14
Framework ide act

stellatsv:
# Description of the PR

Closes #563.

Before it is merged, I will need to revert a couple of things that I altered in order to quickly load a mock `Tutorial` object. I have kept them still so the reviewers can check the functionality.

The `Tutorials` are included in the `Assignments` tree. When a `Tutorial` is selected an icon is added to the toolbar which when clicked initiates the `TutorialAction`. If another type of `Exercise/ExerciseGroup` is selected the icon is not shown. This way, the framework functionality can be merged as is and the new `TutorialAction` will not be visible to the users yet.

We will need some new icons for the `TutorialAction` and the `Tutorial` entries in the tree, it can be done as part of #604. 

So, a `Tutorial` is comprised of different `Tasks`. Each `Task` needs a different `ActivitiesListener`. The listener is created and registered for each `Task` when needed and unregistered once the `Task` has been completed. Also, after each `Task` is done an `Event` signal is sent to `TutorialViewModel` which initiates the next `Task` or if there are no more `Tasks`, completes the `Tutorial`. The class `TutorialViewModel` sort of oversees the succession of `Tasks` in the `Tutorial`. 

`StartTutorialDialog` asks the user to confirm that they wish to begin the selected `Tutorial`. Then, it signals `TutorialViewModel` to begin showing the next `Task` in the sequence (the first one). For each `Task`, there is an instruction window shown with the help of `TaskViewModel` and `TaskView`. 

So far, I have been using a listener for files being opened. There is the case where the activity of the `Task` has already been performed, i.e. the file is already open. This could perhaps happen again if, for example, the user has already renamed a variable/class/method with the desired name. In the current implementation, this is acknowledged in the `TutorialViewModel` using an `Event` signal. The instruction window for this `Task` is not shown. Instead, there is the ability to inform the user that this `Task` that was supposed to be next is already achieved and that they will be forwarded to the next `Task` (and proceed with the `Task` sequence).

To make the process a bit easier to navigate there is also this **informal** diagram:

frameworkPR

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
        <li>- [x] new <b>unit</b> tests created</li>
        <li>- [x] all <b>unit</b> tests pass</li>
      </ul>
    </td>
    <td>
      <ul>
        <li>- [ ] new <b>integration</b> tests created</li>
[...]

PR #613, opened 2021-04-19
Proof of concept of the UI overlay

Taikelenn:
## Note: this is a proof of concept, which should preferably wait for #612 to be merged first, and then adapt itself to the changes therein so that everything works smoothly.

# Description of the PR
The UI overlay features three new concepts:
+ OverlayPane - a Swing component (LayeredPane) that covers the entire window and dims it, while potentially excluding a certain component from the dimming in order to bring attention to it.
**This class is partially static** - that is because there's potentially only one pane installed at once. It was the simplest to implement for a PoC. Feel free to propose a more suitable pattern that would be better suited for a release version.
+ BalloonPopup - another Swing component (JPanel) that attaches itself to an arbitrary component and is showing text (instructions) about what to do in the IDE activity (see screenshots)
+ ComponentLocator - a static class that traverses all IntelliJ's components and locates the one we want. This can (and should) be extended to support the component selection methods we deem reasonable and useful. For now the components are located by a substring in their class name, which is robust enough for components that actually have their own classes.

Popups automatically resize and reposition themselves when needed. Actions have been added to the A+ menu to facilitate manual testing, they will obviously be removed in actual release.

To do (somewhat in order of importance):
- Popups always attach to the left or right to the component (horizontal positioning) - add support for vertical positioning as well
- Adjustable maximum popup width
- Intercept mouse events (clicks) that occur inside the dimmed area

# Video showcase
https://user-images.githubusercontent.com/73069541/115216449-2f96f080-a10d-11eb-820b-eb4cf1259111.mp4


src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/OverlayPane.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+    activeOverlay.revalidatePane();
+
+    activeOverlay = null;
+  }
+
+  /**
+   * Marks a component not to be dimmed.
+   */
+  public static void showComponent(Component c) {
```
OlliKiljunen, line 150: Nit: The overlay could be accessed through an object instance instead of static methods.

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/OverlayPane.java
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  public static boolean isOverlayInstalled() {
+    return activeOverlay != null;
+  }
+
+  /**
+   * Installs the overlay.
+   */
+  public static void installOverlay() {
```
OlliKiljunen, line 120: Can only be called from EDT, i assume? That's fine but it's best to be explicitly stated.

OlliKiljunen (approved):
Superb! 👍 This is really excellent implementation. I left a few nits but nothing that should block this from being merged.

stellatsv (approved):
Definitely approve-worthy, the UI looks great!

PR #637, opened 2021-05-16
Course ended banner

### LINKED ISSUE #610: Show banner if course has ended
The api has the ending time for a course, so a banner could be shown according to that

jaakkonarhi:
# Description of the PR

+ Banner for when the course has ended
+ Closes #610 

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.


src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseProject.java
```diff
@@ -46,7 +48,8 @@
   @NotNull
   public final Event exercisesUpdated;
 
-  private volatile User user;
+  @NotNull
+  private final ObservableProperty<User> user = new ObservableReadWriteProperty<>(null);
```
OlliKiljunen, line 52: We have this convention to make these public to avoid having getters. (`getUser().set(...)` is confusing)

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseProject.java
```diff
@@ -107,7 +111,12 @@ public void removePasswordFromStorage(@NotNull PasswordStorage.Factory passwordS
 
   @NotNull
   public String getUserName() {
-    return user == null ? "" : user.getUserName();
+    return user.get() == null ? "" : Objects.requireNonNull(user.get()).getUserName();
```
OlliKiljunen, line 114: Either
```
var myUser = user.get(); 
return myUser == null ? "" : myUser.getUserName();
```
or
```
return Optional.ofNullable(user.get()).map(User::getUserName).orElse("")
```
The current code has possibility that `user.set(...)` is called between the conditional and the latter expression.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseProject.java
```diff
@@ -107,7 +111,12 @@ public void removePasswordFromStorage(@NotNull PasswordStorage.Factory passwordS
...
+  }
+
+  @NotNull
+  public ObservableProperty<User> getUser() {
+    return user;
   }
```
OlliKiljunen: This shouldn't be needed.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseProject.java
```diff
@@ -126,25 +135,21 @@ public void setExerciseGroups(@NotNull List<ExerciseGroup> exerciseGroups) {
...
-      this.user = null;
+      this.user.set(null);
     } else {
-      this.user = new User(authentication, course.getExerciseDataSource());
-      var bannerViewModel = PluginSettings.getInstance().getMainViewModel(project)
-              .bannerViewModel.get();
-      if (bannerViewModel != null) {
-        bannerViewModel.update();
-      }
+      this.user.set(new User(authentication, course.getExerciseDataSource()));
```
OlliKiljunen: Use `var oldUser = user.getAndSet(newUser)` for atomicity.

src/test/java/fi/aalto/cs/apluscourses/presentation/CourseEndedBannerViewModelTest.java
```diff
@@ -7,14 +7,19 @@
...
 import fi.aalto.cs.apluscourses.model.Authentication;
 import fi.aalto.cs.apluscourses.model.ModelExtensions;
+import fi.aalto.cs.apluscourses.model.User;
+import fi.aalto.cs.apluscourses.utils.observable.ObservableReadWriteProperty;
 import org.junit.Test;
 
 public class CourseEndedBannerViewModelTest {
   @Test
   public void testBanner() {
     var courseProject = mock(CourseProject.class);
```
OlliKiljunen, line 17: Use actual object instance instead of a mock here. Then call `courseProject.user.set(user)`. You can mock `Project` and `Authentication`. I think they can be just "empty" mock objects (no need to use `when` here anywhere).

OlliKiljunen (commented):
I looked only the last commit. Here are my comments.

src/main/java/fi/aalto/cs/apluscourses/ui/BannerView.java
```diff
@@ -0,0 +1,58 @@
...
+    banner.setBorder(BorderFactory.createCompoundBorder(
+            BorderFactory.createMatteBorder(0, 0, 1, 0, JBColor.border()),
+            BorderFactory.createEmptyBorder(0, 10, 0, 10))
+    );
+
+    this.container.add(BorderLayout.CENTER, bottomComponent);
+    this.container.add(BorderLayout.NORTH, banner);
+
+    this.bindable = new Bindable<>(banner, (panel, text) -> {
+      panel.setVisible(!text.equals(""));
```
stellatsv, line 39: It is safer to use `("").equals(text)` here.

stellatsv (approved):
Very nice! I have left only one small comment.

src/main/java/fi/aalto/cs/apluscourses/ui/BannerView.java
jaakkonarhi: Why is that safer?

stellatsv: If `text` happened to be null a `NullPointerException` would be thrown. With the `equals` method being called from `("")` which is not null the exception can be avoided.

jaakkonarhi: I thought it was that, but some website said that if either string is null it returns false, but apparently that's not true

src/main/java/fi/aalto/cs/apluscourses/presentation/CourseEndedBannerViewModel.java
```diff
@@ -0,0 +1,43 @@
...
+      return;
+    }
+    try {
+      var endingTime = course.getExerciseDataSource().getEndingTime(course, authentication);
+      if (endingTime.compareTo(ZonedDateTime.now()) < 0) {
+        this.text.set(getText("ui.BannerView.courseEnded"));
+      }
+    } catch (IOException e) {
+      logger.error("Failed to fetch ending time", e);
+    }
```
OlliKiljunen, line 41: This logic could be in the model: ``course::isEnded`` or something. Ending time could also be stored in the course object.

src/main/java/fi/aalto/cs/apluscourses/BannerViewModel.java
```diff
@@ -0,0 +1,10 @@
+package fi.aalto.cs.apluscourses;
+
+import fi.aalto.cs.apluscourses.utils.observable.ObservableProperty;
+import fi.aalto.cs.apluscourses.utils.observable.ObservableReadWriteProperty;
+
+public abstract class BannerViewModel {
+  public final ObservableProperty<String> text = new ObservableReadWriteProperty<>("");
+
+  public abstract void update();
+}
```
OlliKiljunen, line 10: Do we need this kind of class? Can't the BannerView to be bound to any text (`ObservableProperty<String>`)?

OlliKiljunen (commented):
I left a couple of comments. This doesn't necessarily require changes but I'd like to hear your thoughts of the issues I raised before approving.

src/main/java/fi/aalto/cs/apluscourses/presentation/CourseEndedBannerViewModel.java
jaakkonarhi: I didn't wanna put it in Course because everything else there comes from the course config

OlliKiljunen: > I didn't wanna put it in Course because everything else there comes from the course config

Seems like another symptom of some design issues we discussed. I still think this logic should go to the model but let's have it here for now if we can't come up with a class it would belong to.

OlliKiljunen (approved):

OlliKiljunen:
One comment more: Have you considered making `text` and `color` `ObservableReadOnlyProperty`s and for them, `declareDependentOn` `user`? Ideally, there shouldn't be `update` methods but data should be computed when it's requested.

That would, however require you to implement some kind of caching for `ObservableReadOnlyProperty`. If you are interested in trying that, you can make a subclass of it that holds the value in `Optional<T>` field. Then override `get` to call the `getter` only if the cached value ain't there (and update it). Then override `onValueeChanged` to invalidate the cached value before calling `super.onValueChanged(...)`.

Taikelenn (changes requested):
Please make tests

Taikelenn (approved):

src/main/java/fi/aalto/cs/apluscourses/utils/observable/ObservableCachedReadOnlyProperty.java
```diff
@@ -0,0 +1,42 @@
...
+  protected synchronized void onValueChanged(@Nullable Object source) {
+    isInitialized = false;
+    cached = null;
+    super.onValueChanged(source);
+  }
+
+  @Nullable
+  @Override
+  public T get() {
+    synchronized (this) {
```
OlliKiljunen, line 34: Nit: This could as well be `public synchronized T get()`.

src/test/java/fi/aalto/cs/apluscourses/utils/observable/ObservableCachedReadOnlyPropertyTest.java
```diff
@@ -0,0 +1,23 @@
...
+  @Test
+  public void testObservableCachedReadOnlyProperty() {
+    var supplier = (Supplier<Object>) mock(Supplier.class);
+    var observable = new ObservableCachedReadOnlyProperty<>(supplier);
+    observable.get();
+    observable.get();
+    verify(supplier, times(1)).get();
+    observable.get();
+    verifyNoMoreInteractions(supplier);
+  }
```
OlliKiljunen, line 22: Very good! Nit: You could also assert that `get()` returns (on the both calls) the value that `supplier` gives.

OlliKiljunen (approved):
Nice! Thanks for taking previous suggestions into account. 👍

PR #640, opened 2021-05-17
Add synchronization to ProgressViewModel

jaakkonarhi:
# Description of the PR

+ Should fix an issue if I did it right

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/presentation/ProgressViewModel.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/presentation/ProgressViewModel.java
@@ -3,12 +3,11 @@ package fi.aalto.cs.apluscourses.presentation;
 import static fi.aalto.cs.apluscourses.utils.PluginResourceBundle.getText;
 
 import fi.aalto.cs.apluscourses.model.Progress;
+import fi.aalto.cs.apluscourses.utils.CollectionUtil;
 import fi.aalto.cs.apluscourses.utils.observable.ObservableProperty;
 import fi.aalto.cs.apluscourses.utils.observable.ObservableReadWriteProperty;
 import java.util.ArrayDeque;
-import java.util.ArrayList;
 import java.util.Deque;
-import java.util.List;
 import java.util.Optional;
 
 public class ProgressViewModel {
@@ -44,21 +43,14 @@ public class ProgressViewModel {
    */
   public void stopAll() {
     synchronized (lock) {
-      progresses.clear();
+      progresses.forEach(Progress::finish);
     }
     this.updateValues();
   }
 
   private void updateValues() {
     synchronized (lock) {
-      List<Progress> toBeRemoved = new ArrayList<>();
-      for (var progress : progresses) {
-        if (progress.getValue() >= progress.getMaxValue()) {
-          progress.updated.removeCallback(this);
-          toBeRemoved.add(progress);
-        }
-      }
-      toBeRemoved.forEach(progresses::remove);
+      CollectionUtil.removeIf(progresses, Progress::isFinished, this::unregisterProgress);
       if (progresses.isEmpty()) {
         this.maxValue.set(0);
         this.value.set(0);
@@ -92,5 +84,9 @@ public class ProgressViewModel {
 
   private boolean isIndeterminate() {
     return Optional.ofNullable(this.indeterminate.get()).orElse(false);
+  }
+
+  private void unregisterProgress(Progress progress) {
+    progress.updated.removeCallback(this);
   }
 }
--- a/src/main/java/fi/aalto/cs/apluscourses/presentation/ProgressViewModel.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/presentation/ProgressViewModel.java
@@ -6,7 +6,9 @@ import fi.aalto.cs.apluscourses.model.Progress;
 import fi.aalto.cs.apluscourses.utils.observable.ObservableProperty;
 import fi.aalto.cs.apluscourses.utils.observable.ObservableReadWriteProperty;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Deque;
+import java.util.List;
 import java.util.Optional;
 
 public class ProgressViewModel {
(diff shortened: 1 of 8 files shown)


src/main/java/fi/aalto/cs/apluscourses/presentation/ProgressViewModel.java
```diff
@@ -28,43 +29,64 @@ public ProgressViewModel() {
...
 
+  /**
+   * Increments a Progress and updates the value of the visible progress.
+   */
   public void increment(Progress progress) {
-    progress.increment();
+    synchronized (lock) {
+      progress.increment();
+    }
     this.updateValue();
```
OlliKiljunen, line 90: `updateValue` is called here and it assumes that `progresses` is non-empty (calls `getLast()`). However, how do we now here `progresses` is non-empty? This problem doesn't get away by adding synchronization to this class.

A possible "fix" is change this to call `updateValues` which handles the empty deque situation. `updateValue` can then be removed.

I, however, don't like too much that `Progress` class can be only accessed through this view model. May I suggest the following:
- remove `ProgressViewModel::increment`.
- add `updated` event to `Progress` class. Make the viewmodel listen to the event on `start`. Make `Progress::increment` call `updated.trigger()`.
- remove `stop` method from the `ProgressViewModel`. Instead add it `stop` (or perhaps `finish`) to `Progress` class. That method sets progress finished (`value = maxValue`) and triggers `updated` event.
- In `ProgressViewModel::updateValues`, when you get the last progress, you check if it's finished, and if true, remove it from the deque and unregister the event handler.
- Wherever you use progresses, use model (`Progress`) objects. Call `Progress::increment` and `Progress::finish` methods and let the viewmodel take care of updating itself via eventing we implemented. Use viewmodel only to start progresses (this could actually happen via some interface but let's not go that far for now) and presenting (i.e. pass it to the view).

How does this sound to you?

OlliKiljunen (commented):
Syncing here is basically ok: all accesses to `progresses` happen inside `synchronized(lock)`.

The error Pawel got wasn't however, that much related to data races over `progresses` but `updateValue` not making sure the deque is not empty. See my comment and discuss. :slightly_smiling_face:

src/main/java/fi/aalto/cs/apluscourses/presentation/ProgressViewModel.java
jaakkonarhi: Should ProgressViewModel::updateValues check if any progress is finished in that case

jaakkonarhi: Because else some finished progresses might stay in the deque

OlliKiljunen: I don't think it matters. They get eventually popped once all the progresses on top of them have been popped too.

src/main/java/fi/aalto/cs/apluscourses/presentation/ProgressViewModel.java
```diff
@@ -28,46 +31,49 @@ public ProgressViewModel() {
...
-      this.indeterminate.set(progress.getIndeterminate());
+    synchronized (lock) {
+      List<Progress> toBeRemoved = new ArrayList<>();
+      for (var progress : progresses) {
+        if (progress.getValue() >= progress.getMaxValue()) {
+          progress.updated.removeCallback(this);
+          toBeRemoved.add(progress);
+        }
+      }
+      toBeRemoved.forEach(progresses::remove);
```
OlliKiljunen, line 61: If it feels terrifying to add this much logic to a method, you can make a helper method to CollectionUtil:
```java
public static <T> void removeIf(Collection<T> collection,
                                Predicate<T> predicate,
                                Consumer<T> callbackForRemoved) {
  //...
}
```
which you can unit test with e.g. strings.

Then you can use that here like
```java
CollectionUtil.removeIf(progresses, Progress::isFinished, this::unregisterProgress);
```
where `unregisterProgress` is a simple private method that calls `progress.updated.removeCallback(this)` (or use lambda).

OlliKiljunen (approved):
Good work!

I approve although there is one concern related to `stopAll` method. You can say better if it needs to be fixed.

src/main/java/fi/aalto/cs/apluscourses/presentation/ProgressViewModel.java
```diff
@@ -28,46 +31,49 @@ public ProgressViewModel() {
...
-
+  /**
+   * Stops all Progresses.
+   */
   public void stopAll() {
-    progresses.clear();
+    synchronized (lock) {
+      progresses.clear();
+    }
     this.updateValues();
```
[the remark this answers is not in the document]
jaakkonarhi: In that case updateValues would get called multiple times, so I don't think that's better

src/test/java/fi/aalto/cs/apluscourses/utils/CollectionUtilTest.java
```diff
@@ -24,4 +31,30 @@ public void testIndexOf() {
...
+    collection.add("Audi");
+    collection.add("BMW");
+    collection.add("Chevrolet");
+    collection.add("Daimler");
+    collection.add("Alfa Romeo");
+    collection.add("Bentley");
+    collection.add("Chrysler");
+    collection.add("Dodge");
+
+    CollectionUtil.removeIf(collection, s -> s.startsWith("C"));
```
OlliKiljunen, line 50: This should check that the collection returned contains "Chrysler" and "Chevrolet". Also, the return type is currently `List<T>` which suggests that the ordering matters. Consider changing it to `Collection<T>` (even if the returned object was `ArrayList<T>`).

stellatsv (approved):
Very thorough work!

PR #651, opened 2021-06-03
Transparent balloon popups for IDE activities

### LINKED ISSUE #638: IDE act: add more dimmable components
Right now only the project tree is dimmable, which is not very useful. Planned changes:
- [x] ~~Add more dimmable components (~~editor~~, A+ tool window components etc.)~~
- [ ] Think about positioning of popups attached to large components (for example with the text editor, there might be not enough space for popups) - perhaps variable width of popups
- [ ] Some unit testing of popup placement logic (optional)
- [ ] Juha suggested highlighting sections of code - possibly block scrolling as well

Taikelenn:
# Description of the PR
Closes #638.

This PR adds many various useful changes:
- makes the popups not opaque, which allows them to be placed on components if there is no available space in the IDE window. It is also a necessary step for future improvements such as _shadows_ or rounded corners
- increases checkstyle line length limit to 120 characters (up from previous 100)
- adds potential support for highlighting of more generic rectangular shapes (rounded rectangles, ellipses)
- brings back highlighting of the line numbers in the editor

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.


src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/PopupTransparencyHandler.java
```diff
@@ -0,0 +1,65 @@
...
+      resetAnimationProgress();
+
+      shouldRedraw = false;
+    }
+
+    if (currentCoefficient > MAX_TRANSPARENCY) {
+      currentCoefficient = MAX_TRANSPARENCY;
+      resetAnimationProgress();
+
+      shouldRedraw = false;
```
OlliKiljunen, line 57: Would omitting redraw cause the popup to never achieve min/max levels? Anyway, if this animation redraws the component, what, ~10 times maybe, does it matter if there's one time more? Or did I get something wrong?

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/PopupTransparencyHandler.java
```diff
@@ -0,0 +1,65 @@
...
+public class PopupTransparencyHandler {
+
+  private final TransparentComponent component;
+
+  private long lastTransitionUpdate;
+
+  private static final float MIN_TRANSPARENCY = 0.55f;
+  private static final float MAX_TRANSPARENCY = 1.0f;
+  private static final float TRANSITION_STEP = 0.0015f;
+
```
OlliKiljunen, line 12: (1.0 - 0.55) / 0.0015 = 300. So does that mean the animation takes 0.3 seconds? Just trying to understand :)

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/PopupTransparencyHandler.java
```diff
@@ -0,0 +1,65 @@
+package fi.aalto.cs.apluscourses.ui.ideactivities;
+
+public class PopupTransparencyHandler {
```
OlliKiljunen, line 3: "Popup" in the name is perhaps unnecessary because this isn't popup specific (except for the parameter of `update(boolean)` but that could perhaps, in this class' context, be called something like `increaseOpaquity`).

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/OverlayPane.java
```diff
@@ -42,11 +42,15 @@ private Area getDimmedArea() {
...
       if (c.getComponent().isShowing()) {
         for (var rectangle : c.getArea()) {
-          var windowRect = SwingUtilities.convertRectangle(c.getComponent(), rectangle, this);
-          overlayArea.subtract(new Area(windowRect));
+          var rectangleArea = new Area(rectangle);
+          rectangleArea.transform(translation);
+
+          overlayArea.subtract(rectangleArea);
         }
       }
```
OlliKiljunen, line 55: This is just optimizing, right? No functional changes here? But good optimizing! :+1:

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/BalloonPopup.java
```diff
@@ -52,56 +79,144 @@ public boolean isVisible() {
...
+  @Override
+  public float getTransparencyCoefficient() {
+    return transparencyCoefficient;
+  }
+
+  /**
+   * Sets the popup transparency level.
+   *
+   * @param coefficient The transparency coefficient - a value between 0.0f and 1.0f.
+   *                    0.0f means completely transparent, 1.0f means completely opaque.
```
OlliKiljunen, line 108: Okay here's the info I was looking for in an earlier comment. :+1:

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/BalloonPopup.java
```diff
@@ -52,56 +79,144 @@ public boolean isVisible() {
...
+          popupY = componentWindowPos.y + anchorComponent.getHeight() + 5;
+        } else {
+          popupY = componentWindowPos.y - popupHeight - 5;
+        }
+
+        popupX = componentWindowPos.x + (anchorComponent.getWidth() - popupWidth) / 2;
       }
 
-      popupX = componentWindowPos.x + (anchorComponent.getWidth() - popupWidth) / 2;
+      setTransparencyCoefficient(1.0f);
```
OlliKiljunen, line 184: Hmm... why?

OlliKiljunen (commented):
This looks like very good code. But I'm afraid I don't fully understand it. There are animations but they are only updated when something happens (mouse moves in/out or when swing repaints)... So, I think could happen that animation keep hanging for a long time. Which may be what you/we want but let's make this clear. (Because I would have expected some timer to handle the animation.)

I neither approve nor request changes until I have heard your comments. It well may be that this is totally fine as-is, but first I have to understand what's going on. :blush: :+1:

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/BalloonPopup.java
```diff
@@ -52,56 +79,144 @@ public boolean isVisible() {
...
+  public void mouseReleased(MouseEvent e) {
+    // not used
+  }
+
+  @Override
+  public void mouseEntered(MouseEvent e) {
+    transparencyHandler.resetAnimationProgress();
+    transparencyHandler.update(true);
+
+    revalidate();
```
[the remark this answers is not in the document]
Taikelenn: You are correct, revalidate isn't needed

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/BalloonLabel.java
```diff
@@ -0,0 +1,42 @@
+package fi.aalto.cs.apluscourses.ui.ideactivities;
+
+import java.awt.AlphaComposite;
+import java.awt.Graphics;
+import java.awt.Graphics2D;
+import javax.swing.JLabel;
+
+public class BalloonLabel extends JLabel implements TransparentComponent {
+  private float transparencyCoefficient;
```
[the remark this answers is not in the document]
Taikelenn: I'll add the comment here as well

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/PopupTransparencyHandler.java
Taikelenn: Yes

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/BalloonPopup.java
Taikelenn: This is when the popup is placed near a component's side (not on the component) and in that case it's never transparent - this resets transparency level to fully opaque

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/OverlayPane.java
Taikelenn: `convertRectangle` only works for Rectangles. `AffineTransform` works on all RectangularShapes, including RoundedRectangle2D. Jaakko wanted rounded rectangles at some point, so this is a necessary change.

It also allows us to highlight buttons with more precision (since buttons are rounded rectangles in IntelliJ).

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/PopupTransparencyHandler.java
Taikelenn: You are correct

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/PopupTransparencyHandler.java
Taikelenn: You are somehow right, however the bottom line is that `shouldRedraw` is now useless. It was introduced to only redraw the component during animation, but has since been replaced by `isInAnimation`.

We need to redraw the component while the animation is running because this makes the animation look smooth. We also cannot just redraw the component every frame ("GUI update") because it'll consume 9000% of the CPU cycles.

jaakkonarhi (approved):
Works well

OlliKiljunen (approved):
Thanks for clarifications. Very well done PR now! 👍

PR #657, opened 2021-06-10
Hyperlink to A+ Token Action

### LINKED ISSUE #509: Hyperlink to A+ Token Action
In Assignments view the label informing the user to set the A+ token could be turned into a hyperlink that when clicked would trigger the action. The priority for this is very low since this action is not expected to be performed very often. Also, these hyperlink-labels (both in Modules and Assignments views) could be styled more like a link (e.g. with the use of LinkLabel).

jaakkonarhi:
# Description of the PR

+ Closes #509 
+ You can't change the color of the text 

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
@@ -48,6 +48,7 @@ public class ExercisesView {
     cl = (CardLayout) cardPanel.getLayout();
     exerciseGroupsTree.getEmptyText().setText("");
     exerciseGroupsTree.setOpaque(true);
+    exerciseGroupsTree.addMouseListener(mouseAdapter);
     emptyText.setText(getText("ui.exercise.ExercisesView.loading"));
     emptyText.setHorizontalAlignment(SwingConstants.CENTER);
     emptyText.setVerticalAlignment(SwingConstants.CENTER);
@@ -73,14 +74,12 @@ public class ExercisesView {
         return;
       }
 
-      exerciseGroupsTree.removeMouseListener(mouseAdapter);
       if (viewModel.isProjectReady()) {
         emptyText.setText(getText("ui.module.ModuleListView.turnIntoAPlusProject"));
         if (viewModel.isAuthenticated()) {
           exerciseGroupsTree.getEmptyText().setText(
                   getText("ui.exercise.ExercisesView.allAssignmentsFiltered"));
         } else {
-          exerciseGroupsTree.addMouseListener(mouseAdapter);
           exerciseGroupsTree.getEmptyText().setText(
                   getText("ui.exercise.ExercisesView.setToken"));
           exerciseGroupsTree.getEmptyText().appendLine(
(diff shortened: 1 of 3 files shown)


stellatsv (changes requested):
This works great! One note would be to move the listener class from `ExercisesView` to `APlusToolWindowFactory`. You can look at how this is done for the `ModulesView`.

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -106,4 +114,16 @@ public JLabel getEmptyTextLabel() {
...
 
+  private class NoTokenMouseAdapter extends MouseAdapter {
+    @Override
+    public void mouseClicked(MouseEvent e) {
+      if (exerciseGroupsTree.isEmpty()
+          && exerciseGroupsTree.getEmptyText().getText().contains(
+              getText("ui.exercise.ExercisesView.setToken"))) {
+        DataContext context = DataManager.getInstance().getDataContext(e.getComponent());
+        ActionUtil.launch(APlusAuthenticationAction.ACTION_ID, context);
+      }
```
OlliKiljunen, line 125: Is there a need for consuming the event?

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -67,12 +73,14 @@ public void viewModelChanged(@Nullable ExercisesTreeViewModel viewModel) {
         return;
       }
 
+      exerciseGroupsTree.removeMouseListener(mouseAdapter);
```
OlliKiljunen, line 76: Do we need to remove and re-add the listener always when the viewmodel changes? Can it just be there all the time?

OlliKiljunen (approved):
I don't see any remarkable problem in this implementation, and I don't see much reason to move the mouse event listener into the tool window factory.

Actually, how I would perhaps solved this is that the `TreeView` is given an action that it launches when it's clicked as empty. So, I would only have to override `processMouseEvent` in the TreeView and not work with the listeners. But this seems fair enough, too.

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
jaakkonarhi: I'm not sure about this, but I don't think so

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
jaakkonarhi: It can be there all the time

stellatsv (approved):

PR #659, opened 2021-06-11
Add separate error message for missing files

### LINKED ISSUE #402: More detailed error message when a module is deleted
When a module directory is deleted from the file system (not from the IDE but otherwise), the module shows as "Error" in module list. Double-clicking the module resets it, as expected, but the problem is "Error" text that should be more informative.

It should be noted that it actually is kind of an error: after the deletion, if you go to File->Project Structure->Modules, you see that the module is still listed there and its class path files are reported missing. But we should replace "Error" with something like "Files missing" in this situation.

A bit similar to #291 .

Taikelenn:
# Description of the PR
Closes #402.
This adds a new component state "files missing" which is set when the user deletes the module directory.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.


jaakkonarhi (approved):

OlliKiljunen (changes requested):
This doesn't work. `stateMonitor` is used for syncing the component installation; it shouldn't be used for other purposes. In particular, only way that external code should affect it, is `setUnresolved`, which tells the component it should resolve its own state again.

What I suggest is that we add `diagnoseError` (or something alike) method into the `Component` or its subclasses that tries to find out what could be the cause of the error (checking if files are missing etc.) and returns a magic number / enum value telling the cause (it could be cached until the state is changed next time). The view model could then use that and get a suitable textual description based on the value.

OlliKiljunen (approved):
Perfect! :+1:

PR #654, opened 2021-06-07
Exit options for tutorials

jaakkonarhi:
# Description of the PR

+ Opens an exit dialog if you click the dimmed area or the Cancel Tutorial button in the progress dropdown.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
        <li>- [ ] new <b>unit</b> tests created</li>
        <li>- [x] all <b>unit</b> tests pass</li>
      </ul>
    </td>
    <td>
      <ul>
        <li>- [ ] new <b>integration</b> tests created</li>
        <li>- [x] all <b>integration</b> tests pass</li>
      </ul>
    </td>
    <td>
      <ul>
        <li>- [ ] new <b>e2e</b> tests created</li>
        <li>- [x] all <b>e2e</b> tests pass</li>
      </ul>
    </td>
    <td>
      <ul>
        <li>- [x] <b>manual</b> testing went well</li>
      </ul>
    </td>
  </tr>
</table>  
  
# Have you updated the [TESTING.md](https://github.com/Aalto-LeTech/intellij-plugin/blob/master/TESTING.md) or other relevant documentation on _your_ branch?

- [ ] Yes.
- [ ] Not yet. I will do it next.
- [x] Not relevant.

PR #660, opened 2021-06-11
Periodically validate component status

jaakkonarhi:
# Description of the PR

+ Opens component if it's not visible and checks once a second if it's visible

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.


src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/EditorPresenter.java
```diff
@@ -1,17 +1,40 @@
...
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 public class EditorPresenter extends IntelliJComponentPresenterBase {
-  public EditorPresenter(@NotNull String instruction, @NotNull String info) {
-    super(instruction, info);
+  @NotNull
+  private final String path;
+
+  public EditorPresenter(@NotNull String instruction,
```
stellatsv, line 14: The constructor could be private.

stellatsv (approved):
The functionality is very good! I have left one comment regarding the code.

src/main/java/fi/aalto/cs/apluscourses/model/task/Task.java
```diff
@@ -96,5 +105,24 @@ public synchronized boolean startTask(ActivityFactory activityFactory) {
...
+    @Override
+    public void run() {
+      ApplicationManager.getApplication().invokeLater(() -> {
+        if (presenter == null) {
+          return;
+        }
+        if (!presenter.isVisible()) {
+          presenter.highlight();
+        }
+      });
```
OlliKiljunen, line 124: Here we have at least a couple of threading issues, I'm afraid...
1) `presenter` can become null after the null-check so line 121 can throw a null pointer exception.
2) `presenter.highlight()` can also be called after `endTask` has executed `presenter.removeHighlight()`, leaving the highlight on when the task is done. (NOTE: Changing `endTask` in a way that it sets `presenter` null first and, only after that, calls `removeHighlight` with a locally stored reference WON'T fix this!)

All in all, I find this design kind of error-prone, and difficult to reason about. I strongly suggest that we move refreshing logic (timer) inside the presenters. If I have understood the issue correctly, the fact that highlighting can suddenly just disappear is due to some weird behaviour of IntelliJ/Swing. So, why would we want that anomaly "leak" onto our `Presenter` interface? IMO, it would make much more sense if `Task` could use presenters expecting them to work as they should be (without spurious disappearances) and they are presenter objects who make sure this is fulfilled.

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/ComponentDatabase.java
```diff
@@ -33,10 +38,44 @@
   }
 
   public static @Nullable Component getProgressButton() {
-    return ComponentLocator.getComponentsByClass("JButton")
-        .stream().filter(c -> c instanceof JButton)
-        .filter(c -> ComponentLocator.hasActionOfClass((JButton) c, "ProgressAction"))
-        .findFirst().orElse(null);
+    return null; // Changed in #654
```
OlliKiljunen, line 41: So, is this a line that will be overwritten by #654 ?

Otherwise, changes in this file look very fine!

src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/IntelliJComponentPresenterBase.java
```diff
@@ -56,4 +69,11 @@ private void removeHighlightInternal() {
   }
 
   protected abstract GenericHighlighter getHighlighter();
+
+  protected abstract boolean tryToShow();
+
+  @Override
+  public boolean isVisible() {
+    return getHighlighter() != null && getHighlighter().getComponent().isVisible();
```
OlliKiljunen, line 77: Store the highlighter in a local variable. Here, the second call could actually return null.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/IntelliJComponentPresenterBase.java
```diff
@@ -28,8 +33,16 @@ public void highlight() {
...
   @RequiresEdt
   private void highlightInternal() {
+    if (overlayPane != null) {
+      overlayPane.remove();
+    }
+
     GenericHighlighter highlighter = getHighlighter();
     if (highlighter == null) {
+      if (tryToShow()) {
+        highlightInternal();
```
OlliKiljunen, line 43: Possible livelock if `getHiglighter` always returns null and `tryToShow` true (well, not that much "livelock" as it ends up in a stack overflow quite soon). Can't the logic in `tryToShow` be handled by highlighter itself?

OlliKiljunen (changes requested):
Good work! :+1: 

However, there are synchronization issues in **Task.java** that need to be fixed. I hope you also consider my refactoring ideas.

Didn't test it.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/EditorPresenter.java
```diff
@@ -1,17 +1,40 @@
...
+    super(instruction, info, project);
+    this.path = path;
+  }
+
+  @NotNull
+  public static EditorPresenter create(@NotNull String instruction,
+                                       @NotNull String info,
+                                       @NotNull Project project,
+                                       @NotNull Arguments actionArguments) {
+    return new EditorPresenter(instruction, info, project, actionArguments.getString("filePath"));
```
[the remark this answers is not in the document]
jaakkonarhi: in some cases the file path would be in both arguments, so maybe the arguments should be combined

src/main/java/fi/aalto/cs/apluscourses/ui/ideactivities/ComponentDatabase.java
jaakkonarhi: yes

src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/IntelliJComponentPresenterBase.java
jaakkonarhi: i do that in later prs, should i change it here too

jaakkonarhi: i changed it here too

src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/IntelliJComponentPresenterBase.java
jaakkonarhi: the logic can't be in the highlighter, because the GenericHighlighter doesn'r have enough info about the component

src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/IntelliJComponentPresenterBase.java
```diff
@@ -56,4 +73,33 @@ private void removeHighlightInternal() {
...
+
+  private class TaskRefresher extends TimerTask {
+
+    @Override
+    public void run() {
+      ApplicationManager.getApplication().invokeLater(() -> {
+        if (!isVisible()) {
+          highlight();
+        }
+      });
```
OlliKiljunen, line 102: Remark: This calls `highlight` which is also internally wrapped in `invokeLater`. Is there perhaps some unnecessary `invokeLater`s here. (Having nested `invokeLater`s is not a problem but usually a sign of unclear organization of code.)

src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/ProjectTreePresenter.java
```diff
@@ -1,17 +1,29 @@
...
+                              @NotNull String info,
+                              @NotNull Project project) {
+    super(instruction, info, project);
   }
 
   @Override
   protected @Nullable GenericHighlighter getHighlighter() {
-    return new GenericHighlighter(ComponentDatabase.getProjectPane());
+    var component = ComponentDatabase.getProjectPane();
+    if (component == null && tryToShow()) {
```
OlliKiljunen, line 19: I see this method as a "getter" but it apparently have quite a visible side effect: it displays the project tree. I'd strongly suggest reconsidering this.

OlliKiljunen (changes requested):
Thank you for addressing my earlier comments. You improved this a lot. There are still some things I'd like to see changed, or at least discuss about. See my comments. :)

src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/IntelliJComponentPresenterBase.java
OlliKiljunen: You fixed the livelock / stack overflow issue, anyway. :+1:

src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/ProjectTreePresenter.java
jaakkonarhi: oops i left that there accidentally

src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/IntelliJComponentPresenterBase.java
```diff
@@ -56,12 +73,33 @@ private void removeHighlightInternal() {
...
+
+  private void startTimer() {
+    timer.scheduleAtFixedRate(new TaskRefresher(), REFRESH_INTERVAL, REFRESH_INTERVAL);
+  }
+
+  private class TaskRefresher extends TimerTask {
+    @Override
+    public void run() {
+      if (!isVisible()) {
+        highlight();
```
OlliKiljunen, line 101: `highlight` calls `highlightInternal` which calls `startTimer` which creates a new `TaskRefresher`. So there will be an increasing number of `TaskRefresher`s...

I suggest that `highlight`, not `highlightInternal`, starts the timer and both `highlight` and `TaskRefresher::run` call `highlightInternal`. Would that make sense to you?

OlliKiljunen (changes requested):
Very good! In particular, I like how this code gets simpler in each iteration. There's still one error I think. Otherwise, very nice code.

OlliKiljunen (approved):
Excellent! Thank you for fixing this. The code looks great now! 👍

PR #692, opened 2021-07-19
Force UTF-8 in JSON responses

Taikelenn:
# Description of the PR

By [RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259) section 8.1, JSON must be encoded as UTF-8. However this doesn't occur automatically for some reason (perhaps due to non-standard Content-Type header in A+ responses), so we enforce the proper encoding.

In practice, this used to be an issue only on Windows because AFAIK modern Linux distros use UTF-8 everywhere. Windows, however, uses language-specific code pages :(

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/model/Course.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/model/Course.java
@@ -13,6 +13,7 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -190,7 +191,7 @@ public abstract class Course implements ComponentSource {
       throws IOException, MalformedCourseConfigurationException {
     InputStream inputStream = CoursesClient.fetch(url);
     return Course.fromConfigurationData(
-        new InputStreamReader(inputStream), url.toString(), modelFactory);
+        new InputStreamReader(inputStream, StandardCharsets.UTF_8), url.toString(), modelFactory);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/main/java/fi/aalto/cs/apluscourses/model/Course.java
```diff
@@ -190,7 +191,7 @@ public static Course fromUrl(@NotNull URL url, @NotNull ModelFactory modelFactor
       throws IOException, MalformedCourseConfigurationException {
     InputStream inputStream = CoursesClient.fetch(url);
     return Course.fromConfigurationData(
-        new InputStreamReader(inputStream), url.toString(), modelFactory);
+        new InputStreamReader(inputStream, StandardCharsets.UTF_8), url.toString(), modelFactory);
```
OlliKiljunen, line 194: Hmm... 🤔 Ideally, this kind of manipulation would happen much closer to the DAL but it's not a problem of this PR, _per se_.

OlliKiljunen (approved):
Very good! 👍

src/main/java/fi/aalto/cs/apluscourses/model/Course.java
Taikelenn: I guess ideally we should have a method CoursesClient.fetchJSON which would be used for reading all JSONs and this method would already handle UTF-8

jaakkonarhi (approved):
nice

PR #694, opened 2021-07-20
Lazy loading

jaakkonarhi:
# Description of the PR

+ Loads exercises and submission results after opening exercise group

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.


src/main/java/fi/aalto/cs/apluscourses/intellij/model/ExercisesUpdater.java
```diff
@@ -121,20 +125,44 @@ private void addExercises(@NotNull List<ExerciseGroup> exerciseGroups,
...
+          }
+          var exercise = dataSource.getExercise(exerciseId, points, course.getTutorials(),
+              authentication, ZonedDateTime.now().minusDays(7));
+          exerciseGroup.addExercise(exercise);
+
+          progress.increment();
+
+          addSubmissionResults(exercise, points, authentication, progress);
+
+          courseProject.setExerciseTree(exercisesTree);
```
OlliKiljunen, line 145: This is set multiple times because it is in the loop. Intended? Anyway, it shouldn't.

src/main/java/fi/aalto/cs/apluscourses/model/DummyExercise.java
```diff
@@ -0,0 +1,38 @@
+package fi.aalto.cs.apluscourses.model;
+
+import java.util.OptionalLong;
+import org.jetbrains.annotations.NotNull;
+import org.json.JSONObject;
+
+public class DummyExercise extends Exercise {
```
OlliKiljunen, line 7: Does this class actually differ from `Exercise`? It doesn't override anything. :thinking:

OlliKiljunen (commented):
I didn't finish my review yet, because I have a question first.

This seems like a quite sophisticated solution but also a bit more complicated than what I would have expected. Especially, in my concern there is the `CourseProject` as a coordinator object responsible for the whole process.

Have you considered a design where each dummy object would take care of "upgrading" itself when it is expanded? You could use "Proxy" pattern for that. That is, only dummies are created by other objects. Then when a dummy gets signal that it should "upgrade", it constructs an actual object and holds a reference to that. After upgraded, the dummy delegates all the calls to it to the actual object.

Would this make any sense to you?

src/main/java/fi/aalto/cs/apluscourses/intellij/model/ExercisesUpdater.java
jaakkonarhi: this way you can see the results as they load in

src/main/java/fi/aalto/cs/apluscourses/model/DummyExercise.java
jaakkonarhi: i wanted to make a separate class to use instanceOf

jaakkonarhi:
The proxy pattern seems more complicated to me than my way, so I prefer my way

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
```diff
@@ -48,7 +50,7 @@
     return (SelectableNodeViewModel<?>) TREE_MODEL_BUILDER.getUserObject(node);
   }
 
-  private BaseTreeViewModel<?> viewModel = null;
+  protected BaseTreeViewModel<?> viewModel = null;
```
OlliKiljunen, line 53: We don't want to set this field from the subclasses, do we? So perhaps just adding a getter to it makes more sense than changing the visibility of the field.

src/main/java/fi/aalto/cs/apluscourses/ui/exercise/ExercisesView.java
```diff
@@ -126,11 +127,26 @@ public JLabel getEmptyTextLabel() {
...
 
+  private static class ExercisesTreeView extends TreeView {
+    @Override
+    public void setViewModel(@Nullable BaseTreeViewModel<?> viewModel) {
+      if (this.viewModel != null
+          && viewModel != null
+          && this.viewModel.getModel() instanceof ExercisesTree
+          && viewModel.getModel() instanceof ExercisesTree
+          && ((ExercisesTree) this.viewModel.getModel()).getSelectedStudent()
+          != ((ExercisesTree) viewModel.getModel()).getSelectedStudent()) {
```
OlliKiljunen, line 145: Store `this.viewModel` and model in a local variables.

OlliKiljunen (changes requested):
Alright!

I guess we can go with this design for now at least. There are, however, quite a many smaller issues I'd like to see fixed. See my comments.

Though, overall pretty nice PR.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/ExercisesUpdater.java
```diff
@@ -147,21 +174,40 @@ private void addSubmissionResults(@NotNull Exercise exercise,
...
-      var cachePreference = submissionsInGrading.remove(id)
-          ? CachePreferences.GET_NEW_AND_KEEP
-          : CachePreferences.GET_MAX_ONE_WEEK_OLD;
-      var submission = dataSource.getSubmissionResult(
-          baseUrl + id + "/", exercise, authentication, cachePreference);
-      if (submission.getStatus() == SubmissionResult.Status.WAITING) {
-        submissionsInGrading.add(id);
+      SubmissionResult submission;
+      if (exercise instanceof DummyExercise) {
+        submission = new DummySubmissionResult(id, exercise);
```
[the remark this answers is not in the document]
jaakkonarhi: They need the results so the assistants can search for submissions using the id

OlliKiljunen (approved):
Thank you for fixing those things I pointed out in my previous review. This looks good now, even though I'm still thinking that the overall design is more complicated that it ideally should be, but we've discussed this thoroughly enough and accepted the possibility of future problems, so let's go on with this.

Anyway, this is an important feature and all but easy to implement, so I really appreciate that you tackled this and solved it. Great work! 👍 Sorry for it taking so long to review from me.

Taikelenn (approved):

Issue #641, opened 2021-05-17
Project is disposed


jaakkonarhi:
bannertime branch
```
java.lang.IllegalStateException: Project is disposed.
	at fi.aalto.cs.apluscourses.intellij.model.APlusProject.getProject(APlusProject.java:37)
	at com.intellij.openapi.application.impl.ApplicationImpl.runReadAction(ApplicationImpl.java:844)
	at com.intellij.openapi.application.ReadAction.compute(ReadAction.java:61)
	at fi.aalto.cs.apluscourses.intellij.model.APlusProject.getBasePath(APlusProject.java:44)
	at fi.aalto.cs.apluscourses.intellij.model.IntelliJModule.getFullPath(IntelliJModule.java:55)
	at fi.aalto.cs.apluscourses.intellij.model.IntelliJModule.hasLocalChanges(IntelliJModule.java:139)
	at fi.aalto.cs.apluscourses.model.Module.hasLocalChanges(Module.java:143)
	at fi.aalto.cs.apluscourses.ui.module.ColoredModuleListRenderer.customizeCellRenderer(ColoredModuleListRenderer.java:34)
	at fi.aalto.cs.apluscourses.ui.module.ColoredModuleListRenderer.customizeCellRenderer(ColoredModuleListRenderer.java:14)
	at com.intellij.ui.ColoredListCellRenderer.getListCellRendererComponent(ColoredListCellRenderer.java:47)
	at com.intellij.ui.ExpandedItemListCellRendererWrapper.getListCellRendererComponent(ExpandedItemListCellRendererWrapper.java:24)
	at com.intellij.ui.components.WideSelectionListUI.updateLayoutState(WideSelectionListUI.java:197)
	at java.desktop/javax.swing.plaf.basic.BasicListUI.maybeUpdateLayoutState(BasicListUI.java:1389)
	at java.desktop/javax.swing.plaf.basic.BasicListUI.locationToIndex(BasicListUI.java:1004)
	at java.desktop/javax.swing.JList.locationToIndex(JList.java:1600)
	at java.desktop/javax.swing.JList.getToolTipText(JList.java:1535)
	at com.intellij.ide.IdeTooltipManager.isTooltipDefined(IdeTooltipManager.java:248)
	at com.intellij.ide.IdeTooltipManager.showForComponent(IdeTooltipManager.java:217)
	at com.intellij.ide.IdeTooltipManager.maybeShowFor(IdeTooltipManager.java:203)
[...]

PR #699, opened 2021-07-22
Ideact module dependencies

jaakkonarhi:
# Description of the PR

+ Downloads module called "Ideact_$moduleName" when tutorial started
+ Shows a notification when deps downloaded
+ Removes the module when tutorial ended
+ If downloadDependencies false, checks if modules exist

```
    "moduleDependencies": ["Football1"],
    "downloadDependencies": false,
    "deleteDependencies": false,
```

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.


src/main/java/fi/aalto/cs/apluscourses/intellij/actions/TutorialAction.java
```diff
@@ -87,13 +88,16 @@ private void doTutorial(@NotNull Project project) {
...
     TutorialViewModel tutorialViewModel =
         new TutorialViewModel(tutorialExercise, new IntelliJActivityFactory(project),
-            new TaskNotifier(notifier, project), dialogs);
+            taskNotifier, dialogs);
     if (dialogs.confirmStart(tutorialViewModel)) {
+      tutorialViewModel.getTutorial().downloadDependencies(courseViewModel.getModel(), project, taskNotifier);
       mainViewModelProvider.getMainViewModel(project).tutorialViewModel.set(tutorialViewModel);
       tutorialViewModel.getTutorial().tutorialCompleted
-          .addListener(mainViewModel, this::onTutorialComplete);
+          .addListener(mainViewModel, mainVm -> onTutorialComplete(mainVm, courseViewModel.getModel()));
```
OlliKiljunen, line 100: `Event`s (and `ObservableProperty`s) were designed to handle data flow from model to view. That's why they use weak references. The idea was that the view gets updated as long as it exists and when it no longer exists, it will be GC'd even if the model was still there. Another safe use case of these is data flow from "child objects" to "parent objects" (like from `Module` to `Course`), when parent already holds references to its children. If a child gets removed from a parent, it will get eventually GC'd (though the parent can still get updates from it before GC steps in).

To me, it seems this use case doesn't fit in either of the above-mentioned, so perhaps using our `Event` class is not the best idea here... 🤔 In practice, the weak referencing probably doesn't matter here, though, because the main view model is probably alive longer than the tutorial. But it is difficult to see here whether it's correct or not. The lambda creating a closure with `this` (an action object) and `courseViewModel` makes it even more complicated.

Perhaps just a crude observer-implementation without weak referencing would make more sense. We can discuss it, if you like.

OlliKiljunen (commented):
Now there's a thing that I can't remember this feature. Why do we want this?

Anyway, the code looks pretty good, though there's that one issue that perhaps we should discuss further. It's not originated from this PR but this PR make it quite clear there's something shady in that design.

stellatsv:
> Now there's a thing that I can't remember this feature. Why do we want this?
> 
> Anyway, the code looks pretty good, though there's that one issue that perhaps we should discuss further. It's not originated from this PR but this PR make it quite clear there's something shady in that design.

This feature is useful in the sense that it provides a clean slate for the student which corresponds directly to what the teacher had in mind when making the tutorial. It would however be useful to keep the module and not delete it directly after the Tutorial is done. One way to achieve this would perhaps be to delete the module only if the student chooses to redo the Tutorial? We had also discussed the possibility of making this feature optional. What are your thoughts on this @jaakkonarhi @OlliKiljunen?

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/TutorialAction.java
OlliKiljunen: This concern still holds.

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
```diff
@@ -115,6 +122,9 @@ public boolean isUpdatable() {
     if (stateMonitor.get() != LOADED) {
       return false;
     }
+    if (forceUpdatable.get()) {
```
OlliKiljunen, line 125: Is this never re-set?

src/main/java/fi/aalto/cs/apluscourses/model/Tutorial.java
```diff
@@ -1,9 +1,19 @@
 package fi.aalto.cs.apluscourses.model;
 
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.ui.Messages;
+import fi.aalto.cs.apluscourses.intellij.model.ProjectModuleSource;
+import fi.aalto.cs.apluscourses.intellij.notifications.TaskNotifier;
+import fi.aalto.cs.apluscourses.intellij.utils.EdtTask;
```
OlliKiljunen, line 7: We don't want to refer to these classes from `model` package.

src/main/java/fi/aalto/cs/apluscourses/model/Tutorial.java
```diff
@@ -45,4 +75,97 @@ public static Tutorial fromJsonObject(@NotNull JSONObject jsonObject) {
...
+   */
+  public void downloadDependencies(@NotNull Course course,
+                                   @NotNull Project project,
+                                   @NotNull TaskNotifier taskNotifier) {
+    taskNotifier.notifyDownloadingDeps(false);
+
+    var modules = getModules(course);
+
+    var componentInstallerFactory = new ComponentInstallerImpl.FactoryImpl<>(new SimpleAsyncTaskManager());
+    var dialogs = new ReinstallDialogs(project);
```
OlliKiljunen, line 90: These should be given as parameters. Actually, the functionality of this whole method would better be handled in the caller (the action).

src/main/java/fi/aalto/cs/apluscourses/model/Tutorial.java
```diff
@@ -45,4 +75,97 @@ public static Tutorial fromJsonObject(@NotNull JSONObject jsonObject) {
...
+        @Override
+        protected Integer execute() {
+          return Messages.showYesNoDialog(project,
+              PluginResourceBundle.getAndReplaceText("ui.reinstallDialogs.confirm", name),
+              name,
+              Messages.getWarningIcon());
+        }
+      }.executeAndWait() == Messages.YES;
+    }
+  }
```
OlliKiljunen, line 170: This is a wrong place for framework-specific implementation class.

src/main/java/fi/aalto/cs/apluscourses/model/Tutorial.java
```diff
@@ -45,4 +75,97 @@ public static Tutorial fromJsonObject(@NotNull JSONObject jsonObject) {
...
+  /**
+   * Deletes the module deps for a tutorial.
+   */
+  public void deleteDependencies(@NotNull Course course) {
+    for (var component : getModules(course)) {
+      try {
+        component.remove();
+        component.unload();
+      } catch (IOException e) {
+        // fail silently
```
OlliKiljunen, line 111: Are you sure?

OlliKiljunen (changes requested):
Thanks for taking into account our discussion related to the optionality of the feature! However, this PR is still quite convoluted of its design. Especially, considering this is not a mandatory feature, it should complicate things as slightly as possible, which seems not to be the case, currently.

IMO, this PR should, thus, be more refactored before it can be merged. I wrote some comments but I'm also happy to discuss further if needed.

src/main/java/fi/aalto/cs/apluscourses/model/Module.java
jaakkonarhi: It's not

src/main/java/fi/aalto/cs/apluscourses/model/Tutorial.java
```diff
@@ -45,4 +51,32 @@ public static Tutorial fromJsonObject(@NotNull JSONObject jsonObject) {
...
+  /**
+   * Deletes the module deps for a tutorial.
+   */
+  public abstract void deleteDependencies(@NotNull Course course, @NotNull Project project);
+
+
+  /**
+   * Returns true and shows notifications if modules are missing.
+   */
+  public abstract boolean dependenciesMissing(@NotNull Project project);
```
OlliKiljunen, line 69: Giving course or project (an IntelliJ class btw!) as an argument does not make sense here because there is already conceptual relationship between tutorials and classes/projects they belong to. Thus, they are not "free" parameters of these methods but only provided "for the sake of convenience" (giving a wrong course/project would probably lead to erroneous behavior). Thus, there almost certainly would be a better way to do it where the design would guarantee the use of correct course/project.

The use of `Project` is, as we all know, problematic anyway, but that hopefully gets fixed with #641.

OlliKiljunen (commented):
Thanks for addressing my comments! This PR has gradually improved but I have still my concern of unnecessary convolution it brings.

So, I'm still a bit against this feature, sorry. What is most recently discussed with Juha & co., do they want this feature now or could we just postpone it for a later release?

PR #759, opened 2021-09-26
Ensure that window stays constant between tutorial steps

Taikelenn:
# Description of the PR
Stops tutorials from "invading" other IntelliJ windows within the same process. This is achieved by obtaining the current window object at tutorial start and executing the highlighting method (`highlightInternal`) only when the active window is the one that started the tutorial.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>e2e</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/IntelliJComponentPresenterBase.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/IntelliJComponentPresenterBase.java
@@ -3,6 +3,7 @@ package fi.aalto.cs.apluscourses.intellij.model.task;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.ModalityState;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.wm.WindowManager;
 import com.intellij.util.concurrency.annotations.RequiresEdt;
 import fi.aalto.cs.apluscourses.model.task.CancelHandler;
 import fi.aalto.cs.apluscourses.model.task.ComponentPresenter;
@@ -12,6 +13,7 @@ import fi.aalto.cs.apluscourses.ui.ideactivities.OverlayPane;
 import java.util.Timer;
 import java.util.TimerTask;
 import javax.swing.Action;
+import javax.swing.JOptionPane;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -53,6 +55,11 @@ public abstract class IntelliJComponentPresenterBase implements ComponentPresent
 
   @RequiresEdt
   private void highlightInternal() {
+    // if we are focused on another window than the tutorial one, don't touch anything
+    if (JOptionPane.getRootFrame() != WindowManager.getInstance().getFrame(project)) {
+      return;
+    }
+
     if (overlayPane == null) {
       overlayPane = OverlayPane.installOverlay();
     }


src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/IntelliJComponentPresenterBase.java
```diff
@@ -46,6 +48,14 @@ public void highlight() {
 
   @RequiresEdt
   private void highlightInternal() {
+    var tutorialViewModel = PluginSettings.getInstance().getMainViewModel(project).tutorialViewModel.get();
+    var rootFrame = tutorialViewModel != null ? tutorialViewModel.getTutorialWindowFrame() : null;
+
+    // if we are focused on another window than the tutorial one, don't touch anything
+    if (JOptionPane.getRootFrame() != rootFrame) {
```
OlliKiljunen, line 55: To me, it seems that the condition could be replaced with
```java
JOptionPane.getRootFrame() != WindowManager.getInstance().getFrame(project)
```

src/main/java/fi/aalto/cs/apluscourses/presentation/ideactivities/TutorialViewModel.java
```diff
@@ -52,6 +58,7 @@ public TutorialViewModel(@NotNull TutorialExercise tutorialExercise,
     currentTaskIndex = 1;
     unlockedIndex = 1;
     tasksAmount = tasks.size();
+    tutorialWindowFrame = JOptionPane.getRootFrame();
```
OlliKiljunen, line 61: This gets the focused window frame at the moment, right? I don't think it's safe to assume that at the time this object is created, the related window is necessarily focused.

src/main/java/fi/aalto/cs/apluscourses/presentation/ideactivities/TutorialViewModel.java
```diff
@@ -6,7 +6,9 @@
 import fi.aalto.cs.apluscourses.model.task.ActivityFactory;
 import fi.aalto.cs.apluscourses.model.task.Task;
 import fi.aalto.cs.apluscourses.utils.APlusLocalizationUtil;
+import java.awt.Frame;
 import java.util.List;
+import javax.swing.JOptionPane;
```
OlliKiljunen, line 11: I don't take pleasure in seeing AWT/Swing classes in view model classes... :)

OlliKiljunen (changes requested):
There shouldn't be a reason to store `rootFrame`, at least in a view model.

First, this whole solution seemed a bit hacky to me. Like, why we have to skip the refreshment of the overlay pane if another window is focused? Can't we just make sure the overlay is drawn on that window where it should be. Nevertheless, I think I'm now starting to understand the problems related to the "pane" nature of overlay panes; that they're not window-related _per se_. So I think I'm fine with this solution, in principle. However, we should get rid of storing the AWT frame inside a view model.

src/main/java/fi/aalto/cs/apluscourses/intellij/model/task/IntelliJComponentPresenterBase.java
Taikelenn: I wasn't aware of WindowManager

Taikelenn:
> 
> 
> There shouldn't be a reason to store `rootFrame`, at least in a view model.
> 
> First, this whole solution seemed a bit hacky to me. Like, why we have to skip the refreshment of the overlay pane if another window is focused? Can't we just make sure the overlay is drawn on that window where it should be. Nevertheless, I think I'm now starting to understand the problems related to the "pane" nature of overlay panes; that they're not window-related _per se_. So I think I'm fine with this solution, in principle. However, we should get rid of storing the AWT frame inside a view model.

The root frame issue is not isolated to the overlay. Frankly, OverlayPane is one of the few components that works correctly - once constructed, it always draws on the same window. The real problem is ComponentDatabase and the usage of OverlayPane - ComponentDatabase looks for a component by enumerating them starting from the root frame, and if another project (=window) is currently active, then it's that exact window that will be enumerated.

This manifests itself in the TaskRefresher::run() method of IntelliJComponentPresenterBase.java:
1. `ComponentDatabase.getProgressButton()` is called - it returns null, because the another window does not have a running tutorial and thus does not have a progress button. This triggers a call to `highlightInternal()`.
2. `HighlightInternal` notices that an overlay is currently installed - because of this, it is removed and installed again, which means that a new, incorrect root frame is obtained.
3. The overlay is now installed on the wrong window. To make matters worse: depending on the current presenter, the `tryToShow()` call will cause the IDE to forcefully snap back to the tutorial window.
4. If you try to switch back to the other window, `run()` will be called again and the process repeats from step 1.

This can be solved in two ways:
1. Stop using `getRootFrame()` and instead pass the root frame object to all methods that require it. We could use this `WindowManager` stuff to get the proper root frame, but all it changes is that we'll now be passing a Project object instead of a JFrame.
2. Don't call `highlightInternal()` if another frame is focused. This is a tempting method because it doesn't require refactoring anything, it will still work correctly (because the root frame is guaranteed to be the same as the tutorial's root frame) and it will also fix stealing focus back to the tutorial (which method 1 doesn't fix). With WindowManager this might just be reduced to two lines of code at most.

src/main/java/fi/aalto/cs/apluscourses/presentation/ideactivities/TutorialViewModel.java
Taikelenn: Agreed, this will be removed anyway.

src/main/java/fi/aalto/cs/apluscourses/presentation/ideactivities/TutorialViewModel.java
Taikelenn: Me neither, but as a read-only field? Only used for comparison? Pretty please? :3

Jokes aside, this will be removed and replaced with WindowManager

OlliKiljunen (approved):
Thank you for addressing my comment and especially thanks for explaining the problem out thoroughly. I agree with your reasoning. This is a valid solution. :+1:

Perhaps at some point, component database should be changed to be project-specific. If I got you right, that could solve it. But now, let's go with this.

PR #842, opened 2022-02-21
Remove writing to final field

### LINKED ISSUE #841: BannerView modifying final field

Taikelenn:
# Description of the PR
Closes #841 by overriding a method instead of setting a field that has become final in IntelliJ IDEA 2022
@OlliKiljunen should the `backgroundColor` field be volatile? After all, there are no guarantees regarding which threads are these two methods inside `MyEditorNotificationPanel` restricted to.

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/ui/MyEditorNotificationPanel.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/ui/MyEditorNotificationPanel.java
@@ -4,7 +4,14 @@ import com.intellij.ui.EditorNotificationPanel;
 import java.awt.Color;
 
 public class MyEditorNotificationPanel extends EditorNotificationPanel {
+  private Color backgroundColor;
+
   public void setColor(Color color) {
-    this.myBackgroundColor = color;
+    this.backgroundColor = color;
+  }
+
+  @Override
+  public Color getBackground() {
+    return backgroundColor == null ? super.getBackground() : backgroundColor;
   }
 }


OlliKiljunen (commented):
Please see [my second comment](https://github.com/Aalto-LeTech/aplus-courses/issues/841#issuecomment-1046990102) on the ticket. I'd like to discuss the necessity to inherit `EditorNotificationPanel` before reviewing this ticket.

OlliKiljunen:
> @OlliKiljunen should the backgroundColor field be volatile? After all, there are no guarantees regarding which threads are these two methods inside MyEditorNotificationPanel restricted to.

Methods in Swing (such as `getBackground()`) [are not thread safe](https://docs.oracle.com/en/java/javase/11/docs/api/java.desktop/javax/swing/package-summary.html#threading) and shouldn't be called outside EDT. We can (and should) set the similar restriction for `setColor()` method. Now, there's a guarantee. :) So it doesn't have to volatile.

Taikelenn:
> Please see [my second comment](https://github.com/Aalto-LeTech/aplus-courses/issues/841#issuecomment-1046990102) on the ticket. I'd like to discuss the necessity to inherit `EditorNotificationPanel` before reviewing this ticket.

I'm pretty sure it was used solely for convenience reasons (i.e. already implemented & looks nice). It could certainly be a regular JPanel

jaakkonarhi (approved):

Taikelenn:
> > > Please see [my second comment](https://github.com/Aalto-LeTech/aplus-courses/issues/841#issuecomment-1046990102) on the ticket. I'd like to discuss the necessity to inherit `EditorNotificationPanel` before reviewing this ticket.
> > 
> > 
> > I'm pretty sure it was used solely for convenience reasons (i.e. already implemented & looks nice). And perhaps consistency with IntelliJ-styled banners in the editor section. It could certainly be a regular JPanel, but if JetBrains already did the work (laying out the banner, making it consistent with themes) using it is objectively simpler
> 
> Inheritance shouldn't be used for the purpose of code re-use, but to model is-a relationships. Any idea how much work it would be take to re-implement the required functionality/design? (One can basically copy the relevant parts from `EditorNotificationPanel` because licensing ain't an issue between the projects.)

I kind of agree with that. To be honest, I wouldn't really mind using EditorNotificationPanel directly (without inheritance) if only it hadn't made the background color field final. Unfortunately they did (which, to me, is a questionable change - I don't see any reason why a Swing component would have an immutable color õ.õ).

Then again, it also depends on your point of view. If necessary, I can see the current implementation as follows: our class `MyEditorNotificationPanel` is an `EditorNotificationPanel` with the extended functionality of having a dynamic background color. It really does sound like inheritance, doesn't it?

Of course we can re-implement this banner from the basic JPanel, it shouldn't be more than ~60-70 lines of code. If you wish, I can make a parallel pull request with separate implementation using JPanel

OlliKiljunen:
> I can see the current implementation as follows: our class MyEditorNotificationPanel is an EditorNotificationPanel with the extended functionality of having a dynamic background color. It really does sound like inheritance, doesn't it?

The thing is, we don't use it for _editor_. If the original implementor of the superclass has thought it to be used only with editor windows that might cause more compability problems in the future.

> If you wish, I can make a parallel pull request with separate implementation using JPanel

I'd say a ticket of it is enough at this point.

OlliKiljunen (approved):
This apparently works as a hack. At some point, we should reimplement this class inheriting it directly from `JPanel`.

PR #850, opened 2022-03-11
Add support for submitting exercises by double-clicking an item

jaakkonarhi:
# Description of the PR

image
Also double-clicking submissions opens them in the IDE

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [ ] Yes.
- [x] Not relevant.


src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ShowFeedbackAction.java
```diff
@@ -29,7 +30,7 @@
 public class ShowFeedbackAction extends AnAction {
   public static final String ACTION_ID = ShowFeedbackAction.class.getCanonicalName();
 
-  private static final String[] SUPPORTED_COURSES = new String[] {"O1", "Programming Studio 2/A"};
+  public static final List<String> SUPPORTED_COURSES = List.of("O1", "Programming Studio 2/A");
```
Taikelenn, line 33: I feel that hardcoding names of courses in the code is the wrong approach. However, since this PR is only being merged into another branch, I'll leave the discussion about this to be continued in that other PR.

src/main/java/fi/aalto/cs/apluscourses/ui/base/TreeView.java
```diff
@@ -123,12 +122,12 @@ private void update() {
     setSelectionRows(selection);
   }
 
-  public void addNodeAppliedListener(ActionListener listener) {
-    nodeAppliedListeners.add(listener);
+  public void addNodeAppliedListener(@NotNull String key, @NotNull ActionListener listener) {
+    nodeAppliedListeners.put(key, listener);
```
Taikelenn, line 126: This is potentially confusing: `addNodeAppliedListener` suggest that the key is just some String, but then inside `mouseClicked` the key is actually expected to be a class string, not just a generic string.

If the change from a set to a map is necessary, consider changing the first parameter of this function to `Class<?>` and then doing `nodeAppliedListeners.put(key.getName(), listener)` to enforce this contract.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ShowFeedbackAction.java
OlliKiljunen: I think it should be the course config file that tells which features it uses. Even if this list was moved to resources or wherever, maintaining it would be challenging.

Furthermore, changing type to `List` suggests that this list is perhaps modified run-time, which it shouldn't be. I'd keep it as an array unless there's a good reason to have it as a list.

src/main/java/fi/aalto/cs/apluscourses/intellij/toolwindows/APlusToolWindowFactory.java
```diff
@@ -85,6 +86,8 @@ private static ExercisesView createExercisesView(@NotNull Project project) {
     MainViewModel mainViewModel = PluginSettings.getInstance().getMainViewModel(project);
 
     ExercisesView exercisesView = new ExercisesView();
+    mainViewModel.courseViewModel.addValueObserver(exercisesView, ((observer, value) -> observer.setCourseName(
+        Optional.ofNullable(value).map(c -> c.getModel().getName()).orElse(""))));
```
OlliKiljunen, line 90: The view shouldn't have setter, instead it should get notified of a change and update itself appropriately. That way, we can modify the view independently of the rest of the system. The logic here (and in `ExercisesView::setCourseName`), should ideally be moved to the view model so that the view could stay as "dumb" as possible.

OlliKiljunen (approved):
I wrote a couple of comments on these changes.

PR #859, opened 2022-03-14
Multiline REPL

Taikelenn:
# Description of the PR
This version of REPL will surround the multiline statement with bracketed paste markers and thus inform the REPL process to execute the entire statement as a single operation.


Taikelenn:
Todo:
- add a comment why the package is so weird (textSent is package private)
- explain why we're padding the buffer to 64 bytes
- why we're sending the "being bracketed paste" separately
- why 64 bytes and not some other number
- add some reference to ANSI sequences for bracketed paste

Taikelenn:
I will also do us all a favour and disable some (IMHO pointless Scalastyle checks:
- "Avoid using return" - yes, generally you should avoid it in Scala, but not here
- "Field name does not match..." - straight out wrong, "Constant names should be in upper camel case" according to https://docs.scala-lang.org/style/naming-conventions.html
- "Argument should be named" - makes the code unnecessarily verbose

src/main/scala/fi/aalto/cs/apluscourses/intellij/Repl.scala
```diff
@@ -31,6 +32,15 @@ class Repl(module: Module) extends ScalaLanguageConsole(module: Module) {
     }
   }, AWTEvent.FOCUS_EVENT_MASK)
 
+  // We need this here because the overridden ConsoleExecuteAction needs to determine whether
+  // the console is hosting a Scala 3 REPL or something else
+  val isScala3REPL: Boolean = ModuleUtils.nonEmpty(
+    ModuleRootManager.getInstance(module)
+      .orderEntries()
+      .librariesOnly()
+      .satisfying(x => x.getPresentableName.contains("scala3-") || x.getPresentableName.contains("scala-sdk-3."))
```
OlliKiljunen, line 41: This seems a bit error-prone, although I see that it may be the only way to resolve the Scala version. I wonder if it would be a good idea to better isolate this part due to its problematic nature?

src/main/scala/org/jetbrains/plugins/scala/console/apluscourses/ConsoleExecuteAction.scala
```diff
@@ -0,0 +1,104 @@
+// The reason for this class being in a separate package is that the ConsoleExecuteAction class
+// uses the ScalaLanguageConsole.textSent() method, which is package private. Therefore, in order to
+// call it, we must be in the same package as the console: org.jetbrains.plugins.scala.console.
+
```
OlliKiljunen, line 4: Should there also be some copyright notice that this code is copied (and modified) from **ScalaConsoleExecuteAction.scala** ?

src/main/scala/org/jetbrains/plugins/scala/console/apluscourses/ConsoleExecuteAction.scala
```diff
@@ -0,0 +1,104 @@
...
+  // This is yet another quirk of JLine - it supports various terminals, but IntelliJ's console
+  // isn't one, and therefore JLine defaults to "DumbTerminal", which is a terminal with no
+  // special features.
+  // When inserting bracketed paste sequences, the JLine library scans the input for the
+  // "bracketed paste end" sequence by reading from the terminal in 64-byte chunks.
+  // JLine erroneously blocks until the whole chunk is read, therefore we need to pad the data
+  // to 64 characters in order to appease JLine.
+  private val JLineBufferLength = 64
+
+  private def padTextToBlockLength(text: String): Array[Byte] = {
```
OlliKiljunen, line 36: This method would love unit tests ;)

src/main/scala/org/jetbrains/plugins/scala/console/apluscourses/ConsoleExecuteAction.scala
```diff
@@ -0,0 +1,104 @@
...
+      // see com.intellij.execution.impl.ConsoleViewImpl.print
+      console.flushDeferredText()
+      historyController.addToHistory(text)
+
+      editor.getCaretModel.moveToOffset(0)
+      editor.getDocument.setText("")
+    }
+
+    val outputStream: OutputStream = processHandler.getProcessInput
+    val finalText = text.split('\n').mkString("\n")
```
OlliKiljunen, line 86: Check if this is necessary

OlliKiljunen (approved):

src/main/scala/fi/aalto/cs/apluscourses/intellij/Repl.scala
Taikelenn: It doesn't really *need* to work in all potential cases for all projects; it just needs to work for projects created by the plugin. I feel like this is good enough for this purpose, but I wouldn't extract this code anywhere else precisely because I don't want to convey the message that the method works for all modules/projects and is therefore ready to use in other places in code.

src/main/scala/org/jetbrains/plugins/scala/console/apluscourses/ConsoleExecuteAction.scala
```diff
@@ -0,0 +1,104 @@
...
+    // Non-A+ REPLs or those that host Scala 2 should not be modified.
+    if (!console.isInstanceOf[Repl] || !console.asInstanceOf[Repl].isScala3REPL) {
+      super.actionPerformed(e)
+      return
+    }
+
+    val processHandler = ScalaConsoleInfo.getProcessHandler(editor)
+    val historyController = ScalaConsoleInfo.getController(editor)
+
+    val document = console.getEditorDocument
```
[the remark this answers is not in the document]
Taikelenn: I added a permalink to the beginning of the file indicating the source, this should be enough (after all, line numbers change)

PR #860, opened 2022-03-15
Hash submissions and detect duplicates

Taikelenn:
# Description of the PR
Currently draft
obraz

# Testing
<table>
  <tr>
    <th>unit</th>
    <th>integration</th>
    <th>manual</th>
  </tr>
  <tr>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
    <td>
      <ul>
      </ul>
    </td>
  </tr>
</table>  
  
- [X] Yes.
- [ ] Not relevant.


src/main/java/fi/aalto/cs/apluscourses/intellij/utils/Interfaces.java
```diff
@@ -88,6 +114,133 @@ public static void refreshPath(VirtualFile file, Runnable callback) {
...
+        files.entrySet().stream().sorted(Map.Entry.comparingByKey()).forEachOrdered(entry -> {
+          submissionString.append(entry.getKey());
+          submissionString.append('|'); // a separating character that is not in base-64 alphabet
+          submissionString.append(hashFileToString(shaDigest, entry.getValue()));
+          submissionString.append(',');
+        });
+
+        final String cacheKey = "hash_c" + courseId + "_e" + exerciseId;
+        final byte[] finalHashBytes = shaDigest.digest(submissionString.toString().getBytes(StandardCharsets.UTF_8));
+        final String finalHash = Base64.getEncoder().encodeToString(finalHashBytes);
```
OlliKiljunen, line 215: I guess this is ok. I would perhaps find it more intuitive if each file was first hashed (`<key>:<content>` to SHA256) and these hashes were XORred to make the final hash; that way the order wouldn't matter either. Hashing hashes always raises a concern of a diminishing domain space but in this case, it is really a non-issue because the number of hashes are so small.

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/Interfaces.java
```diff
@@ -88,6 +114,133 @@ public static void refreshPath(VirtualFile file, Runnable callback) {
...
+
+        final JSONObject cachedHashesJson = submissionsCache.getValue(cacheKey, CachePreferences.PERMANENT);
+        final SubmissionHashes cachedHashes = new SubmissionHashes(cachedHashesJson);
+        if (cachedHashesJson == null) {
+          // this is the first hash, so the entry doesn't yet exist in the cache
+          cachedHashes.addHash(finalHash);
+          submissionsCache.putValue(cacheKey, cachedHashes.toJsonObject(), CachePreferences.PERMANENT);
+          return true;
+        }
```
OlliKiljunen, line 224: I don't see reason for this conditional block. It could be removed and the code would function the same, right?

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/Interfaces.java
```diff
@@ -88,6 +114,133 @@ public static void refreshPath(VirtualFile file, Runnable callback) {
...
+          if (!DuplicateSubmissionDialog.showDialog()) {
+            return false;
+          }
+        } else {
+          cachedHashes.addHash(finalHash);
+        }
+
+        submissionsCache.putValue(cacheKey, cachedHashes.toJsonObject(), CachePreferences.PERMANENT);
+        return true;
+      } catch (UncheckedIOException | NoSuchAlgorithmException e) {
```
OlliKiljunen, line 236: Which method may throw `UncheckedIOException` (a RuntimeException). I would narrow down this `try` block, if possible.

OlliKiljunen (commented):
Looks very good! 👍  Still, I have a couple of questions I'd like to discuss before approval, so see my comments.

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/Interfaces.java
Taikelenn: Correct, this conditional block is a remnant of a past version that used different caching method

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/Interfaces.java
Taikelenn: The method `hashFileToString` can throw IOException. Normally I'd just mark that method with `throws IOException` and handle this instead of `UncheckedIOException`, but unfortunately the invocation of `hashFileToString` is inside a lambda, and for some reason a lambda inside a try-block is not actually covered by that block.

It's essentially a bypass for the IOException being a checked exception; in a sane programming language this shouldn't have been an issue whatsoever ～＿～

OlliKiljunen: Oh, I see. What about changing `forEachOrdered` to a normal `for` loop?

Taikelenn: This would have fixed it! But then the code wouldn't be beautiful, functional, Java 37 -style, but rather imperative, instant legacy code :pp
But one needs to sort the map anyway, so using streams all the way was more natural IMO. I'll change to a for loop if necessary.

Also, I don't really consider that try-block spanning the whole method a bad thing: I wouldn't want a potential exception blocking a user from submitting the exercise.

OlliKiljunen (approved):
Thank you for answering my questions. This is good to be merged now.

jaakkonarhi (approved):

Issue #839, opened 2022-02-17
Test with IntelliJ 2022.1 EAP


OlliKiljunen:
- [ ] Win10: @OlliKiljunen 
- [ ] Mac: @jaakkonarhi 
- [ ] Ubuntu: @Taikelenn

stellatsv:
We'll do a round of manual testing for each platform.

PR #865, opened 2022-03-22
EAP testing

Taikelenn:
# Description of the PR
As described in #839 we'll be doing testing in EAP:
- Windows @OlliKiljunen https://download.jetbrains.com/idea/ideaIC-221.5080.9.exe
- Linux @Taikelenn https://download.jetbrains.com/idea/ideaIC-221.5080.9.tar.gz
- Mac OS @jaakkonarhi https://download.jetbrains.com/idea/ideaIC-221.5080.9.dmg

Links to other EAP versions (if necessary): https://confluence.jetbrains.com/display/IDEADEV/IDEA+2022.1+latest+builds
Plugin .zip available on Slack.


Taikelenn:
| Manual test            |                                                    |
| ---------------------- | -------------------------------------------------- |
| Date                   | 2022-03-23                                         |
| Operating system       | Ubuntu 21.10                                       |
| IDEA version           | 2022.1 EAP 6                                       |
| Plugin version         | 3.5                                                |
| Tester                 | @Taikelenn                                         |
| Note                   |                                                    |

When the software doesn't comply with the testing manual...

| ...it is marked with... | ...if the reason for the incompliance is that...  |
| ----------------------- | ------------------------------------------------- |
| **FAILED!!!**           | there's a bug in the software.                    |
| **REMARK.**             | the software should (perhaps) be changed.         |
| normal                  | the testing manual should be changed.             |

### Notes

- **2.2-2.12**. The process for creating a project has changed. One should select "Empty project", press "Create", then go to File -> Project Structure, and set the JDK.
- **3.5**. **Remark** - Add jsoup to the library list
- **4.10**. **Remark** - O1Library and all other modules now appear as "submodules" inside the project main module. The filesystem structure seems to be unchanged, though.
- **5.2**. **Remark** - I had to close the Modules window for the module to finish installing
- **5.13**. **Remark** - Merely opening the Modules window should be enough for search to work. Right now you need to click on some module, which is redundant.
- **7.3**. **Remark** or manual update required - The module has disappeared, but the directory is still present and visible in the project tree.
- **8.5**. **Remark** - This isn't in the testing manual, but I'd expect the news view to also be immediately populated. However, it was empty and I had to refresh it manually.
- **12.2**. All entries are checked, not just non-submittable and optional.

jaakkonarhi:
0.1.B No such option
2.6 Didn't open
3.1 Logo broken :(
4.13 Action disabled (because of indexing)
4.18 Maybe we don't need to restart the whole IDE
4.21 Well there's a button
5.1 I guess I should update these
6.7 RandomTextApp looks like it's not imported even though it is
8.5 Someone complained on Telegram that they can't close the course ended banner
9.10 I wanna check again if I can change the text color

OlliKiljunen (approved):
So there was a character encoding issue with Windows but nothing too crucial.

PR #882, opened 2022-04-19
Remove most of PR template

Taikelenn:
# Description of the PR
Deleted the PR template which asked about testing and stuff; we always just mark "all tests pass" regardless of the actual state of things. Also, the CI is running tests as well, so there's really no point in marking that manually.


OlliKiljunen (approved):
:+1:

PR #891, opened 2022-05-12
REPL custom arguments

Taikelenn:
# Description of the PR
In the course configuration JSON, there is now an optional string called "replArguments". These are additional arguments that are going to be automatically passed to the REPL process.
For example: `"replArguments": "-new-syntax"` means that `-new-syntax` will be passed to Scala REPL process.


src/main/scala/fi/aalto/cs/apluscourses/intellij/actions/ReplAction.scala
```diff
@@ -23,6 +23,7 @@ import org.jetbrains.plugins.scala.console.actions.RunConsoleAction
 import org.jetbrains.plugins.scala.console.configuration.ScalaConsoleRunConfiguration
 import org.jetbrains.plugins.scala.project.ProjectExt
 
+import scala.collection.mutable
```
OlliKiljunen, line 26: Nit: `import scala.collection.mutable.StringBuilder` ?

OlliKiljunen (approved):
:+1:

stellatsv (approved):
No objections

PR #896, opened 2022-05-24
Showing REPL notifications that a module has been changed

Taikelenn:
# Description of the PR
This PR adds a "banner" on top of the REPL window. The banner becomes visible when a user opens a REPL, then makes some code changes and focuses back onto the REPL window without restarting it.

The banner's purpose is to notify the user that the REPL has to be restarted in order for the code changes to be applied.

Screenshot:
obraz


scalastyle/scalastyle-checks.xml
```diff
@@ -61,7 +61,7 @@
  <check level="warning" class="org.scalastyle.scalariform.EqualsHashCodeChecker" enabled="true"></check>
  <check level="warning" class="org.scalastyle.scalariform.IllegalImportsChecker" enabled="true">
   <parameters>
-   <parameter name="illegalImports"><![CDATA[sun._,java.awt._]]></parameter>
+   <parameter name="illegalImports"><![CDATA[sun._]]></parameter>
```
OlliKiljunen, line 64: Is it known why importing java.awt classes is encouraged by the original authors of this file?

src/main/java/fi/aalto/cs/apluscourses/intellij/utils/ReplChangesObserver.java
```diff
@@ -0,0 +1,82 @@
...
+   * @param module The module which has been changed.
+   */
+  public static void onModuleChanged(@NotNull Module module) {
+    synchronized (modifiedModules) {
+      modifiedModules.add(module);
+    }
+  }
+
+  public static boolean hasModuleChanged(@NotNull Module module) {
+    return modifiedModules.contains(module);
```
OlliKiljunen, line 55: Access to `modifiedModules` is not synchronized. Should it be? Or would it be easiest to make it a `synchronizedSet`?

src/main/java/fi/aalto/cs/apluscourses/ui/ReplBannerPanel.java
```diff
@@ -0,0 +1,121 @@
...
+import java.awt.FlowLayout;
+import java.awt.event.MouseEvent;
+import java.awt.event.MouseListener;
+import javax.swing.BorderFactory;
+import javax.swing.JLabel;
+import javax.swing.JPanel;
+
+public class ReplBannerPanel extends JPanel {
+
+  private boolean isPermanentlyHidden = false;
```
OlliKiljunen, line 22: Note to self: Only accessed from EDT.

OlliKiljunen (approved):
Really good! Although not complicated enough for me to like. ;) I wrote some nits. Please, read them and decide whether some changes are needed.

scalastyle/scalastyle-checks.xml
Taikelenn: No, I don't know why java.awt is considered "illegal". But I was able to find such rule in Scalastyle configurations dating as back as 2012, so I don't think anybody remembers the true reason.

Anyway, I consider this wrong, so I changed it :p

stellatsv (approved):
Good that you addressed Olli's comments. I tested this and it works well.

PR #898, opened 2022-05-30
Download filter

stellatsv:
# Description of the PR
Filter for the Modules (Downloaded and Not Downloaded).
Works in a similar way as the Assignment filter.


src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ModuleFilterOptionAction.java
```diff
@@ -0,0 +1,30 @@
+package fi.aalto.cs.apluscourses.intellij.actions;
+
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.project.DumbAwareToggleAction;
+import fi.aalto.cs.apluscourses.presentation.filter.Option;
+import org.jetbrains.annotations.NotNull;
+
+public class ModuleFilterOptionAction extends DumbAwareToggleAction {
```
OlliKiljunen, line 8: Is it reasonable have a separate class instead of reusing the one of the exercise filters?

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ModuleFilterOptionsActionGroup.java
```diff
@@ -0,0 +1,83 @@
...
+   * Returns filter option actions.
+   *
+   * @param e AnActionEvent
+   * @return An array of FilterOptionActions
+   */
+  private ModuleFilterOptionAction @NotNull [] getFilterOptionActions(@Nullable AnActionEvent e) {
+    return Optional.ofNullable(e)
+        .map(AnActionEvent::getProject)
+        .map(mainViewModelProvider::getMainViewModel)
+        .map(MainViewModel::getModules)
```
OlliKiljunen, line 61: This is the only line specific to modules (vs. exercises)? I'd like to see more code reused from the exercise filters.

src/main/java/fi/aalto/cs/apluscourses/intellij/actions/ModuleFilterOptionAction.java
```diff
@@ -0,0 +1,30 @@
...
+
+  @Override
+  public void setSelected(@NotNull AnActionEvent e, boolean state) {
+    option.isSelected.set(state);
+  }
+
+  public Option getOption() {
+    return option;
+  }
+}
```
Taikelenn, line 30: This class is exactly the same as FilterOptionAction, except with a different name. Is there a reason for this separation?

src/main/java/fi/aalto/cs/apluscourses/presentation/base/FilterEngine.java
```diff
@@ -0,0 +1,83 @@
+package fi.aalto.cs.apluscourses.presentation.base;
+
+import fi.aalto.cs.apluscourses.presentation.filter.Filter;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class FilterEngine {
```
OlliKiljunen, line 7: Hmm... Now when we look at this class, it has very little to do with _filtering_ anymore. More suitable name for it could be e.g. "LastComeServedExecutor" or something like that. What do you think? Would you mind changing that?

src/main/java/fi/aalto/cs/apluscourses/ui/base/BaseListView.java
```diff
@@ -91,6 +96,49 @@ public void setModel(@Nullable ListModel<E> model) {
     }
   }
 
+  /**
+  * Set the model of the this tree to the given view model, or do nothing if the given view model
```
OlliKiljunen, line 100: ~tree~ list

src/main/java/fi/aalto/cs/apluscourses/ui/base/BaseListView.java
```diff
@@ -91,6 +96,49 @@ public void setModel(@Nullable ListModel<E> model) {
...
+
+  private void update() {
+    BaseListViewModel<E> localViewModel;
+    synchronized (viewModelLock) {
+      localViewModel = this.viewModel;
+    }
+    var listModel = new DefaultListModel<E>();
+    //keeping selection
+    listModel.addAll(localViewModel.streamVisibleItems().collect(Collectors.toList()));
+    setModel(listModel);
```
OlliKiljunen, line 139: I'm wondering if it would be good to include
```java
synchronized (viewModelLock) {
  if (localViewModel != this.viewModel) return;
}
```
before the last line. What do you think? And the same thing for the treeview. And could the code be reused between these two classes?

src/main/java/fi/aalto/cs/apluscourses/presentation/base/BaseListViewModel.java
```diff
@@ -25,13 +29,18 @@
  *            {@link ListElementViewModel}
  */
 public class BaseListViewModel<E extends ListElementViewModel<?>> extends AbstractListModel<E>
```
OlliKiljunen, line 31: It crossed my mind (yes, in the middle of a night) that this class, perhaps, should no longer extend Swing's `ListModel`. That made me also think there might be a bug in this code related to the functionality of ListModel used in this class: I'm afraid that functionality has no effect any more when objects of this class ain't given to ListView's setModel method.

A solution could be making a new implementation of ListModel, working as an adapter between this class and ListView.

I just wrote these thoughts quickly up so I don't forget the issue. This needs further investigation.

OlliKiljunen:
I take over this PR now :)

OlliKiljunen (approved):

PR #916, opened 2022-07-19
Cache duplicate submissions only if operation is successful

Taikelenn:
# Description of the PR
Previously, a submission was hashed & added to the submitted list even if the submission failed. This meant that if someone's submission failed due to network error, subsequent attempts would show the "duplicate submission" dialog. This has been fixed here.


OlliKiljunen (approved):
Okay, I see the functionality of `DuplicateSubmissionChecker` is now divided to two methods, one called _before_ and one _after_. I think that's good enough although I feel a bit unsure of hash calculated twice (not that much of a performance (non-)issue but more of the clarity and feeling of certainty about concurrency and such although it seems very unlikely any problems would ever occur). I would have probably designed it so that `isDuplicateSubmission` returns some object to which you can later call some `commit` method to apply the change in the storage.

But this is fine and possibly even better than what I would have come up with so approval.

PR #920, opened 2022-07-31
Do not show REPL banner in non-A+ courses

Taikelenn:
# Description of the PR
Will not show the "Reload REPL to apply code changes" banner if the module is not in an A+ project

### DIFF
--- a/src/main/scala/fi/aalto/cs/apluscourses/intellij/Repl.scala
+++ b/src/main/scala/fi/aalto/cs/apluscourses/intellij/Repl.scala
@@ -2,6 +2,7 @@ package fi.aalto.cs.apluscourses.intellij
 
 import com.intellij.execution.ui.ConsoleViewContentType
 import com.intellij.openapi.module.Module
+import fi.aalto.cs.apluscourses.intellij.services.PluginSettings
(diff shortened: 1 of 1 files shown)


src/main/scala/fi/aalto/cs/apluscourses/intellij/Repl.scala
```diff
@@ -22,14 +23,17 @@ class Repl(module: Module) extends ScalaLanguageConsole(module: Module) {
   banner.setVisible(false)
   add(banner, BorderLayout.NORTH)
 
-  // creating a new REPL resets the "module changed" state
-  ReplChangesObserver.onStartedRepl(module)
+  // Do not show the warning banner for non-A+ courses
+  if (PluginSettings.getInstance.getCourseProject(module.getProject) != null) {
```
OlliKiljunen, line 27: Correct me if I'm wrong: if the REPL is opened for a non-A+ project, and after that, the project is turned into an A+ project, that REPL instance stays "non-tracking". All subsequently opened REPL instances, however, get this functionality. Seems reasonable enough to me.

OlliKiljunen (approved):

PR #929, opened 2022-08-16
Added Scala 2 library as class JAR

Taikelenn:
# Description of the PR
Will solve the "no scala-library*.jar dependency" problem when compiling modules. Why Scala 3 requires Scala 2 as a class root is beyond me.

Hardcoding the Scala 2 version is the JAR name is definitely unsafe and should be changed, but given that O1 has settled on 3.1.3, we can be confident this won't change during O1 2022.

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/model/Scala3Sdk.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/model/Scala3Sdk.java
@@ -2,6 +2,7 @@ package fi.aalto.cs.apluscourses.intellij.model;
 
 import fi.aalto.cs.apluscourses.utils.content.Content;
 import fi.aalto.cs.apluscourses.utils.content.RemoteZippedDir;
+import java.util.Arrays;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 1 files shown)


src/main/java/fi/aalto/cs/apluscourses/intellij/model/Scala3Sdk.java
```diff
@@ -29,6 +29,9 @@ public Scala3Sdk(@NotNull String name,
   @Override
   @NotNull
   protected String @NotNull [] getClassRoots() {
-    return new String[] {"scala3-library_3-" + scalaVersion + ".jar"};
+    return new String[] {
+        "scala-library-2.13.8.jar",
```
OlliKiljunen, line 33: You can call `getJarFiles()` and find the filename starting with "scala-library" to avoid hardcoding the version number. 

Perhaps skip Scala2 dependency altogether, if no such JAR is found (as I imagine they could get rid of that in the future).

(Consider adding `streamJarFiles()` to IntelliJLibrary to make handling them smoother.)

OlliKiljunen (approved):

PR #930, opened 2022-08-17
Autoexec REPL commands for Scala 3

Taikelenn:
# Description of the PR
Scala 3 has removed the `-i` switch for preloading REPL commands. Very unfortunate.
So, our alternative approach: get the stdin of the REPL process and write the commands directly.
We also need to:
- make sure this doesn't occur for Scala 2
- call `ScalaLanguageConsole.textSent` so that the executed commands are reflected in the REPL's autocompletion system
- skip the `scala>` prompt for each autoexecuted command

### DIFF
--- a/src/main/scala/fi/aalto/cs/apluscourses/intellij/Repl.scala
+++ b/src/main/scala/fi/aalto/cs/apluscourses/intellij/Repl.scala
@@ -39,12 +39,7 @@ class Repl(module: Module) extends ScalaLanguageConsole(module: Module) {
 
   // We need this here because the overridden ConsoleExecuteAction needs to determine whether
   // the console is hosting a Scala 3 REPL or something else
-  val isScala3REPL: Boolean = ModuleUtils.nonEmpty(
-    ModuleRootManager.getInstance(module)
-      .orderEntries()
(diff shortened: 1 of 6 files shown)


OlliKiljunen:
Before starting to review it, as a general comment/question: should (could?) Scala2 and Scala3 REPLs be handled in separate classes now when it seems they're becoming quite different?

Taikelenn:
> Before starting to review it, as a general comment/question: should (could?) Scala2 and Scala3 REPLs be handled in separate classes now when it seems they're becoming quite different?

It is an option, but I'd say not yet. Right now there are just two differences between Scala 2 and 3 REPLs: multiline support & preloading commands. Ideally, we'd hope that Jetbrains fixes multiline support so we can drop our hacky code, but even if they don't fix it, I feel like separating the REPL into two classes would add significant amount of code without much benefit (and it's not like we're planning to add 10 other Scala3-specific features).
We can think about refactoring the Repl class (maybe create an "investigate" issue for that) but right now we're a bit time-constrained, so I'd say to leave this for later (if ever).

OlliKiljunen (approved):

PR #943, opened 2022-09-15
Small fixes

jaakkonakaza:
# Description of the PR

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -51,10 +51,10 @@
                     class="fi.aalto.cs.apluscourses.intellij.actions.APlusAuthenticationAction"
                     text="Set A+ Token"
                     description="Set an access token for submitting assignments to A+."/>
-            <action id="fi.aalto.cs.apluscourses.intellij.actions.ResetProjectAction"
-                    class="fi.aalto.cs.apluscourses.intellij.actions.ResetProjectAction"
-                    text="Reset A+ Project"
-                    description="Resets the open A+ course project."/>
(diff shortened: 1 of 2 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -51,10 +51,10 @@
...
                     text="Set A+ Token"
                     description="Set an access token for submitting assignments to A+."/>
-            <action id="fi.aalto.cs.apluscourses.intellij.actions.ResetProjectAction"
-                    class="fi.aalto.cs.apluscourses.intellij.actions.ResetProjectAction"
-                    text="Reset A+ Project"
-                    description="Resets the open A+ course project."/>
+            <!--            <action id="fi.aalto.cs.apluscourses.intellij.actions.ResetProjectAction"-->
+            <!--                    class="fi.aalto.cs.apluscourses.intellij.actions.ResetProjectAction"-->
+            <!--                    text="Reset A+ Project"-->
+            <!--                    description="Resets the open A+ course project."/>-->
```
OlliKiljunen, line 57: Juha said some students used this feature to fix some issues. So, should we still keep it?

Another note: if we're getting rid of it, should the lines be removed instead of just commenting out and the class also be removed?

OlliKiljunen (approved):

Taikelenn (commented):
Are we absolutely certain that we want to remove "Reset project" action without replacing it with something else?
If yes, we should delete the class too.

jaakkonakaza:
It's gonna be replaced by something else so I think it can stay for now

PR #954, opened 2022-10-05
Reuse HttpClient to pool connections

Taikelenn:
# Description of the PR
Create a single HttpClient instance and reuse it throughout the lifetime of IntelliJ. This is actually in line with the official performance guidelines https://hc.apache.org/httpclient-legacy/performance.html
TODO: consider if there's an easy wait to periodically evict client-sided CLOSE_WAITs

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
@@ -15,7 +15,7 @@ import org.apache.commons.io.FileUtils;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpRequest;
 import org.apache.http.HttpResponse;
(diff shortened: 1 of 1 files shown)


src/main/java/fi/aalto/cs/apluscourses/utils/CoursesClient.java
```diff
@@ -309,8 +304,11 @@ private static void requireResponseEntity(@NotNull HttpResponse response)
   }
 
   @NotNull
-  private static CloseableHttpClient createClient() {
-    return HttpClients.custom().setUserAgent(getUserAgent()).build();
+  private static synchronized HttpClient createClient() {
```
OlliKiljunen, line 307: `getClient` would be perhaps better name for this.

OlliKiljunen (approved):
👍

PR #970, opened 2022-11-23
Update deprecated APIs

Taikelenn:
# Description of the PR
This addresses JetBrains' compatibility warnings:
- Replaced deprecated call to jsoup library with a new one
- Replaced invocation of a deprecated method with a new one
- Updated minimum IntelliJ version (222.2680.4 is IntelliJ 2022.2). The minimum version is required by the `ContentFactory.getInstance()` call.
- Deleted reference to Apache Commons IO because IntelliJ provides it as well. (We're already doing this with jsoup, so I figured why not)
- Removed some unused code

The resulting plugin is compatible with the upcoming IntelliJ version (2022.3), tested with the [plugin compatibility verifier](https://github.com/JetBrains/intellij-plugin-verifier).

Note: IntelliJ 2022.3 will probably be released very soon, and it will break the plugin due to jsoup changes. We might want to consider releasing an update in the nearest future.
For future reference, we should probably run the plugin verifier more often.


build.gradle
```diff
@@ -22,7 +22,6 @@ repositories {
 }
 
 dependencies {
-    implementation group: 'commons-io', name: 'commons-io', version: '2.11.0'
```
OlliKiljunen, line 25: > Deleted reference to Apache Commons IO because IntelliJ provides it as well. (We're already doing this with jsoup, so I figured why not)

But "doing this" with jsoup is the reason our current plugin version ain't gonna be compatible with IDEA 2022.3, right? So, that's one "why-not" I can think of.

What about "whys" then? Memory, I guess, is one. Do we know the sizes of these libs compared to the size of our plugin and IntelliJ? Are there other "whys"?

src/main/resources/META-INF/plugin.xml
```diff
@@ -5,7 +5,7 @@
             url="https://research.cs.aalto.fi/LeTech/">LeTech Group at Aalto University
     </vendor>
     <!--  this is the range of supported IntelliJ IDEA version the plugin can be used with  -->
-    <idea-version since-build="213.5744.223"/>
+    <idea-version since-build="222.2680.4"/>
```
OlliKiljunen, line 8: I'd like to see a bit newer version here but it is what it is. 🤷‍♂️ Is this version installed in Aalto computers, do you know?

OlliKiljunen (approved):
Approved :+1:

PR #990, opened 2023-01-23
Added automatic cache eviction through course configuration file

### LINKED ISSUE #988: Last modified timestamp to course config file
Basically, what we want is to allow teachers to "force" students to clear their cached API requests (cache purge implemented in #987). The reason is that when a teacher changes the course (especially, structure of the exercises), this change should be visible for the students immediately - not only once they have manually cleared cache or cache gets naturally outdated (in 7 days).

This could happen via "courseLastModified" timestamp in the course config file. The teacher could update that whenever they want the caches to be cleared. The plugin should read that value when loading course config file (on startup + peridiocally when running IntelliJ), and remove all cached API requests that are older than the timestamp.

Taikelenn:
# Description of the PR
Fixes #988 by adding an optional field to the course configuration file called "courseLastModified". This is a long integer, representing a Unix timestamp (in seconds) when the course data was last updated. Any cached entries created before this timestamp will be ignored.

Note: I'm not particularly proud of this solution, mainly because:
- the initial expiration date should be passed to the Cache constructor instead of immediately calling `updateCacheExpiration` (my excuse: given the inheritance hierarchy of the CacheImpl class, the current solution is much shorter and achieves the same result)
- the cache should really be (in my opinion) just a key-value mapping, and it shouldn't be dealing with expiration and stuff (this should be a task for the CoursesClient, but we already have a "refactor cache" issue sitting somewhere)
- it assumes that the clock on the user's machine is set correctly (which nowadays should be a given, though)

It should be a decent solution to solve the immediate problem and get the release going, though.

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
@@ -98,7 +98,7 @@ public class CourseUpdater extends RepeatedTask {
     var progressViewModel =
         PluginSettings.getInstance().getMainViewModel(project).progressViewModel;
     var progress =
-        progressViewModel.start(2, getText("ui.ProgressBarView.refreshingCourse"), false);
+        progressViewModel.start(3, getText("ui.ProgressBarView.refreshingCourse"), false);
     var selectedLanguage = PluginSettings
         .getInstance()
         .getCourseFileManager(project).getLanguage();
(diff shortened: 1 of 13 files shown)


src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
```diff
@@ -116,7 +116,16 @@ protected void doTask() {
...
     progress.increment();
-    updateModules(fetchModulesInfo());
+    JSONObject newCourseConfig = null;
+    try {
+      newCourseConfig = fetchCourseConfiguration();
+    } catch (JSONException | IOException e) {
+      progress.finish();
+      return;
+    }
+    progress.increment();
```
OlliKiljunen, line 126: Nit: this logic could perhaps somehow be moved to another function.

OlliKiljunen (approved):
This is good! The new functionality fits very nicely to the existing code. Great work! 👍

src/main/java/fi/aalto/cs/apluscourses/intellij/model/CourseUpdater.java
Taikelenn: That hypothetical function would either need to have these two exceptions as a checked exception (which will result in the same code as currently) or catch exceptions and return null (which would require a null check instead of a catch-clause; all in all, won't make things simpler)

PR #999, opened 2023-02-21
Change semicolon position in ignored file list

### LINKED ISSUE #996: Fix ".repl-commands" in ignored files list
Quoting @jsorva :
>I randomly came across the following line in one of my IJ settings files:

```xml
    <ignoreFiles list="*.pyc;*.pyo;*.rbc;*.yarb;*~;.DS_Store;.git;.hg;.repl-commands;.svn;CVS;__pycache__;_svn;vssver.scc;vssver2.scc.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands.repl-commands" />
```

Problem seems to be this function in [ProjectViewUtil.java](https://github.com/Aalto-LeTech/aplus-courses/blob/9311ae97b166e682702e1a67c9b301a8909afc6a/src/main/java/fi/aalto/cs/apluscourses/intellij/utils/ProjectViewUtil.java)
```java
  public static void ignoreFileInProjectView(@NotNull String ignoredFileName,
                                             @NotNull Project project) {
    FileTypeManager fileTypeManager = FileTypeManager.getInstance();
    WriteCommandAction.runWriteCommandAction(project, () -> fileTypeManager
        .setIgnoredFilesList(fileTypeManager.getIgnoredFilesList() + ignoredFileName + ";"));
  }
```
originating from 08a81f8190f5ad877868a9b7db321c41868ee9a3.

Semicolon should be inserted before the filename, not after it, because apparently IntelliJ removes the trailing semicolon automatically. But I guess there is some better way to do it. We should also consider cleaning up the mess our plugin has created. Note that in the Juha's example above, `vssver2.scc` gets not correctly ignored because it is postfixed with that stuff.

Taikelenn:
# Description of the PR
Fixes #996 - the previous implementation added a semicolon in the wrong spot.

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/utils/ProjectViewUtil.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/utils/ProjectViewUtil.java
@@ -21,6 +21,6 @@ public class ProjectViewUtil {
                                              @NotNull Project project) {
     FileTypeManager fileTypeManager = FileTypeManager.getInstance();
     WriteCommandAction.runWriteCommandAction(project, () -> fileTypeManager
-        .setIgnoredFilesList(fileTypeManager.getIgnoredFilesList() + ignoredFileName + ";"));
+        .setIgnoredFilesList(fileTypeManager.getIgnoredFilesList() + ";" + ignoredFileName));
   }
 }


OlliKiljunen (approved):
This seems to fix it. Good! :+1: 

However, a couple of questions:
1) What do you think about programmatically fixing those setting files that are already corrupted by the current version?
2) I'm not sure but I guess that the current implementation used to be correct before -- that is, when it was first written. Probably, at some point, it was changed in the platform code that the semicolon is not put in the end of the list, which caused the anomaly. If that's the case, then this fix doesn't look fully future-proof, either. Do you think, there could be some other way to fix this (in model, preferably)?

Taikelenn:
1. Ok
2. I haven't been here at the time, but I downloaded the sources for IntelliJ 2020.1 and the tokenization process seems to be the same. I believe that relying on the fact that the ignored list is always terminated by a semicolon was never correct. Manually adding a semicolon before our entry should be future-proof enough.

OlliKiljunen (approved):
AFAIU, this should fix it. Well done! 👍

For future reference: duplicate values don't matter because the `FileTypeManager` uses set logic to handle the values. (It is actually a bit weird that it doesn't take a collection of strings but one semicolon-separated string because pretty much the only thing it does with that string is to split it and handle each part separately. `getIgnoredFiles`, then again, joins those parts into a single string again. 🤔)

PR #1000, opened 2023-02-22
Optimize imports

Taikelenn:
# Description of the PR
Simply ran IntelliJ's import optimizer, doesn't impact plugin's functionality whatsoever.

(in hindsight, I should have reserved the PR identifier #1000 for something more profound; anyway, happy birthday (?) 🎉)


OlliKiljunen (approved):
Thanks!

PR #1005, opened 2023-03-28
Multilanguage exercises

### LINKED ISSUE #964: Exercise list should conform to the chosen language
For some reason, the exercises are always listed in English although they might have Finnish (or other language) name too (this was probably due to some shortcut taken in a hurry).

#### Technical instructions for the implementation:

Currently, the exercise names are stored in the object model as strings of A+ multilang format, e.g. `"|fi: Esimerkki|en: Example|"` (check the API for the exact format), and the English name is parsed from that when shown in the UI.

Consider creating a class named like `MultiLanguageText` which has a method `get(language)`. (The language can be passed as a string using [two-letter language codes](https://en.wikipedia.org/wiki/ISO_639-1) or some appropriate class.) Then store the name of the exercise in this type inside `Exercise` class, and get the text in correct language from it when shown in the UI. Reuse this pattern elsewhere, if needed. (`MultiLanguageText::get` could have as a second---optional---parameter that defines the fallback behavior in case the given language is not supported: e.g. return text in some default language. Consider!)

Taikelenn:
# Description of the PR
Fixes #964
When fetching the exercises, the plugin now shows them in the correct language (the one that the user has selected when linking the A+ course)

Why I haven't used the `MultiLanguageText` pattern suggested in the issue description? While that approach would have been more "object-oriented", I see no benefit in using such a class in this situation:
- This would be a nice feature if the plugin needed to display the texts in multiple language simultaneously, but it doesn't. Even if the user changes the course language, all exercises are constructed anew, so a potential `MultiLanguageText::get(language)` would only be called with a single possible argument value.
- The current way is much easier to implement - pass the language to the object constructor and that's it. With the `MultiLanguageText` pattern, one would need to pass the language around to all places where the text is displayed, including obscure places such as TreeView search callbacks.
- News are already implemented that way. (Although I noticed this only after I wrote this PR. It even seems I unknowingly used the same variable names as Jaakko with his news implementation. Great minds think alike, it seems)

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/utils/APlusLocalizationUtil.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/utils/APlusLocalizationUtil.java
@@ -7,27 +7,21 @@ import org.jetbrains.annotations.Nullable;
 
 public class APlusLocalizationUtil {
 
-  /**
-   * Parses the English name from the given string, or returns the string if the string contains
-   * no localization. An example of a valid string is {@code "|en:hello|fi:terve|"}.
-   *
-   * @param name A string with a name in multiple languages, or just a single name.
-   */
-  @NotNull
-  public static String getEnglishName(@NotNull String name) {
-    String englishName = StringUtils.substringBetween(name, "|en:", "|");
-    // The name seems to not contain localization
-    return Objects.requireNonNullElse(englishName, name);
-  }
-
   private static String extractLocalizedText(@NotNull String localizedString, @NotNull String languageCode) {
     return StringUtils.substringBetween(localizedString, "|" + languageCode + ":", "|");
   }
 
+  /**
+   * Parses the given localized string and returns the text corresponding to the specified language.
+   * If the entry in the requested language does not exist, the method attempts to localize the string in English.
+   * If that fails too, the whole raw string is returned.
+   * An example of a valid localized string is {@code "|en:hello|fi:terve|"}.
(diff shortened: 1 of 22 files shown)


src/main/java/fi/aalto/cs/apluscourses/utils/APlusLocalizationUtil.java
```diff
@@ -3,20 +3,27 @@
...
+  private static String extractLocalizedText(@NotNull String localizedString, @NotNull String languageCode) {
+    return StringUtils.substringBetween(localizedString, "|" + languageCode + ":", "|");
+  }
+
   /**
-   * Parses the English name from the given string, or returns the string if the string contains
-   * no localization. An example of a valid string is {@code "|en:hello|fi:terve|"}.
+   * Parses the given localized string and returns the text corresponding to the specified language.
+   * If the entry in the requested language does not exist, the method attempts to localize the string in English.
+   * If that fails too, the whole raw string is returned.
```
OlliKiljunen, line 17: Although not important right now, I'd find it more intuitive that if the string has the localized format (with vertical bars and colons), it's not returned as-is, but some language variant (say, the first) is returned as a backup plan if there's neither the requested nor English available. The raw string would be only returned if there was no vertical bars and  colons at all.

OlliKiljunen (approved):
Good reasoning with your design decision! :+1: Approved!

PR #1007, opened 2023-04-18
Check for presence of third-party dependencies

### LINKED ISSUE #965: In startup, check that plugin is correctly installed
A user may install plugin (via disk) incorrectly by selecting---instead of a ZIP file---a JAR file inside it*. This leads to a setup where plugin is installed but libraries it uses are not.

Our plugin should, in initialization activity, check that all the required libraries should be found by its own class loader. (Ideally, the list of dependencies shouldn't be hard-coded. Consider e.g. putting the names of the libraries in the resources in the Gradle build pipeline.)

*) https://youtrack.jetbrains.com/issue/IDEA-289051

Taikelenn:
# Description of the PR
_**Note: do not merge** until the message box & wiki page wording is discussed during a weekly_

Closes #965
During plugin's initialization, the plugin will attempt to reference classes related to all 3rd party dependencies and show a message box if they happen to be missing.

msg

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/utils/PluginIntegrityChecker.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/utils/PluginIntegrityChecker.java
@@ -12,10 +12,11 @@ public class PluginIntegrityChecker {
    */
   public static boolean isPluginCorrectlyInstalled() {
     try {
+      // reference the dependencies' classes; an exception will be thrown if any of these is missing
       return !ZipFile.class.toString().isEmpty() && !JSONObject.class.toString().isEmpty()
           && !Safelist.class.toString().isEmpty() && !HttpClient.class.toString().isEmpty();
-    } catch (Exception ignored) {
-      // the potential exception here is some kind of ReflectiveOperationException, which
+    } catch (Throwable ignored) {
+      // the potential exception here is usually NoClassDefFoundError, which
       // could only be caused by a missing class (= missing dependency)
     }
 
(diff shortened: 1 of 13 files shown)


src/main/java/fi/aalto/cs/apluscourses/utils/PluginIntegrityChecker.java
```diff
@@ -0,0 +1,29 @@
...
+   */
+  public static boolean isPluginCorrectlyInstalled() {
+    try {
+      // reference the dependencies' classes; an exception will be thrown if any of these is missing
+      return !ZipFile.class.toString().isEmpty() && !JSONObject.class.toString().isEmpty()
+          && !Safelist.class.toString().isEmpty() && !HttpClient.class.toString().isEmpty();
+    } catch (Throwable ignored) {
+      // the potential exception here is usually NoClassDefFoundError, which
+      // could only be caused by a missing class (= missing dependency)
+    }
```
OlliKiljunen, line 21: Would it be of any use to show the name of a missing class in the error window?

OlliKiljunen (approved):
:+1: Thanks for extra refactoring, too!

src/main/java/fi/aalto/cs/apluscourses/utils/PluginIntegrityChecker.java
Taikelenn: I don't think it's of much use because the user can't utilize the information to fix the problem themselves. They need to reinstall the plugin one way or another.

PR #1010, opened 2023-05-03
Exercise hiding rules in course configuration file

### LINKED ISSUE #973: Hide certain exercises
The teacher should be able to hide certain exercises / exercise groups (i.e. "weeks") from the exercise list in their course. Add an item to the course config file where those exercises/weeks are specified. Consider whether to use IDs, names (regex?), ordinal numbers or something else to specify the exercises to be hidden.

Taikelenn:
# Description of the PR
Closes #973
In the course configuration file, it is now possible to create a `hiddenElements` array to hide exercises and exercise groups based on ID or regex.

Example:
```
"hiddenElements": [
    {
        "byId": 1234
    },
    {
        "byRegex": "^[a-z]{4}$"
    },
    {
        "byRegex": {
            "fi": "palautekyselyn tulokset",
            "en": "course feedback survey"
        }
    }
]
```

Will hide:
- Exercise (or group) ID 1234
- All 4-letter exercise names
- All exercises containing "palautekyselyn tulokset" (if Finnish was chosen) or "course feedback survey" (if English was chosen)

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/utils/CourseHiddenElements.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/utils/CourseHiddenElements.java
@@ -1,0 +1,79 @@
+package fi.aalto.cs.apluscourses.utils;
+
+import fi.aalto.cs.apluscourses.model.MalformedCourseConfigurationException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class CourseHiddenElements {
+
+  private final @NotNull List<Integer> hiddenIDs = new ArrayList<>();
+
+  private final @NotNull List<String> hiddenRegexes = new ArrayList<>();
+
+  private final @NotNull Map<String, List<String>> hiddenLanguageSpecificRegexes = new HashMap<>();
+
+  @NotNull
+  public static CourseHiddenElements fromJsonObject(@NotNull JSONArray hiddenEntriesArray) {
+    List<Integer> hiddenIDs = new ArrayList<>();
+    List<String> hiddenRegexes = new ArrayList<>();
+    Map<String, List<String>> hiddenLanguageSpecificRegexes = new HashMap<>();
+
+    for (int i = 0; i < hiddenEntriesArray.length(); ++i) {
+      JSONObject entryInfo = hiddenEntriesArray.getJSONObject(i);
+
+      int hiddenID = entryInfo.optInt("byId", -1);
+      if (hiddenID != -1) {
+        // Hide by exercise/group ID
+        hiddenIDs.add(hiddenID);
+        continue;
+      }
+
+      Object hiddenRegex = entryInfo.get("byRegex");
+      if (hiddenRegex instanceof String) {
+        // Hide by language-independent regex
+        hiddenRegexes.add((String) hiddenRegex);
+        continue;
+      } else if (hiddenRegex instanceof JSONObject) {
+        // Hide by language-dependent regex
+        JSONObject hiddenRegexObject = (JSONObject)hiddenRegex;
+        for (String key : hiddenRegexObject.keySet()) {
+          String regex = hiddenRegexObject.getString(key);
+          hiddenLanguageSpecificRegexes.computeIfAbsent(key, k -> new ArrayList<>()).add(regex);
+        }
+
+        continue;
+      }
+
+      throw new JSONException("Element number " + i + " in the \"hiddenEntries\" array is of an unsupported format");
+    }
+
+    return new CourseHiddenElements(hiddenIDs, hiddenRegexes, hiddenLanguageSpecificRegexes);
+  }
+
+  public CourseHiddenElements() {
+    this(null, null, null);
+  }
+
+  public CourseHiddenElements(@Nullable List<Integer> hiddenIDs,
+                              @Nullable List<String> hiddenRegexes,
+                              @Nullable Map<String, List<String>> hiddenLanguageSpecificRegexes) {
+    if (hiddenIDs != null) {
+      this.hiddenIDs.addAll(hiddenIDs);
+    }
+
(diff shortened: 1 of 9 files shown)


src/main/java/fi/aalto/cs/apluscourses/intellij/model/ExercisesUpdater.java
```diff
@@ -138,7 +144,9 @@ private void addExercises(@NotNull List<ExerciseGroup> exerciseGroups,
           }
           var exercise = dataSource.getExercise(exerciseId, points, course.getOptionalCategories(),
               course.getTutorials(), authentication, CachePreferences.GET_MAX_ONE_WEEK_OLD, selectedLanguage);
-          exerciseGroup.addExercise(exercise);
+          if (!hiddenElements.shouldHideObject(exercise.getId(), exercise.getName(), selectedLanguage)) {
```
OlliKiljunen, line 147: Once again, I'm a bit confused of this `ExercisesUpdater` class. We did the filtering already when we loaded dummy exercises, right (in `ExerciseGroup::fromJsonObject`)? But now we do filtering again, because the results of the first round of filtering are not available here. Is it so? :thinking:

I feel a code smell here, but it might be this class in general, not the addition of this PR.

src/main/java/fi/aalto/cs/apluscourses/utils/CourseHiddenElements.java
```diff
@@ -0,0 +1,113 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class CourseHiddenElements {
+
+  private final @NotNull List<Long> hiddenIDs = new ArrayList<>();
+
+  private final @NotNull List<Pattern> hiddenRegexes = new ArrayList<>();
+
+  private final @NotNull Map<String, List<Pattern>> hiddenLanguageSpecificRegexes = new HashMap<>();
```
OlliKiljunen, line 21: I'd make these subclasses of a common superclass/interface: one that filters by id, one that filters by regex, etc.

There could also be composite implementations of the same superclass: one with AND logic and the other with OR logic.

What do you think?

OlliKiljunen (commented):
Nice code and tests! :+1: However, I wish you commented on my points before I approve. (Changes are not necessarily needed but let's discuss!)

src/main/java/fi/aalto/cs/apluscourses/utils/CourseHiddenElements.java
```diff
@@ -0,0 +1,113 @@
...
+    Map<String, List<Pattern>> hiddenLanguageSpecificRegexes = new HashMap<>();
+
+    for (int i = 0; i < hiddenElementsArray.length(); ++i) {
+      JSONObject entryInfo = hiddenElementsArray.getJSONObject(i);
+
+      long hiddenID = entryInfo.optLong("byId", -1);
+      if (hiddenID != -1) {
+        // Hide by exercise/group ID
+        hiddenIDs.add(hiddenID);
+        continue;
```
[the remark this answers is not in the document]
Taikelenn: Yes, that is correct. To be fair, your example is something I haven't though of, and perhaps it should throw an exception if there are more than one fields inside one filter. What do you think about that?

src/main/java/fi/aalto/cs/apluscourses/utils/CourseHiddenElements.java
Taikelenn: I might agree with moving these three filtering features into separate classes. However, I'm skeptical of the value added by AND and OR logic:
- the OR logic can be achieved by just adding another filter
- the AND logic doesn't make much sense with the ID filter (since it can only match one or zero entries) so it could only be useful with the regex filters. And the regex language already has support for logical expressions anyway

src/main/java/fi/aalto/cs/apluscourses/intellij/model/ExercisesUpdater.java
Taikelenn: It's because the exercise list is initially filled with dummy exercises based on the "exercises" API `/api/v2/courses/266/exercises/`. We need to filter these so that we don't add dummy exercises that are supposed to be hidden.

Later, when the dummy exercises are converted to real exercises, `addExercise` is called, which will remove the dummy exercise and add the new one, so we need to prevent this function call for hidden exercises. The list of all exercise IDs here is obtained from a different API (the `points` one), so we can't just reuse the previous results.

There are two alternative solutions I can think of:
1. During the initial API call that is used to create dummy exercises, do the filtering as it currently is. Save all filtered exercise IDs somewhere, and when converting dummy entries to real ones, instead of calling `shouldHideObject` twice, just check for the exercise ID before even fetching it.
Advantages: no need to fetch hidden exercises
Disadvantages: the code becomes even more complicated, because now we have to store hidden exercise IDs somewhere. Additionally, the ideal place of creating such a list of hidden IDs (`fromJsonObject`) is objectively bad (modifying class fields in a deserialization function isn't good), so creating that list itself will be even more code, and that kinda defeats the purpose

2. Assuming that all dummy exercises are present before we fetch the real ones, we can change `addExercise` to do nothing if there is no dummy exercise in the list.
Advantage: no need to call `shouldHideObject` twice
Disadvantages: I'm not confident in that assumption, this sounds almost like a hack and certainly won't make the code more readable

Personally, I don't think the alternative solutions are better than the current one.

OlliKiljunen: Thanks for the explanation! And good reasoning; I agree with you: let's do as it is in this PR.

src/main/java/fi/aalto/cs/apluscourses/utils/CourseHiddenElements.java
Taikelenn: I feel like the AND- and OR-filters are somewhat of an overengineering for a feature that will almost invariably be used to "just hide one feedback exercise with substring X"

src/main/java/fi/aalto/cs/apluscourses/utils/CourseHiddenElements.java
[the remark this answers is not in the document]
Taikelenn: I have removed the JSON keys like "byId" as you suggested

PR #1013, opened 2023-05-24
Update dependencies

### LINKED ISSUE #1011: Problem with showing feedback within IntelliJ
Observed behavior: nothing happens when the user tries to open feedback in IntelliJ.

Opening the same feedback in a browser works as it should.

Taikelenn:
# Description of the PR
- Updated the `runIde` to finally use IntelliJ 2023.1
- The above required updating Gradle to 7.6
- The above required updating Java to version 17

Explanation of the changes:
- GitHub Actions had to be updated to use Java 17.
- `sourceCompatibility` change bumps the source language version to Java 17.
- The additional JUnit entries are required because the newest version of Gradle attempts to fetch JUnit's dependencies in a different way, which caused issues.
- The change to Scala code was needed because a compilation error (which I frankly do not understand) occurred.

As far as my testing goes, this also fixes #1011.

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/setup-java@v3
         with:
           distribution: temurin
-          java-version: 1.8
+          java-version: 17
 
       # The checkout action by default fetches the event that triggered the workflow, which in this case is the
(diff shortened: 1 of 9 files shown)


OlliKiljunen (approved):
Good upgrades! :+1:

PR #1027, opened 2023-06-21
Data Structures & Algorithms features

Taikelenn:
# Description of the PR
This PR adds the following features:
1. When linking IntelliJ course to an A+ course: if there is only one language available, it is selected automatically.
2. Course configuration file gained a new, **optional** boolean field: `requireAuthenticationForModules`. If true, the "Modules" buttons is hidden when no A+ token is provided. The option default to false.
3. Course configuration file gained a new, **optional** object: `callbacks`. For now, the only thing you can do with it is:
```
"callbacks": {
  "postDownloadModule": [
    "AddModuleWatermark"
  ]
},
```
which will run the AddModuleWatermark callback after a module is downloaded. It is used to add watermarks in TRAKY courses. The available callback identifiers (strings) are defined somewhere in the source code.

In the future, the callbacks object can be expanded with more callbacks and callback types.

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/utils/Callbacks.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/utils/Callbacks.java
@@ -2,10 +2,10 @@ package fi.aalto.cs.apluscourses.utils;
 
 import com.intellij.openapi.project.Project;
 import fi.aalto.cs.apluscourses.model.Module;
-import java.lang.reflect.Method;
+import fi.aalto.cs.apluscourses.utils.callbacks.AddModuleWatermark;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.json.JSONArray;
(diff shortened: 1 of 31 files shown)


src/main/java/fi/aalto/cs/apluscourses/utils/Callbacks.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    for (int i = 0; i < callbackArray.length(); i++) {
+      String className = callbackArray.getString(i);
+
+      Class<?> callbackClass = Class.forName("fi.aalto.cs.apluscourses.utils.callbacks." + className);
+      Method method = Arrays.stream(callbackClass.getDeclaredMethods())
+          .filter(m -> m.getName().equals(methodName))
+          .findFirst()
+          .orElseThrow();
```
OlliKiljunen, line 33: I don't think we should use reflection here (even though this looks reasonably secure).

I suggest we just use a single dictionary of callbacks (`Map<String, Callback>`) where all the available callbacks are registered.

Taikelenn: For once I agree that reflection here isn't the best idea. No type safety and questionable extensibility since adding new callbacks requires updating the plugin anyway.
Reflection was my first thought when implementing this, but in hindsight it's not a good solution here :(((

OlliKiljunen (approved):
Good work!

PR #1030, opened 2023-07-23
Delay setting background colors until after the component is added

### LINKED ISSUE #1029: New IntelliJ UI: REPL warning banner lost its background color
obraz

This banner should be red. The same problem also seems to exhibit itself in the "Course ended" banner.

Taikelenn:
# Description of the PR
Fixes #1029

The background colors in our custom JPanels are now working correctly in the new UI. Previous IntelliJ versions (and the old UI) are not affected by this code change.

Reason for this change (**warning: speculation zone**)
Color-wise, the new UI is beautifully uniform. It seems implausible that all UI developers at JetBrains agreed to rewrite their Swing components, so something else must be going on.

What I believe is happening is that whenever a new Swing component is added to the component tree, the background color of all JPanels inside is reset to the theme's default color. Therefore, to override this behavior, we need to set the colors after adding the components.

In the course ended banner, we just needed to add changing color of the internal JPanel.
In the REPL banner, we delayed changing the color until the banner is actually shown.

untitled

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@ plugins {
     id 'scala'
     id 'checkstyle'
     id 'com.github.alisiikh.scalastyle' version '3.5.0'
-    id 'org.jetbrains.intellij' version '1.13.3'
+    id 'org.jetbrains.intellij' version '1.15.0'
     id 'org.sonarqube' version '4.3.0.3225'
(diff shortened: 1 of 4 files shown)


OlliKiljunen (approved):
Good good! :+1:

PR #1040, opened 2023-08-13
Change some interface constructions to appease Plugin Verifier

### LINKED ISSUE #1035: Remove lambda calls to default methods in interfaces
Due to a Plugin Verifier false positive, our plugin will keep throwing these spurious incompatibility warnings caused by two classes:
- Attempt to execute an invokeinterface instruction on a private method `FileFinder.lambda$findFiles$1(Path, String)`
- Attempt to execute an invokeinterface instruction on a private method `TaskManager.lambda$all$0(List)`

We should also probably follow up with the already reported issue on YouTrack: https://youtrack.jetbrains.com/issue/MP-5026/Invalid-compatibility-problem-re-Illegal-private-method-invocation

Taikelenn:
# Description of the PR
Fixes #1035 
The following changes have been made:
- `FileFinder.findFiles` - invoking default interface method in lambda has been replaced by an old-fashioned for-loop
- `TaskManager.all` - replaced lambda with an old-fashioned "new Runnable" class instantiation
- `AddModuleWatermark.addWatermark` - here I noticed that the Plugin Verifier was complaining about missing method `toList`. It seems that `toList` is only available since Java 16, so I replaced that with a Java 11-compatible construction (and a better one, IMO)

### DIFF
--- a/src/main/java/fi/aalto/cs/apluscourses/model/FileFinder.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/model/FileFinder.java
@@ -19,9 +19,16 @@ public interface FileFinder {
         .orElseThrow(() -> new FileDoesNotExistException(directory, filename));
   }
 
+  /**
+   * Resolves an array of filenames to an array of Paths.
+   */
   @NotNull
   default Path[] findFiles(@NotNull Path directory, @NotNull String[] filenames)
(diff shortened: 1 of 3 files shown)


OlliKiljunen (approved):
Nice and clear 👍 Thanks!

PR #1068, opened 2023-10-12
Remove internal methods

### LINKED ISSUE #1066: Reconsider the use of internal methods
We have two of them:
- Internal method `com.intellij.ide.plugins.PluginManagerCore.enablePlugin(PluginId)`
- Internal method `com.intellij.openapi.module.Module.getModuleTypeName()`

Consider whether we can replace them with something that is not internal. From the top of my head: suitable replacements would be `PluginEnabler.HEADLESS.enableById` and `ModuleType.get`

Taikelenn:
# Description of the PR
Fixes #1066
Replaces internal methods with better alternatives. TODO: test


OlliKiljunen (approved):
The code looks ok! :+1:

PR #1077, opened 2023-11-16
Changed an API method call to match the new signature

### LINKED ISSUE #1076: Remove deprecated API calls
Most urgently:

#### 1 compatibility problem
A+ Courses 3.15 is binary incompatible with IntelliJ IDEA Ultimate IU-233.11799.30 due to the following problem

- Method not found (1 problem)
  - Invocation of unresolved constructor `ScalaLibraryPropertiesState.<init>(...)` (1 problem)
    - Method `ScalaSdk.getPropertiesState(...)` contains an `invokespecial` instruction referencing an unresolved constructor `ScalaLibraryPropertiesState.<init>(...)`. This can lead to `NoSuchMethodError` exception at runtime.

Also, 13 deprecated API usages that are not so urgent but, ideally, should be fixed in this issue unless they'll be fixed already with #1067 (many of them are related to the old tutorial framework).

OlliKiljunen:
# Description of the PR

Closes #1076.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,9 +40,10 @@ dependencies {
 
 intellij {
     // this is the version of the IntelliJ IDEA API (SDK) used to build the plugin
-    version = '231.9011.34'
+    // https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#intellij-extension-version
+    version = '233.11799.30-EAP-SNAPSHOT'
     updateSinceUntilBuild = false
-    plugins = ['java', 'org.intellij.scala:2023.1.18']
+    plugins = ['java', 'org.intellij.scala:2023.3.10']
 }
 
 patchPluginXml {
--- a/src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
+++ b/src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
@@ -62,6 +62,6 @@ public class ScalaSdk extends IntelliJLibrary
       @Nullable ScalaLibraryPropertiesState currentState) {
     return new ScalaLibraryPropertiesState(
         ScalaLanguageLevel.findByVersion(scalaVersion).get(),
-        getUris(getCompilerRoots()), new String[0]);
+        getUris(getCompilerRoots()), new String[0], null);
   }
 }


src/main/java/fi/aalto/cs/apluscourses/intellij/model/ScalaSdk.java
```diff
@@ -62,6 +62,6 @@ protected ScalaLibraryPropertiesState getPropertiesState(
       @Nullable ScalaLibraryPropertiesState currentState) {
     return new ScalaLibraryPropertiesState(
         ScalaLanguageLevel.findByVersion(scalaVersion).get(),
-        getUris(getCompilerRoots()), new String[0]);
+        getUris(getCompilerRoots()), new String[0], null);
```
OlliKiljunen, line 65: Passing `compilerBridgeBinaryJar=null` should be fine.

https://github.com/JetBrains/intellij-scala/blob/c24694ef073c4b6573f591e4cb1bcc364f744566/scala/scala-impl/src/org/jetbrains/plugins/scala/project/ScalaLibraryProperties.scala#L15C1-L18C86
```java
/**
[...]
 * @param _compilerBridgeBinaryJar optional compiler bridge jar.<br>
 *                                 When it's None, a bundled bridge will be used (see `Scala/lib/jps` directory in Scala plugin distribution).<br>
 *                                 Custom, non-bundled bridge is mostly required to be able to compile code
 *                                 with RC/Nightly versions of new Scala 3.x compiler
 */
```

About compiler bridge: https://www.scala-sbt.org/1.x/docs/Compiler-Interface.html

build.gradle
```diff
@@ -40,9 +40,10 @@ dependencies {
 
 intellij {
     // this is the version of the IntelliJ IDEA API (SDK) used to build the plugin
-    version = '231.9011.34'
+    // https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#intellij-extension-version
+    version = '233.11799.30-EAP-SNAPSHOT'
```
OlliKiljunen, line 44: Does we have to wait until 2023.3 is out before merging this? Did we have some standard procedure on this?

Taikelenn:
Inside `build.gradle`, I suggest setting the IntelliJ version to `233.11799.241` according to this not-yet-officially-published information (but probably correct) https://youtrack.jetbrains.com/articles/IDEA-A-2100661679/IntelliJ-IDEA-2023.3-Latest-Builds
Targetting an EAP version of IntelliJ API sounds sketchy

PR #1093, opened 2024-01-23
Replace JOptionPane dialogs with IntelliJ platform API messages

### LINKED ISSUE #1087: Replace JOptionPane with Intellij's Messages class when showing dialogs
The latter should be preferred. The former can apparently throw an exception: https://youtrack.jetbrains.com/issue/MPS-36626/Thread-context-was-already-set-com.intellij.openapi.actionSystem.ex.ActionContextElement3223c46e

We haven't faced this issue but changing this shouldn't be time-consuming

Taikelenn:
# Description of the PR
Fixes #1087 

As described in the issue, JOptionPane is now causing some troubles with thread contexts. Therefore it has been replaced with IntelliJ Platform API.
I was unable to easily replace one instance of JOptionPane, but fortunately a workaround exists.


OlliKiljunen (approved):
:+1:

PR #1154, opened 2024-10-29
4.1.0

### LINKED ISSUE #1133: Show better message to students not enrolled on courses
image

Currently courses that do not have a public /api/v2/courses/$id endpoint show the message correctly but since it is public for O1, a generic network error message is shown.

`/api/v2/users/me` could be checked for enrollment, but since it does not show the course for TAs they would not be able to access the course. After [this issue on aplus](https://github.com/apluslms/a-plus/issues/1397) is fixed we can use the `users/me` endpoint

### LINKED ISSUE #1135: Show an error message if the user has regenerated their token

### LINKED ISSUE #1138: Add module exports

### LINKED ISSUE #1141: Hide points from feedback lines
Possibly, use "Category" info from A+ (different from "Point category")

### LINKED ISSUE #1142: Hide submissions IDs in the assignment list

### LINKED ISSUE #1148: Changing the A+ Token could be easier
If the student changes the A+ token in A+ it is not straightforward  to change the token in the plugin

* The plugin will show a network error so the plugin settings are not available through the plugin interface. (They can still be changed in IJ Settings)
* The student might not figure out that they need to press the *Log Out* button on the settings page

Related: 
#1143 Plugin settings are hidden when the token is changed 
#1133 The message could also say "or you have changed your token"

jaakkonakaza:
# Description of the PR
- Module exporting
  - The student chooses a module to export, the group they will submit in, the file name of the zip and the output path of the zip
  - The file name is automatically suggested to be "{studentIdOfSubmitter}_${lastNamesOfGroupMembers}_{moduleName}"
  - The suggested output location is the home dir, but it can be modified
  - A text file is added to the zip containing the names of the group members
  - A notification is shown with a link to show the zip in the OS's file explorer
  - Closes #1138 
- Module importing
  - Users can import multiple zips of modules at once 
  - Useful for assistants when grading many zips and students when testing if their zip works properly
  - The names of the modules are taken from the names of the zips to avoid overlap
- Remove submission IDs from assignment tree
  - This feature was used in the past for assistant mode, the student could tell the assistant their submission ID so the assistant can find it quicker from the IDE or A+
  - Closes #1142
- Hide submission points from feedbacks
  - The plugin checks if the name contains "|en:Feedback|" and the difficulty category is ""
  - It's kind of O1 specific but other approaches seem too complicated for now 
  - Closes #1141 
- Improve detection of enrollment
  - Since A+ added the list of courses the user is assistant in, now the enrollment can be detected from the list of staff and enrolled courses
  - Before the staff courses did not include assistant courses, so assistants could not use the plugin if this way was used for detection 
  - Closes #1133
- Detect if the token has been invalidated
  - If the HTTP status of the /me endpoint is Unauthorized, the user is prompted to re-enter their token  
  - Closes #1135 
  - Closes #1148

TODO:
- Update CHANGELOG.md


OlliKiljunen (approved):
Nice work! Here and there, the code could be a bit cleaner but nothing too dramatic. If in a hurry, I'd say it's fine to merge this as-is.

PR #1180, opened 2025-01-07
Fix directory separator problem

jaakkonakaza:
# Description of the PR

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@ pluginName = A+ Courses
 pluginRepositoryUrl = https://github.com/Aalto-LeTech/aplus-courses
 
 # SemVer format -> https://semver.org
-pluginVersion=4.1.1
+pluginVersion=4.1.2
 
 # Supported build number ranges and IntelliJ Platform versions -> https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html
 pluginSinceBuild=243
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -5,7 +5,7 @@ pluginName = A+ Courses
 pluginRepositoryUrl = https://github.com/Aalto-LeTech/aplus-courses
 
 # SemVer format -> https://semver.org
-pluginVersion=4.1.1
+pluginVersion=4.1.2
```
OlliKiljunen, line 8: The next version will be 4.1.3.

src/main/kotlin/fi/aalto/cs/apluscourses/services/ModuleImportExport.kt
```diff
@@ -58,12 +59,14 @@ class ModuleImportExport(
 
                             val zipFile = FileUtil.createTempDirectory("apluscourses", "modules")
                             zip.entries().asSequence().forEach { entry ->
-                                val entryName = if (entry.name.endsWith(".iml")) {
+                                // Fix: Replace backslashes with slashes in entry names
+                                val fixedEntryName = entry.name.replace('\\', '/')
+                                val entryName = if (fixedEntryName.endsWith(".iml")) {
```
OlliKiljunen, line 64: I paste these lines of our previous code here just for record:
```
    // File names in ZIP should always use forward slashes.
    // See section 4.4.17 of the ".ZIP File Format Specification" v6.3.6 FINAL.
    // Available online: https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
```

OlliKiljunen (approved):

jaakkonakaza:
This is merged but github says it's not?

PR #1194, opened 2025-02-04
SBT Support

jaakkonakaza:
# Description of the PR


gradle.properties
```diff
@@ -3,16 +3,16 @@ pluginGroup=fi.aalto.cs
...
-pluginUntilBuild=243.*
+# pluginUntilBuild=243.*
 # IntelliJ Platform Properties -> https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#configuration-intellij-extension
 platformType=IC
-platformVersion=2024.3.1
+platformVersion=251.18673.35
 # Plugin Dependencies -> https://plugins.jetbrains.com/docs/intellij/plugin-dependencies.html
 # Example: platformPlugins = com.jetbrains.php:203.4449.22, org.intellij.scala:2023.3.27@EAP
-platformPlugins=org.intellij.scala:2024.3.18
+platformPlugins=org.intellij.scala:2025.1.3
```
OlliKiljunen, line 15: Do we now if this Scala dependency prevents it from working with IJ 2024.3?

If the course starts before 2025.1 is out we need a SBT supporting version for IJ 2024.3.

src/main/kotlin/fi/aalto/cs/apluscourses/model/component/ScalaSdk.kt
```diff
@@ -1,54 +1,69 @@
...
 
         writeAction {
             val libraryModel = library.modifiableModel
-            ScalaSdkUtils.ensureScalaLibraryIsConvertedToScalaSdk(
-                provider,
-                library,
-                Option.apply(strippedScalaVersion.substringBeforeLast(".")),
-                CollectionConverters.asScala(compilerClasspath).toSeq(),
-                CollectionConverters.asScala(listOf<File>()).toSeq(),
-                ScalaSdkUtils.resolveCompilerBridgeJar(strippedScalaVersion)
```
OlliKiljunen, line 50: So, did we get rid of this method call altogether?

src/main/kotlin/fi/aalto/cs/apluscourses/model/component/ScalaSdk.kt
```diff
@@ -1,54 +1,69 @@
...
+
+            // HACK: this is the only way to access properties that I am aware of
```
OlliKiljunen, line 52: This is from our old code, right? Back to the way we handled it previously?

src/main/kotlin/fi/aalto/cs/apluscourses/services/ModuleImportExport.kt
```diff
@@ -48,7 +46,11 @@ class ModuleImportExport(
             return
         }
         val modulesWithErrors = mutableListOf<String>()
-        FileChooser.chooseFiles(FileChooserDescriptorImpl(), project, null) { files ->
+        FileChooser.chooseFiles(
+            FileChooserDescriptor(false, false, true, true, false, true).withExtensionFilter(
```
OlliKiljunen, line 50: Nit: I assume Kotlin has some way to name arguments in a function call. ;)

src/main/kotlin/fi/aalto/cs/apluscourses/ui/Utils.kt
```diff
@@ -19,13 +28,51 @@ object Utils {
...
             setIcon(icon, false)
         }
+
+    /**
+     * A wrapper for JBList to be used with the UI DSL.
+     */
+    inline fun <reified T> Row.list(
+        items: ObservableMutableProperty<List<T>>,
+        renderer: ListCellRenderer<in T?>
+    ): Cell<JBList<T>> {
```
OlliKiljunen, line 47: I really like this function and the next one, too. Do you know if Observable*s link via weak references so that the UI elements can be GC'ed even if the model objects still exist? Or is the lifecycle of models similar to the corresponding UI elements?

OlliKiljunen (approved):
I left a couple of questions but if you know those are fine, feel free to proceed with this as-is. Good work! 👍

src/main/kotlin/fi/aalto/cs/apluscourses/model/component/ScalaSdk.kt
jaakkonakaza: I decided to use the old way for now because the new one only works with either 2024 or 2025, I will ask the IntelliJ Scala team for a better solution

gradle.properties
jaakkonakaza: AFAIK that version number is used only in the sandbox

src/main/kotlin/fi/aalto/cs/apluscourses/ui/Utils.kt
jaakkonakaza: I think it's tied to the lifecycle of the UI element but not 100% sure

PR #1197, opened 2025-02-04
Close the AI ad

jaakkonakaza:
Closes the AI ad when opening the A+ toolwindow


src/main/kotlin/fi/aalto/cs/apluscourses/toolwindows/APlusToolWindowFactory.kt
```diff
@@ -186,6 +192,15 @@ internal class APlusToolWindowFactory : ToolWindowFactory, DumbAware {
         if (InitializationStatus.isNotCourse(project) || InitializationStatus.isIoError(project)) {
             removeAllTabs()
         }
+
+
+        val fileEditorManager: FileEditorManager = FileEditorManager.getInstance(project)
+        fileEditorManager.allEditors.forEach { editor: FileEditor ->
+            if (editor.file != null && editor.file.name == "AI Assistant") {
```
OlliKiljunen, line 199: Just wondering if there could be any more robust way to check the identity. Can the name be translated to other languages? Anyway, as it's a minor issue, we don't need to cover all the possible edge cases.

OlliKiljunen (approved):

PR #1220, opened 2025-05-21
Update platform version

jaakkonakaza:
Tried to change the versions to fix the pipeline but it was Jetbrains' problem https://youtrack.jetbrains.com/issue/MP-7480/


jaakkonakaza:
https://youtrack.jetbrains.com/issue/MP-7480/

PR #1272, opened 2026-05-26
Fix scala plugin REPL incompatibility; separately download and configure REPL for scala >= 3.8

### LINKED ISSUE #1270: Scala SDK/REPL installment in Scala v3.8+
In Scala v3.8 REPL is separated from the SDK, see: https://scala-lang.org/news/3.8/

This has been taken into account in IntelliJ's Scala plugin v2025.3.26 causing an API change that makes our plugin incompatible with new versions of Scala plugin and, consequently, IntelliJ itself. (see their code comments [here](https://github.com/JetBrains/intellij-scala/blob/idea253.x/scala/scala-impl/src/org/jetbrains/plugins/scala/project/ReplClasspath.scala)) The issue occur in runtime when trying to install a Scala SDK (with a Scala module). Hence, the issue doesn't affect the ongoing Python courses.

There are two steps to be done, listed below. The first one is a quickfix to make the plugin compatible with current IntelliJ versions and Scala up to v3.7. It can be skipped, though, if the second step can be completed soon enough. 
1) **(quickfix for Scala v3.7 and earlier)** Adjust our plugin to the changed API of IntelliJ Scala plugin by calling `ScalaLibraryPropertiesState` constructor with `replClasspath=arrayOf()` (empty array means that the REPL is assumed to be bundled with the SDK).
2) **(make it support all Scala versions)** When installing a Scala SDK, Scala v3.8+ must be handled separately (for some parts of the process). On top of the Scala SDK files, `scala3-repl` JAR must be downloaded and stored with the other JARs. `ScalaLibraryPropertiesState` constructor must be called with `replClasspath` pointing to the correct directory.

murhum1:
# Description of the PR

- Updated Scala plugin version compatibility
- Fixed #1270 by manually adding REPL jar when Scala version >= 3.8
- Add Scala library sources from jars downloaded from maven instead of github zips

As mentioned, I was unsure whether manually downloading REPL was even necessary since it looks to be present in the downloaded Scala SDK zip and as a dependency in scala.jar's MANIFEST.MF classpath, but I had the code present already and it was deemed safer to do a manual download anyway, as we figured there was no guarantee the repl jar would be there in the future as well.

### DIFF
--- a/src/main/kotlin/fi/aalto/cs/apluscourses/model/component/Component.kt
+++ b/src/main/kotlin/fi/aalto/cs/apluscourses/model/component/Component.kt
@@ -2,7 +2,13 @@ package fi.aalto.cs.apluscourses.model.component
 
 import com.intellij.openapi.project.Project
 import fi.aalto.cs.apluscourses.services.CoursesClient
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.nio.file.Files
 import java.nio.file.Path
+import java.nio.file.StandardCopyOption
+import kotlin.io.path.createTempFile
+import kotlin.io.path.nameWithoutExtension
 
 abstract class Component<T>(val name: String, protected val project: Project) {
     var dependencyNames: Set<String>? = null
@@ -22,6 +28,24 @@ abstract class Component<T>(val name: String, protected val project: Project) {
 
     protected suspend fun downloadAndUnzipZip(zipUrl: String, extractPath: Path, onlyPath: String? = null) {
         CoursesClient.getInstance(project).downloadAndUnzip(zipUrl, extractPath, onlyPath)
+    }
+
+    protected suspend fun downloadFile(url: String, target: Path): Path = withContext(Dispatchers.IO) {
+        Files.createDirectories(target.parent)
+        val tmp = Files.createTempFile(
+            target.parent,
+            target.fileName.toString(),
+            ".part"
+        )
+        try {
+            CoursesClient.getInstance(project).download(url, tmp.toFile())
+            Files.move(tmp, target, StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.ATOMIC_MOVE)
+        } catch (t: Throwable) {
+            try {
+                Files.deleteIfExists(tmp)
+            } catch (_: Throwable) { }
+            throw t
+        }
     }
 
     protected abstract fun findDependencies(): Set<String>
--- a/src/main/kotlin/fi/aalto/cs/apluscourses/model/component/Component.kt
+++ b/src/main/kotlin/fi/aalto/cs/apluscourses/model/component/Component.kt
@@ -30,22 +30,8 @@ abstract class Component<T>(val name: String, protected val project: Project) {
         CoursesClient.getInstance(project).downloadAndUnzip(zipUrl, extractPath, onlyPath)
     }
 
-    protected suspend fun downloadFile(url: String, target: Path): Path = withContext(Dispatchers.IO) {
-        Files.createDirectories(target.parent)
-        val tmp = Files.createTempFile(
-            target.parent,
-            target.fileName.toString(),
-            ".part"
-        )
-        try {
-            CoursesClient.getInstance(project).download(url, tmp.toFile())
-            Files.move(tmp, target, StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.ATOMIC_MOVE)
-        } catch (t: Throwable) {
-            try {
-                Files.deleteIfExists(tmp)
-            } catch (_: Throwable) { }
-            throw t
-        }
+    protected suspend fun downloadFile(url: String, target: Path) {
+        CoursesClient.getInstance(project).downloadFile(url, target)
     }
 
     protected abstract fun findDependencies(): Set<String>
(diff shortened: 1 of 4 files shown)


src/main/kotlin/fi/aalto/cs/apluscourses/model/component/Component.kt
```diff
@@ -24,6 +30,24 @@ abstract class Component<T>(val name: String, protected val project: Project) {
         CoursesClient.getInstance(project).downloadAndUnzip(zipUrl, extractPath, onlyPath)
     }
 
+    protected suspend fun downloadFile(url: String, target: Path): Path = withContext(Dispatchers.IO) {
```
OlliKiljunen: `CoursesClient.downloadAndUnzip` uses functions in `com.intellij.platform.ide.progress` and `com.intellij.platform.util.progress` to make the progress visible in IntelliJ's progress bar. Do you think similar approach could be applied here? Would it make sense to move this functionality to e.g. that class, too?

src/main/kotlin/fi/aalto/cs/apluscourses/model/component/ScalaSdk.kt
```diff
@@ -29,7 +32,6 @@ class ScalaSdk(private val scalaVersion: String, project: Project) : Library(sca
         CoursesLogger.info("Downloading Scala SDK $scalaVersion")
         val zipUrl =
             "https://github.com/lampepfl/dotty/releases/download/$versionNumber/scala3-$versionNumber.zip"
```
OlliKiljunen, line 35: Not really related to this PR, but I wonder if `lampepfl/dotty` should be changed to `scala/scala3` in the `zipUrl`. But I guess the former keeps redirecting to the latter, so it wouldn't really matter.

src/main/kotlin/fi/aalto/cs/apluscourses/model/component/ScalaSdk.kt
```diff
@@ -77,40 +100,62 @@ class ScalaSdk(private val scalaVersion: String, project: Project) : Library(sca
...
+            scala3Ver
+        } else {
+            fun findStdlibUnder(root: Path): String? {
+                if (!Files.exists(root) || !root.isDirectory()) return null
+                return Files.walk(root).use { stream ->
+                    stream.filter { Files.isRegularFile(it) }
+                        .map { it.fileName.toString() }
+                        .filter { it.startsWith("scala-library-") && it.endsWith(".jar") }
+                        .map { it.removeSuffix(".jar").substringAfter("scala-library-") }
+                        .filter { it.startsWith("2.") }
```
OlliKiljunen, line 127: Nice avoidance of regular expressions ;)

src/main/kotlin/fi/aalto/cs/apluscourses/model/component/ScalaSdk.kt
```diff
@@ -77,40 +100,62 @@ class ScalaSdk(private val scalaVersion: String, project: Project) : Library(sca
...
+                        .findFirst()
+                        .orElse(null)
+                }
+            }
+            findStdlibUnder(libDir)
+                ?: findStdlibUnder(m2Dir)
+                ?: error("scala-library 2.x jar not found under $libDir or $m2Dir for Scala $scala3Ver")
```
OlliKiljunen, line 134: Nit: not sure if this is such a case where Kotlin's [error](https://kotlinlang.org/docs/exceptions.html#error-function) function should be used. Not a real issue, though.

src/main/kotlin/fi/aalto/cs/apluscourses/model/component/ScalaSdk.kt
```diff
@@ -77,40 +100,62 @@ class ScalaSdk(private val scalaVersion: String, project: Project) : Library(sca
...
-                    path.resolve("scala3-$versionNumber").resolve("src").resolve("scala-$scala2Version")
-                        .resolve("src").resolve("library")
-                ),
-                OrderRootType.SOURCES
-            )
+            fun addSourcesJar(path: Path) {
+                if (Files.exists(path)) {
+                    libraryModel.addRoot(VfsUtil.getUrlForLibraryRoot(path.toFile()), OrderRootType.SOURCES)
+                }
+            }
```
OlliKiljunen, line 157: Did I get it right that, previously, sources were unzipped, but now they are included as a (still zipped) JAR file? I guess the both ways work? The latter sounds more reasonable, anyway.

OlliKiljunen (commented):
Thank you for the PR! The code looks fine. Before approving, I'd like to hear your opinion on my comment in **Component.kt** file.

src/main/kotlin/fi/aalto/cs/apluscourses/model/component/Component.kt
murhum1: Yep, moved the implementation there now, and utilized the same reporter functions as in `downloadAndUnzip`.

src/main/kotlin/fi/aalto/cs/apluscourses/model/component/ScalaSdk.kt
murhum1: Yes, that's what I understood - IntelliJ docs aren't very clear on that, though, but seems to work.

murhum1:
> Thank you for the PR! The code looks fine. Before approving, I'd like to hear your opinion on my comment in **Component.kt** file.

Thanks for the review! Moved the functionality to `CoursesClient` as suggested now.

OlliKiljunen (approved):
Thank you for considering my comments. I think this can be merged now.

